PR #1, opened 2020-01-16
Split to edge-messaging and graph-chat

epoberezkin:


gvansanden:
I don't so much care that Simplex does unified push as that battery usage is and remains problematic.
I have Telegram and Signal without Google notifications, Element with unified push and all of them combined use less of my battery then Simplex.  So it should be possible to do notifcations without unified push at a reasonable battery consumption.  Currently, I'm back at 28%

PR #19, opened 2020-01-25
Simplify API, remove connection ID, add single message API

### LINKED ISSUE #12: Simplify simplex messaging
Other questions to simplify
- Maybe the keys should not be replaceable as well and the only needed endpoint is to secure connection which is only possible once? In this case duplex protocol would do full connection rotation (rather than key rotation).
- Is ID really needed?
- Remove "disable" option

epoberezkin:
Simplify API, remove connection ID, add single message API.

Closes #12

### DIFF
--- a/simplex-messaging-implementation.md
+++ b/simplex-messaging-implementation.md
@@ -4,16 +4,14 @@ This document defines specific elements to be used by client and server implemen
 
 This document defines:
 - [cryptographic algorithms](#cryptographic-algorithms) to sign/verify requests and to encrypt/decrypt messages.
-- required approaches to generate:
-   - [connection IDs](#connection-id) for clients.
-   - [connection URIs](#connection-uris) for servers.
+- [connection URIs](#connection-uris) generation by the servers.
 - [privacy requirements](#privacy-requirements) to the servers.
 - [REST API](#rest-api):
-   - to create connections and to update connection attributes.
-   - to send and to retrieve messages.
-- [WebSockets API](#websockets-api) to subscribe to connections:
+   - to create and to secure connections.
+   - to send, to retrieve and to delete messages.
+- [WebSockets API](#websockets-api):
+   - to subscribe to connections.
    - to receive the new messages.
-   - to update connection attributes.
 - any other requirements for simplex messaging servers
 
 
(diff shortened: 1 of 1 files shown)


simplex-messaging-implementation.md
```diff
@@ -214,86 +191,103 @@ All further requests to the recipient and sender connection URIs MUST be rejecte
...
+Optional query string parameter can be used for pagination:
+- `fromMessageID` (string, optional): if set, the server will retrieve the messages received starting from the message with server message ID (unique per server) passed in this parameter.
 
-Simplex messaging server MUST permanently remove the retrieved messages, unless specifically instructed by the clients to keep them.
+Simplex messaging server will NOT delete the messages when this endpoint is called, to remove messages once retrieved [Delete messages](#delete-messages) should be called.
 
 __Please note__: server implementations MUST NOT track in any form how many times or whether the messages were retrieved.
 
 If the unknown message ID is passed in `afterMessageID` parameter, the request should be rejected with HTTP status code 400 (Bad Request).
```
efim-poberezkin: should be `fromMessageID`

simplex-messaging-implementation.md
```diff
@@ -214,86 +191,103 @@ All further requests to the recipient and sender connection URIs MUST be rejecte
...
+Simplex messaging clients MUST use this endpoint to delete the previously retrived and stored (or processed) messages.
 
 Request body should be sent as JSON object with the following properties:
-- `connectionID` (string): existing connection ID (see [Connection ID](#connection-id)).
-- `pageSize` (number, optional): if set, the server will delete up to the requested number of messages, otherwise all messages, in both cases from the message ID in `fromMessageID` parameter (or from the earliest available).
-- `fromMessageID` (string, optional): the server will delete the messages received, starting from the message with the server message ID passed in this parameter. If not specified, it defaults to the server message ID of the earliest received message.
+- `fromMessageID` (string): the server will delete the messages starting from the message with the server message ID passed in this parameter.
+- `count` (number): the server will delete up to the requested number of messages from the message ID in `fromMessageID` parameter.
```
efim-poberezkin: why not make `count` optional to make client slightly more simple? e.g. when there'no large messages and it's the only/the last page it could just request to delete all messages starting with `fromMessageID`. Also wouldn't `upToMessageID` be simpler than `count`? the client workflow to request to message deletion would already be somewhat convoluted with necessity to keep large messages, I think it might be easier to skip them with [`fromMessageID`, `upToMessageID_1`), [`upToMessageID_1[+1]`, `upToMessageID_2`), etc. than with counts.

epoberezkin: proposed only single message deletion endpoint, as clients won't use bulk deletion normally and use websockets instead.

Issue #6, opened 2020-01-19
Use separate keys to encrypt and to sign messages


epoberezkin:
- Not okay to use the same EK to both encrypt and sign
- it's actually not possible to verify messages in edge-messaging protocol if only one connection available, unless sender provides verification key (so in addition to SK to sign requests, sender needs to provide VK to verify messages).
- it requires the same number of simplex connections for both parties in duplex (limited to the smallest number of servers available to each party); without this constraint two parties can use different number of simplex connections (redundancy level) defined by the number of servers available to them.

epoberezkin:
Still not clear why separate key pairs are needed - the only possible problem can arise if the party can be made to use to sign encrypted message - it would lead to its decryption. This is not part of the protocol though, so having separate keys only complicates things (probably) without adding any security.

PR #22, opened 2020-01-25
Remove connection ID and alternative flow

### LINKED ISSUE #8: Remove connection ID
- [x] [simplex-messaging.md](https://github.com/simplex-chat/protocol/blob/master/simplex-messaging.md)
- [x] [simplex-messaging-implementation.md](https://github.com/simplex-chat/protocol/blob/master/simplex-messaging-implementation.md)
- [x] [graph-chat.md](https://github.com/simplex-chat/protocol/blob/master/graph-chat.md)

### LINKED ISSUE #20: Remove alternative flow with 2 out-of-band messages

epoberezkin:
closes #8, closes #20


graph-chat.md
```diff
@@ -30,20 +30,18 @@ The following symbols are used below:
...
   - `RKAi` - Alice's recipient key of `CAi`.
   - `SKAi` - Bob's sender key of `CAi`.
 - keys created for Bob's connections
+  - `RUBi` - server-generated recipient connection URI of `CBi` (to be used by Bob to retrive messages).
   - `EKBi` - Bob's assymetric key pair used for:
     - Alice to encrypt and Bob to decrypt messages from Alice sent via `CBi`.
     - Bob to sign and Alice to verify messages from Bob sent via `CAi`.
-  - `IDBi` - client-generated ID of connection `CBi`
-  - `RUBi` - server-generated recipient connection URI of `CBi` (to be used by Bob to retrive messages).
   - `SUBi` - server-generated sender connection URI of `CBi` (to be used by Alice to send messages).
```
efim-poberezkin: shouldn't `RUAi`, `SUAi`, `RUBi` and `SUBi` be moved under a separate section / sections (e.g. "connection URIs" or "[Alice's / Bob's] connection URIs")? I mean they're not technically keys like `EKs`, `RKs` and `SKs`, aren't they?

simplex-messaging.md
```diff
@@ -118,17 +115,6 @@ This protocol also can be used for off-the-record messaging, as Alice and Bob ca
 How simplex connections are used by the participants (graph vertices) is defined by graph-chat protocol and is not in scope of this low level simplex messaging protocol.
 
 
-## Alternative flow to establish a simplex connection
```
efim-poberezkin: I'm still not fully convinced it's such a bad idea to have this around even if it's a premature optimization and we don't want to spend time implementing it yet. Maybe mark the whole section as `[out of scope for protocol v1 implementation]` or smth along those lines?

graph-chat.md
epoberezkin: it can be changed together with #6

simplex-messaging.md
epoberezkin: I don't think we should until we see that the performance of the standard flow is not good enough (and I believe it will be fast anyway). Any extra flow to maintain is an additional security risk to manage, test, debug, analyse, etc. - better not have it.

efim-poberezkin: okay, well, right.. anyway it'd be possible to restore it via vcs..

PR #43, opened 2021-04-21
docs, smp: align with implementation

efim-poberezkin:


simplex-messaging.md
```diff
@@ -77,12 +77,14 @@ The out-of band invitation message is sent via some trusted alternative channel
 The [ABNF][8] syntax of the message is:
 
 ```abnf
-outOfBandMsg = encryptionKey CRLF senderConnId CRLF server CRLF serverKeyHash
-encryptionKey = encoded ; base64
-senderConnId = encoded
-server = hostname [":" port]
+outOfBandMsg = "smp::" server "::" queueId "::" encryptionKey
+server = hostname [":" port] ["#" serverKeyHash]
+hostname = 1*VCHAR
```
efim-poberezkin, line 82: @epoberezkin this is probably incorrect, what do you think would be best - keep like this / write correct abnf / remove it as before?

efim-poberezkin: nvm, I've found you can use `prose-val` for such things:
https://tools.ietf.org/html/rfc5234#section-4

I propose we gradually switch to it in other places as well to avoid unresolvable rulesets.

epoberezkin (approved):

PR #61, opened 2021-05-30
support ad-hoc groups (broadcasts)

epoberezkin:


apps/dog-food/Main.hs
```diff
@@ -70,19 +70,31 @@ data ChatCommand
...
+  | DeleteGroup Group
+  | ListGroup Group
+  | SendGroupMessage Group ByteString
+  deriving (Show)
 
 chatCommandP :: Parser ChatCommand
 chatCommandP =
   ("/help" <|> "/h") $> ChatHelp
-    <|> ("/markdown" <|> "/m") $> MarkdownHelp
+    <|> ("/group #" <|> "/g #") *> (NewGroup <$> group)
```
efim-poberezkin, line 84: add these to help? not a long explanation of workflow as for connections, but a list of commands and what they do

efim-poberezkin: even if we don't want to announce this as a feature for public use, make a release and explain in readme, at least help would be helpful for ourselves

epoberezkin: I'd rather not... readme is related to the current release at the moment, we can merge it to master but not advertise.

efim-poberezkin (approved):

PR #62, opened 2021-06-10
simplex-chat schema, refactor chat to use SMP agent functions

epoberezkin:


apps/simplex-chat/Store.hs
```diff
@@ -0,0 +1,23 @@
+{-# LANGUAGE TemplateHaskell #-}
+
+module Store where
+
+import Data.FileEmbed (embedDir, makeRelativeToProject)
+import Data.Function (on)
+import Data.List (sortBy)
+import Data.Text.Encoding (decodeUtf8)
+import Simplex.Messaging.Agent.Store.SQLite
+import Simplex.Messaging.Agent.Store.SQLite.Migrations (Migration (..))
```
efim-poberezkin, line 10: separate migrations from agent in future?

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
+CREATE TABLE profiles (
+  profile_id INTEGER NOT NULL PRIMARY KEY,
```
efim-poberezkin, line 2: do we expect `profile_id` to be sqlite's _"Integer Primary Key"_ in this case (i.e. an alias for rowid) or not? If we do, it's not clear it'd be an alias for rowid - see analogous bug described for DESC - https://sqlite.org/lang_createtable.html#rowid. If we don't either way it's ambiguous. Also `INTEGER PRIMARY KEY` primary key would already not allow nulls, unlike some other sqlite's primary keys - this would remove the ambiguity.

anyway, why not make `display_name` PK?

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+
+CREATE TABLE known_servers(
+  host TEXT NOT NULL,
+  port TEXT NOT NULL,
+  key_hash BLOB,
+  PRIMARY KEY (host, port)
+) WITHOUT ROWID;
+
+CREATE TABLE contacts (
+  contact_id INTEGER NOT NULL PRIMARY KEY,
```
efim-poberezkin, line 15: same, `NOT NULL` in the middle makes it ambiguous whether the column would be rowid or not

efim-poberezkin: from docs:
`With one exception noted below, if a rowid table has a primary key that consists of a single column and the declared type of that column is "INTEGER" in any mixture of upper and lower case, then the column becomes an alias for the rowid`
so I guess it's not ambiguous.. still I'd remove `NOT NULL`s

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+  display_name TEXT NOT NULL, -- display name set by local user
+  properties TEXT NOT NULL, -- JSON set by local user
+  profile_id INTEGER REFERENCES profiles, -- sent by remote contact
+  contact_status TEXT NOT NULL DEFAULT ''
```
efim-poberezkin, line 19: let's not have empty strings as default please - it's no better than `NULL` except for when you have to work with it in the db - then in many interfaces it's unclear whether you're dealing with an empty string or a null

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+  port TEXT NOT NULL,
+  key_hash BLOB,
+  PRIMARY KEY (host, port)
+) WITHOUT ROWID;
+
+CREATE TABLE contacts (
+  contact_id INTEGER NOT NULL PRIMARY KEY,
+  display_name TEXT NOT NULL, -- display name set by local user
+  properties TEXT NOT NULL, -- JSON set by local user
+  profile_id INTEGER REFERENCES profiles, -- sent by remote contact
```
efim-poberezkin, line 18: if a profile is sent by remote contact, how is it that a profile can be referenced by multiple contacts? shouldn't it be vice versa?

if it's 1 to 1 `profile`'s `properties` could just be stored on `contact` as `external_properties`

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+  contact_status TEXT NOT NULL DEFAULT ''
+);
+
+CREATE TABLE contact_connections (
+  connection_id BLOB NOT NULL PRIMARY KEY,
```
efim-poberezkin, line 23: `connection_id` -> `conn_id`?

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+  contact_id INTEGER NOT NULL REFERENCES contacts ON DELETE RESTRICT,
+  conn_level INTEGER NOT NULL DEFAULT 0,
```
efim-poberezkin, line 25: I'd remove default so that one has to remember to calculate and set `conn_level` from the app, otherwise you risk forgetting to calculate a level different from zero and finding out too late. There's no downside to let this constraint fail - it would only reveal an error in implementation, it doesn't pose a risk to breaking end user's application. As specified by `NOT NULL`, this column is part of the logic and always has to be set, it's not just some descriptive optional field, if anything having `DEFAULT 0` for it only water the constraint's meaning down - why is it specifically 0 that is default, why not 1 or 2 - they would have the same semantic weight - setting a 0 for level 2 conn would be as much a mistake as setting a 2 for level 0 conn

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+  via_conn BLOB REFERENCES contact_connections (connection_id),
+  conn_status TEXT NOT NULL DEFAULT ''
```
efim-poberezkin, line 27: same regarding empty default string

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+) WITHOUT ROWID;
+
+CREATE TABLE contact_invitations (
+  invitation_id BLOB NOT NULL PRIMARY KEY,
+  contact_id INTEGER NOT NULL REFERENCES contacts ON DELETE RESTRICT,
+  invitation_status TEXT NOT NULL DEFAULT ''
```
efim-poberezkin, line 33: same

src/Simplex/Chat/Protocol._hs
```diff
@@ -1,13 +1,42 @@
...
-    connection :: ConnAlias
+  { contactId :: ByteString,
+    profile :: Profile,
+    connections :: [Connection]
+  }
+
+data Connection = Connection
+  { connId :: ConnId,
+    connLevel :: Int,
+    viaConn :: ConnId
```
efim-poberezkin, line 21: `Maybe ConnId`?

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+) WITHOUT ROWID;
+
+CREATE TABLE groups (
+  group_id INTEGER NOT NULL PRIMARY KEY, -- local group ID
+  display_name TEXT NOT NULL, -- shared
```
efim-poberezkin, line 38: why is `display_name` shared here but local on `contacts`? does a group even need a local display name? maybe it could just be part of shared `properties`?

also `properties` -> `shared_properties` / `external_properties`? 1st is it's just for explicitness since we use the same field name for different purposes in different tables, 2nd is we might need `local` / `internal_properties` for groups as well - e.g. whether group is muted or not

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+  properties TEXT NOT NULL -- shared JSON group profile
+);
+
+CREATE TABLE group_members ( -- group members, nullable fields are NULL for the local user
+  group_id INTEGER NOT NULL REFERENCES groups ON DELETE RESTRICT,
+  member_id BLOB NOT NULL, -- shared member ID, unique per group
+  member_role TEXT NOT NULL DEFAULT '', -- owner, admin, moderator, ''
+  member_status TEXT NOT NULL DEFAULT '', -- inv | con | full | off
```
efim-poberezkin, line 46: same for default ''

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+) WITHOUT ROWID;
+
+CREATE TABLE groups (
+  group_id INTEGER NOT NULL PRIMARY KEY, -- local group ID
+  display_name TEXT NOT NULL, -- shared
+  properties TEXT NOT NULL -- shared JSON group profile
+);
+
+CREATE TABLE group_members ( -- group members, nullable fields are NULL for the local user
+  group_id INTEGER NOT NULL REFERENCES groups ON DELETE RESTRICT,
```
efim-poberezkin, line 43: why `RESTRICT`, why not `CASCADE`? doesn't have to delete contacts, only group members

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+  member_id BLOB NOT NULL, -- shared member ID, unique per group
+  member_role TEXT NOT NULL DEFAULT '', -- owner, admin, moderator, ''
+  member_status TEXT NOT NULL DEFAULT '', -- inv | con | full | off
+  contact_id INTEGER REFERENCES contacts ON DELETE RESTRICT,
```
efim-poberezkin, line 47: `NOT NULL`?

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+  invited_by INTEGER REFERENCES contacts ON DELETE RESTRICT,
+  connection_id BLOB UNIQUE,
```
efim-poberezkin, line 49: why duplicate? isn't it already available by `contact_id` from `contacts`?

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+  PRIMARY KEY (group_id, member_id)
+) WITHOUT ROWID;
+
+CREATE TABLE events ( -- messages received by the agent, append only
+  event_id INTEGER NOT NULL PRIMARY KEY,
```
efim-poberezkin, line 54: switch to manually reserved, incremented and set id? we'll have the same problem retrieving this id as we had for messages. particularly it's not reliable for concurrent use as far as I remember

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+  message_id INTEGER, -- REFERENCES messages ON DELETE CASCADE,
+  agent_id INTEGER NOT NULL, -- internal ID
+  external_agent_id INTEGER NOT NULL, -- internal ID
```
efim-poberezkin, line 57: not clear about comment, why `-- internal ID`?

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+  agent_meta TEXT NOT NULL, -- JSON with timestamps and IDs sent in MSG
+  connection_id BLOB NOT NULL,
```
efim-poberezkin, line 59: `contact_id`?

efim-poberezkin: PK should include this? (why do we need a global event_id?)

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+  received INTEGER NOT NULL, -- 1 for received, 0 for sent
+  event_type TEXT NOT NULL, -- event type - see protocol/types.ts
+  event_encoding INTEGER NOT NULL, -- format of event_body: 0 - binary, 1 - text utf8, 2 - JSON (utf8)
+  content_type TEXT NOT NULL, -- content type - see protocol/types.ts
+  event_body BLOB, -- agent message body as sent
+  created_at TEXT NOT NULL DEFAULT (datetime('now'))
+);
+
+CREATE TABLE group_events (
```
efim-poberezkin, line 68: what is the purpose of this table? why not merge it with `group_event_parents`?

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+);
+
+CREATE TABLE blobs (
+  blob_id INTEGER NOT NULL PRIMARY KEY,
+  content BLOB NOT NULL
+);
+
+CREATE TABLE messages ( -- mutable messages presented to user
+  message_id INTEGER NOT NULL PRIMARY KEY,
+  contact_id INTEGER REFERENCES contacts ON DELETE RESTRICT, -- NULL for sent messages
```
efim-poberezkin, line 92: also, why `RESTRICT`? the constraint would not allow anyway, why optimize this? why not `CASCADE`?

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+    REFERENCES group_members (group_id, member_id)
+);
+
+CREATE TABLE blobs (
+  blob_id INTEGER NOT NULL PRIMARY KEY,
+  content BLOB NOT NULL
+);
+
+CREATE TABLE messages ( -- mutable messages presented to user
+  message_id INTEGER NOT NULL PRIMARY KEY,
```
efim-poberezkin, line 91: same for id as for events - why global message_id? should we add `contact_id` to PK and make `message_id` manually incremented per contact?

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+  contact_id INTEGER REFERENCES contacts ON DELETE RESTRICT, -- NULL for sent messages
+  group_id INTEGER REFERENCES groups ON DELETE RESTRICT, -- only for group messages
+  deleted INTEGER NOT NULL, -- 1 for deleted
+  msg_type TEXT NOT NULL,
+  content_type TEXT NOT NULL,
+  msg_text TEXT NOT NULL, -- textual representation
+  msg_props TEXT NOT NULL, -- JSON
```
efim-poberezkin, line 98: make these nullable / move to a message_details table with a nullable reference so that when a message is deleted these are set to null?

migrations/20210612_initial.sql
```diff
@@ -0,0 +1,104 @@
...
+  msg_blob_id INTEGER REFERENCES blobs (blob_id) ON DELETE RESTRICT, -- optional binary content
```
efim-poberezkin, line 99: `ON DELETE SET NULL`?

efim-poberezkin (changes requested):
main questions regarding schema:

- relation between `profiles` and `contacts` is unclear.. maybe I'm not getting something and would need a clarification
- replace global ids with manually incremented ids per contact for `events` and `messages`
- remove default empty strings, remove default level 0 for connections
- unclear on `group_events` table and foreign keys to events
- clarify naming for `properties` and `display_name`
- need to discuss `ON DELETE` clauses

some other minor comments in separate comments

migrations/20210612_initial.sql
epoberezkin: > do we expect profile_id to be sqlite's "Integer Primary Key" in this case

I thought so. Not sure from comments what exactly I should do to make it unambiguous?

> I thought a profile is one of this user's profiles on this chat app

I thought to use it for the user as well, having multiple users, but then I removed users table - it's easier to simply have multiple db files (I think...). But the profile should still be stored in the db - we could have it in .ini and when there are changes compared with DB - update it and send updates to contacts... Not sure yet.

migrations/20210612_initial.sql
epoberezkin: are you saying primary key can't be NULL anyway, right?

migrations/20210612_initial.sql
epoberezkin: given that I removed users there is no distinction now between contact and profile, probably best to just merge the fields into one table. It's 1:1 relationship...

migrations/20210612_initial.sql
epoberezkin: I've made the same in introductions/invitations... we should change it there too then? If I read it into Text/ByteString  - will NULL be read as an empty string?

migrations/20210612_initial.sql
epoberezkin: less confusing like this

migrations/20210612_initial.sql
epoberezkin: > why is display_name shared here but local on contacts?

contacts have both local and shared (this one in profiles).

> does a group even need a local display name?

I don't think so. I am actually now thinking there should be a history of updates in the same way as for messages linked to events, both for groups and contacts profiles... Although we could add local group name to support anonymous groups without shared metadata, but there are shared member IDs still...

> maybe it could just be part of shared properties

I was planning to have it all in JSON in the protocol, but thought it's better to extract the main field into a separate record field - to simplify search, for example.

> properties -> shared_properties

will think... Maybe local_ instead...

epoberezkin: > whether group is muted or not

not sure what you mean

migrations/20210612_initial.sql
epoberezkin: needs to be reviewed when we get to the logic, not sure. You are probably right.

migrations/20210612_initial.sql
epoberezkin: that's actually on purpose to include the current user membership in the same table... Maybe current user should simply be present in contacts table with some boolean flag (or just ID = 0 or 1).

Initially I repeated all member properties for the current user in groups table, but it was messy - there are many fields

migrations/20210612_initial.sql
epoberezkin: no, it's another connection specifically for group messages.

We thought to combine at some point, but it's impossible to identify group then - it requires either global group ID or pairwise IDs for all member pairs - which is what connections provide already.

So it's a separate connection.

migrations/20210612_initial.sql
epoberezkin: wait... I thought it's ok to auto increment if it's in the same transaction?

migrations/20210612_initial.sql
epoberezkin: > not clear about comment, why -- internal ID?

it's the internal ID agent uses for messages

> why do we even need external_id here?

to send message receipts (ACK to the agent)

> if we do why can't it be part of agent_meta?

it could, but then you'd have to read and parse it - seems messier. Meta is mostly for archive atm.

migrations/20210612_initial.sql
epoberezkin: > contact_id? and REFERENCES contacts?

not really needed - connection_id references contact so it's just join. Also, contact can have multiple connections.

> PK should include this?

don't think so

> why do we need a global event_id?

we need to reference events from multiple places - check other tables - having more than one field for all event references is just messy a bit...

also message_id inside event_id is incorrect, as some events won't be shown as messages - e.g. group and profile updates (although they can be shown as messages)... We still need link tables between events and different types of entities...

migrations/20210612_initial.sql
epoberezkin: because group_event_parents is only needed for group events, not for direct connections/messages events. This is effectively a local, slightly processed copy of all agent messages in all connections (both received and sent). Unlike messages, this table is append only and it allows to see the history of all updates.

migrations/20210612_initial.sql
epoberezkin: same, to have single field references. What's the harm of having "global" IDs here? It's only per-user-profile global.

migrations/20210612_initial.sql
epoberezkin: same comment - TBC when we get to the logic.

migrations/20210612_initial.sql
epoberezkin: you lost me - why?

migrations/20210612_initial.sql
epoberezkin: same blob can be reused in multiple messages - e.g. when you forward a message you don't need a copy of the blob. It's one to many. RESTRICT is here to prevent blob deletion if it is used in any message.

protocol/types.ts
```diff
@@ -0,0 +1,71 @@
...
+  | "c.doc"
+  | "c.sticker"
+  | "c.file"
+  | "c.link"
+  | "c.form"
+  | "c.poll"
+  | "c.applet"
+
+// the type of message data transmitted inside SMP agent MSG
+interface MessageData<T extends GroupMessageType> {
```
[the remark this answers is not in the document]
epoberezkin: TBC - don't follow. It's all preliminary anyway, just a brain-dump.

migrations/20210612_initial.sql
efim-poberezkin: specifically INTEGER PRIMARY KEY, yes

migrations/20210612_initial.sql
efim-poberezkin: > to send message receipts (ACK to the agent)
can't we refer by internal_id for agent to figure out external_id to ack by?

migrations/20210612_initial.sql
efim-poberezkin: global IDs is artificial uniqueness, real uniqueness here is provided by `contact_id` and `message_id` - additionally this would mirror implementation in the agent

migrations/20210612_initial.sql
efim-poberezkin: https://sqlite.org/c3ref/last_insert_rowid.html

in transaction it should be fine probably, it's all shady in their documentation

```
For the purposes of this routine, an INSERT is considered to be successful even if it is subsequently rolled back.
...
If a separate thread performs a new INSERT on the same database connection while the sqlite3_last_insert_rowid() function is running and thus changes the last insert rowid, then the value returned by sqlite3_last_insert_rowid() is unpredictable and might not equal either the old or the new last insert rowid.
```

I guess we wouldn't have such problems, but I don't like relying on it. In agent I believe it allowed us to break a transaction

migrations/20210612_initial.sql
efim-poberezkin: I wouldn't mix user's profile with contacts' profiles at all - why not have a separate table "`user_data`" for that or like you offered a configuration file. The downside is it'd be the only record in the table. Then again user can have more configuration in his chat and his settings have to be stored somewhere

migrations/20210612_initial.sql
efim-poberezkin: you commented that it's a shared `properties` - I was saying group can additionally have `local_properties` in user's app - like whether to mute (silence notifications) the group for him or not, maybe some others

migrations/20210612_initial.sql
efim-poberezkin: >  If I read it into Text/ByteString - will NULL be read as an empty string?

not sure, I can make a test later - but you still have to treat either a NULL or an empty string somewhere. Having it as empty string only moves validation further from database

migrations/20210612_initial.sql
efim-poberezkin: you have `deleted` flag so I figure we're going to keep the record of the deleted message, not really delete it from database. If I'm deleting a message in my chat I'd prefer to think at least its contents are really deleted

migrations/20210612_initial.sql
efim-poberezkin: I think it's common approach to delete the blob altogether, and it becomes unavailable in other messages as well

migrations/20210612_initial.sql
```diff
@@ -1,44 +1,57 @@
 CREATE TABLE contact_profiles ( -- remote user profile
```
efim-poberezkin, line 1: just `profiles`, since local user also stores his profile here?

migrations/20210612_initial.sql
```diff
@@ -1,44 +1,57 @@
...
 
 CREATE TABLE contacts (
   contact_id INTEGER PRIMARY KEY,
-  contact_ref TEXT NOT NULL UNIQUE, -- contact name set by local user - must be unique
-  properties TEXT NOT NULL DEFAULT '{}', -- JSON set by local user
-  contact_profile_id INTEGER REFERENCES contact_profiles, -- sent by remote contact
-  contact_status TEXT NOT NULL DEFAULT ''
+  local_contact_ref TEXT NOT NULL UNIQUE, -- contact name set by local user - must be unique
+  local_properties TEXT NOT NULL DEFAULT '{}', -- JSON set by local user
+  contact_profile_id INTEGER UNIQUE REFERENCES contact_profiles, -- profile sent by remote contact, NULL for incognito contacts
```
efim-poberezkin, line 32: why not store nullable `contact_ref` and `properties` on `contacts` directly, and store user profiles (`properties`) in `users` table?

migrations/20210612_initial.sql
```diff
@@ -51,48 +64,49 @@ CREATE TABLE group_profiles ( -- shared group profiles
...
+  agent_msg_id INTEGER NOT NULL, -- internal message ID
+  external_msg_id INTEGER NOT NULL, -- external message ID (sent or received)
+  agent_meta TEXT NOT NULL, -- JSON with timestamps etc. sent in MSG
+  connection_id INTEGER NOT NULL REFERENCES connections,
+  received INTEGER NOT NULL, -- 0 for received, 1 for sent
   event_type TEXT NOT NULL, -- event type - see protocol/types.ts
   event_encoding INTEGER NOT NULL, -- format of event_body: 0 - binary, 1 - text utf8, 2 - JSON (utf8)
   content_type TEXT NOT NULL, -- content type - see protocol/types.ts
   event_body BLOB, -- agent message body as sent
+  group_member_id INTEGER REFERENCES group_members, -- can be NULL for direct messages
```
efim-poberezkin, line 104: what was the reason to remove a separate table for group events?

efim-poberezkin (approved):

PR #67, opened 2021-06-30
update ghc version to 8.10.4 for Docker build

efim-poberezkin:


epoberezkin:
This makes GHC version different from stack resolver version (which is 8.10.3). I was going to migrate to 8.10.4 in v4, we can do it in master too.

efim-poberezkin:
> This makes GHC version different from stack resolver version (which is 8.10.3). I was going to migrate to 8.10.4 in v4, we can do it in master too.

looking up our resolver it does correspond to ghc-8.10.4, doesn't it?
https://www.stackage.org/lts-17.12

PR #69, opened 2021-07-04
establish connection using user profiles

epoberezkin:


apps/simplex-chat/Simplex/Chat.hs
```diff
@@ -145,12 +148,19 @@ chatSubscriber = do
                   setActive $ ActiveC c
                 _ -> pure ()
             _ -> pure ()
-      AgentTransmission {agentConnId, chatDirection = ReceivedDirectMessage Contact {localContactRef = c}, agentMessage} ->
+      AgentTransmission {agentConnId, chatDirection = ReceivedDirectMessage NewContact {activeConn}, agentMessage} ->
+        void . runExceptT $ case agentMessage of
+          CONF confId connInfo -> do
+            -- TODO update connection status
+            saveContact userId activeConn connInfo
```
efim-poberezkin, line 155: this could be done in separate scope, but maybe add primitive user dialogue asking to accept counterparty by profile?

apps/simplex-chat/Simplex/Chat.hs
```diff
@@ -160,6 +170,68 @@ chatSubscriber = do
             unsetActive $ ActiveC c
           _ -> pure ()
       _ -> pure ()
+  where
+    saveContact userId activeConn connInfo = do
```
efim-poberezkin, line 174: would this not shadow `connInfo` from line 153?

apps/simplex-chat/Simplex/Store.hs
```diff
@@ -55,33 +62,122 @@ migrations =
...
+    create _ _ _ 0 = pure $ Left SEDuplicateContactRef
+    create db profileId lcrSuffix attempts = do
+      let lcr = localContactRef' lcrSuffix
+      E.try (insertUser lcr) >>= \case
+        Right () -> do
+          contactId <- insertedRowId db
+          DB.execute db "UPDATE connections SET contact_id = ? WHERE connection_id = ?" (contactId, connId)
+          pure $ Right ()
+        Left e
+          | DB.sqlError e == DB.ErrorConstraint -> create db profileId (lcrSuffix + 1) (attempts - 1)
```
efim-poberezkin, line 157: isn't this error handling obsolete due to max suffix already being retrieved during the same transaction?

apps/simplex-chat/Simplex/Store.hs
```diff
@@ -55,33 +62,122 @@ migrations =
...
+            [sql|
+              INSERT INTO contacts
+                (contact_profile_id, local_contact_ref, lcr_base, lcr_suffix, user_id) VALUES (?, ?, ?, ?, ?)
+            |]
+            (profileId, lcr, contactRef, lcrSuffix, userId)
 
 deleteContact :: MonadUnliftIO m => SQLiteStore -> UserId -> ContactRef -> m ()
 deleteContact st userId contactRef =
   liftIO . withTransaction st $ \db ->
     forM_
```
efim-poberezkin, line 174: I think this method would be better off with separate queries instead of forM_ - it saves reuse of a couple of lines but slightly complicates reading.. but I won't insist

apps/simplex-chat/Simplex/Chat.hs
```diff
@@ -160,6 +170,68 @@ chatSubscriber = do
...
+    getUsers st >>= \case
+      [] -> newUser
+      users -> maybe (selectUser users) pure (find activeUser users)
+  putStrLn $ "Current user: " <> userStr user
+  pure user
+  where
+    newUser :: IO User
+    newUser = do
+      putStrLn
+        "No user profiles found, it will be created now.\n\
```
efim-poberezkin, line 190: `No user profiles` -> `No user profile`?

apps/simplex-chat/Simplex/Chat.hs
```diff
@@ -160,6 +170,68 @@ chatSubscriber = do
...
+        \Please choose your alias and your profile name.\n\
+        \They will be sent to your contacts when you connect.\n\
+        \They are only stored on your device and you can change them later."
+      loop
+      where
+        loop = do
+          contactRef <- getContactRef
+          displayName <- T.pack <$> getWithPrompt "profile name (optional)"
+          liftIO (runExceptT $ createUser st Profile {contactRef, displayName} True) >>= \case
+            Left SEDuplicateContactRef -> do
```
efim-poberezkin, line 200: how can any contact exist before creation of the first user?

apps/simplex-chat/Simplex/Chat.hs
```diff
@@ -160,6 +170,68 @@ chatSubscriber = do
...
+              putStrLn "chosen alias already used by another profile on this device, choose another one"
+              loop
+            Left e -> putStrLn ("database error " <> show e) >> exitFailure
+            Right user -> pure user
+    selectUser :: [User] -> IO User
+    selectUser [user] = do
+      liftIO $ setActiveUser st (userId user)
+      pure user
+    selectUser users = do
+      putStrLn "Select user profile: "
```
efim-poberezkin, line 210: remove space in the end?

migrations/20210612_initial.sql
```diff
@@ -1,21 +1,29 @@
...
--- the first record (id = 1) is reserved for the first local user
-INSERT INTO contact_profiles (contact_profile_id, contact_ref) VALUES (1, '');
-
-
 CREATE TABLE users (
   user_id INTEGER PRIMARY KEY,
-  contact_profile_id INTEGER NOT NULL UNIQUE REFERENCES contact_profiles -- user's profile
+  contact_id INTEGER NOT NULL UNIQUE REFERENCES contacts ON DELETE CASCADE
+    DEFERRABLE INITIALLY DEFERRED,
+  active_user INTEGER -- 1 for active user
```
efim-poberezkin, line 12: would like to discuss, possibly offline - why at all store `active_user` - only for state between sessions? Other than that `currentUser` in chat controller should be enough, shouldn't it? `selectUser` routine would just be non optional every session start, and it would remove some complexity

migrations/20210612_initial.sql
```diff
@@ -1,21 +1,29 @@
...
+CREATE TABLE contacts (
+  contact_id INTEGER PRIMARY KEY,
+  contact_profile_id INTEGER UNIQUE REFERENCES contact_profiles, -- NULL if it's an incognito profile
+  local_contact_ref TEXT NOT NULL,
+  lcr_base TEXT NOT NULL,
+  lcr_suffix INTEGER NOT NULL DEFAULT 0,
+  user_id INTEGER NOT NULL REFERENCES users,
+  user INTEGER, -- 1 if this contact is a user
+  created_at TEXT NOT NULL DEFAULT (datetime('now')),
+  UNIQUE (user_id, local_contact_ref) ON CONFLICT FAIL,
```
efim-poberezkin, line 24: what does `ON CONFLICT FAIL` achieve here? reading up on it, it seems somewhat destructive

efim-poberezkin (approved):

apps/simplex-chat/Simplex/Chat.hs
epoberezkin: yes, possibly later, after protocol is stabilised. Also atm we don't have UX support for such interactions, it would conflict with how we process input as it.

apps/simplex-chat/Simplex/Chat.hs
epoberezkin: No, Haskell only has function parameters in scope, not the variables that were defined in the function body.

apps/simplex-chat/Simplex/Chat.hs
epoberezkin: it's the logic to allow adding profiles later using the same function.

apps/simplex-chat/Simplex/Chat.hs
epoberezkin: yes, this one is not needed

apps/simplex-chat/Simplex/Store.hs
epoberezkin: yes, probably, not helping much

apps/simplex-chat/Simplex/Store.hs
epoberezkin: no, because there can be remote profiles `john` (with base `john`) and `john_1` (with the base `john_1`) and when you increment the first it would clash with the second.

The alternative (and probably more reliable) solution would be to recognise this pattern and for contacts that end with `name_number` always save `name` as the base and `number` as the suffix. The problem with that approach though is that if one creates an alias `john_1941` then the contact that would be created after `john` would be not `john_1` but `john_1942` which is weird.

So another approach would be not to find max, but to find the minimum gap, but I didn't like the complexity of it so went with the simple retry approach that should be ok in realistic cases (unless somebody tries to purposefully exploit it)...

migrations/20210612_initial.sql
epoberezkin: it allows to stay within the transaction when name clash without rolling back the previous changes - the retry loop that increments the suffix relies on it

epoberezkin: I didn't test how it works though yet

PR #72, opened 2021-07-06
chat test with VirtualTerminal

epoberezkin:


src/Simplex/Chat/Terminal.hs
```diff
@@ -1,20 +1,26 @@
...
 {-# LANGUAGE ScopedTypeVariables #-}
 
 module Simplex.Chat.Terminal where
 
+import Control.Monad.Catch (MonadMask)
+import Control.Monad.IO.Class (MonadIO)
 import Simplex.Chat.Styled
 import Simplex.Chat.Types
 import System.Console.ANSI.Types
 import System.Terminal
```
efim-poberezkin, line 14: import qualified? had to jump back and forth between github and ide to figure out what comes from where

tests/ChatTests.hs
```diff
@@ -0,0 +1,74 @@
...
+
+aliceProfile :: Profile
+aliceProfile = Profile {contactRef = "alice", displayName = "Alice"}
+
+bobProfile :: Profile
+bobProfile = Profile {contactRef = "bob", displayName = "Bob"}
+
+testAddContact :: Spec
+testAddContact = describe "add chat contact" $
+  xit "add contact and send/receive message" $
```
efim-poberezkin, line 24: `it`

efim-poberezkin (approved):

tests/ChatTests.hs
epoberezkin: it's disabled on purpose, it requires that the server is running and I can't easily start it in CI - it's in a different codebase, so some extra dance in CI is needed to run it here - not planning to do it right now.

src/Simplex/Chat/Terminal.hs
epoberezkin: maybe - will have a look - it's not this PR specifically that added it...

PR #74, opened 2021-07-08
simplify chat protocol

epoberezkin:


migrations/20210612_initial.sql
```diff
@@ -54,7 +46,6 @@ CREATE TABLE groups (
   invited_by INTEGER REFERENCES contacts ON DELETE RESTRICT,
   external_group_id BLOB NOT NULL,
   local_group_ref TEXT NOT NULL UNIQUE, -- local group name without spaces
-  local_properties TEXT NOT NULL, -- local JSON group properties
```
efim-poberezkin, line 57: why remove this?

src/Simplex/Chat.hs
```diff
@@ -314,10 +323,23 @@ withStore action = do
 chatCommandP :: Parser ChatCommand
 chatCommandP =
   ("/help" <|> "/h") $> ChatHelp
+    <|> ("/group #" <|> "/g #") *> (NewGroup <$> groupRef)
+    <|> ("/add #" <|> "/a #") *> (AddMember <$> groupRef <* A.space <*> contactRef <* A.space <*> memberRole)
+    <|> ("/remove #" <|> "/rm #") *> (RemoveMember <$> groupRef <* A.space <*> contactRef)
+    <|> ("/delete #" <|> "/d #") *> (DeleteGroup <$> groupRef)
+    <|> ("/members #" <|> "/ms #") *> (ListMembers <$> groupRef)
```
efim-poberezkin, line 330: not a critique, just a thought - maybe we could have a subcommand for group commands (e.g. instead of "#" - `/group new`, `/group add`, `/group remove`, etc.), so that it's harder to mix up groups and regular contacts namespaces? also maybe move group ChatCommands under GroupCommand which is also a ChatCommand but has different constructors of its own to somewhat separate group commands on types level as well? I'm not sure of benefits other than clearer distinction between group and contact commands on types and parsers, but I just felt it starts to fill somewhat crowded in ChatCommands and probably will only become more so. For example if we later introduce something like "Service" or "Configuration commands" letting client to control app - they'd also be mixed into the same ChatCommands namespace. Instead we could have separate namespaces for them, groups, actually regular contacts as well, and maybe others

src/Simplex/Chat.hs
```diff
@@ -314,10 +323,23 @@ withStore action = do
...
     <|> ("/markdown" <|> "/m") $> MarkdownHelp
   where
-    contactRef = safeDecodeUtf8 <$> A.takeTill (== ' ')
+    contactRef = safeDecodeUtf8 <$> (B.cons <$> A.satisfy refChar <*> A.takeTill (== ' '))
+    refChar c = c > ' ' && c /= '#' && c /= '@'
+    groupRef = contactRef
+    memberRole =
+      ("owner" $> GROwner)
+        <|> ("admin" $> GRAdmin)
+        <|> ("normal" $> GRMember)
```
efim-poberezkin, line 344: change "normal" -> "member" like suggested by schema comment, or vice versa?

src/Simplex/Chat/protocol.md
```diff
@@ -2,26 +2,25 @@
...
   - allow additional content types namespaces
 
 ## Message syntax
 
 The syntax of the message inside agent MSG:
 
 ```abnf
-agentMessageBody = message / msgContinuation
-message = [chatMsgId] SP msgEvent SP [parameters] SP [contentParts [SP msgBodyParts]]
+agentMessageBody = [chatMsgId] SP msgEvent SP [parameters] SP [contentParts [SP msgBodyParts]]
```
efim-poberezkin, line 23: `msgEvent SP [parameters]` is kind of free-form since according to protocol implementation optionality of parameters depends on specific msgEvent[Type], but it's ok for now I guess

efim-poberezkin (approved):

src/Simplex/Chat.hs
epoberezkin: will think about it... not needed at the moment

src/Simplex/Chat.hs
epoberezkin: in the next PR already

migrations/20210612_initial.sql
epoberezkin: we don't even have the use for profile properties yet, don't think we need these at all.

src/Simplex/Chat/protocol.md
epoberezkin: this is the general syntax for messages that parse to RawChatMessage, your comment is about semantic validity that is not defined in this syntax (and does not need to be).

PR #77, opened 2021-07-12
chat groups: establish connection between host and invitee members

epoberezkin:


src/Simplex/Chat.hs
```diff
@@ -219,11 +223,52 @@ processAgentMessage User {userId, profile} agentConnId agentMessage = do
...
+                  pure ()
+                else do
+                  showConnectedGroupMember gName $ displayName (memberProfile m :: Profile)
+        MSG meta msgBody -> do
+          ChatMessage {chatMsgEvent} <- liftEither $ parseChatMessage msgBody
+          case chatMsgEvent of
+            XGrpMemNew memId memRole memProfile -> do
+              Group {membership, members} <- withStore $ \st -> getGroup st user gName
+              when (memberId membership /= memId && isNothing (find ((== memId) . memberId . fst) members)) $
+                withStore $ \st -> pure () -- add new member as GSMemAccepted
```
efim-poberezkin, line 269: you're probably doing it in next pr, but I'll comment here just in case - we should check for status of the inviter here (whether he's at least admin), because check on the invitee part is not enough since both he and the inviter can be compromised / faulty

src/Simplex/Chat.hs
```diff
@@ -234,10 +279,12 @@ processAgentMessage User {userId, profile} agentConnId agentMessage = do
         setActive $ ActiveC c
       _ -> pure ()
 
-    groupInvitation :: Contact -> (MemberId, GroupMemberRole) -> (MemberId, GroupMemberRole) -> SMPQueueInfo -> GroupProfile -> m ()
-    groupInvitation _ct (fromMemId, fromRole) (memId, memRole) _qInfo _groupProfile = do
+    saveGroupInvitation :: Contact -> GroupInvitation -> m ()
+    saveGroupInvitation ct@Contact {localDisplayName} inv@(GroupInvitation (fromMemId, fromRole) (memId, memRole) _ _) = do
       when (fromRole < GRAdmin || fromRole < memRole) $ throwError $ ChatError CEGroupRole
```
efim-poberezkin, line 284: and here it might be strange to receive such error as its name implies that it user's role that is insufficient.. maybe have a separate error - e.g. "Non Admin Invitation"

src/Simplex/Chat/Help.hs
```diff
@@ -11,27 +11,26 @@ chatHelpInfo :: [StyledString]
...
       "",
-      Markdown (Colored Green) "Step 1: " <> Markdown (Colored Cyan) "/add bob" <> " -- Alice adds her contact, Bob (she can use any name).",
+      Markdown (Colored Green) "Step 1: " <> highlight "/add" <> " -- Alice adds a contact.",
       indent <> "Alice should send the invitation printed by the /add command",
       indent <> "to her contact, Bob, out-of-band, via any trusted channel.",
       "",
-      Markdown (Colored Green) "Step 2: " <> Markdown (Colored Cyan) "/connect alice <invitation>" <> " -- Bob accepts the invitation.",
-      indent <> "Bob also can use any name for his contact, Alice,",
-      indent <> "followed by the invitation he received out-of-band.",
+      Markdown (Colored Green) "Step 2: " <> highlight "/connect <invitation>" <> " -- Bob accepts the invitation.",
```
efim-poberezkin, line 21: re-reading these steps and it's kind of an awkward transition from reader - "Follow these steps" - to Alice and Bob. Not significant though, maybe add a TODO for another pr

src/Simplex/Chat/Store.hs
```diff
@@ -380,81 +394,99 @@ getGroup st User {userId, userContactId} localDisplayName =
...
-            JOIN groups g ON g.group_id = m.group_id
             JOIN contact_profiles p ON p.contact_profile_id = m.contact_profile_id
-            JOIN connections c ON c.group_member_id = m.group_member_id
-            WHERE g.group_id = ?
-            ORDER BY c.connection_id DESC
-            LIMIT 1
+            LEFT JOIN connections c ON c.connection_id = (
+              SELECT max(cc.connection_id)
+              FROM connections cc
+              where cc.group_member_id = c.group_member_id
```
efim-poberezkin, line 401: shouldn't it be "where cc.group_member_id = **m**.group_member_id"? or it doesn't matter..? - got a little confused

efim-poberezkin: at least we were definitely verifying the "m"

src/Simplex/Chat/Store.hs
```diff
@@ -380,81 +394,99 @@ getGroup st User {userId, userContactId} localDisplayName =
...
+      let (b, a) = break ((== Just userContactId) . memberContactId . fst) allMembers
+       in case a of
+            [] -> Left SEGroupWithoutUser
+            u : ms -> Right (b <> ms, fst u)
+
+getGroupInvitation :: StoreMonad m => SQLiteStore -> User -> GroupName -> m ReceivedGroupInvitation
+getGroupInvitation st user localDisplayName =
+  liftIOEither . withTransaction st $ \db -> runExceptT $ do
+    (Group {membership, members, groupProfile}, qInfo) <- getGroup_ db user localDisplayName
+    when (memberStatus membership /= GSMemInvited) $ throwError SEGroupAlreadyJoined
```
efim-poberezkin, line 419: this probably should be checked on chat level, not store - it's only a get method

src/Simplex/Chat/Store.hs
```diff
@@ -380,81 +394,99 @@ getGroup st User {userId, userContactId} localDisplayName =
...
-    [ ":group_id" := groupId,
-      ":member_id" := memberId,
-      ":member_role" := memberRole,
-      ":member_status" := memberStatus,
-      ":invited_by" := invitedBy,
-      ":contact_id" := contactId
-    ]
+    (userId, agentConnId, ConnNew, ConnMember, groupMemberId)
+
+createContactMember_ :: IsContact a => DB.Connection -> User -> Int64 -> a -> MemberInfo -> GroupMemberStatus -> InvitedBy -> IO GroupMember
```
efim-poberezkin, line 463: nitpick, but please move this right after `createGroupMember`? it's strangely separated from the last non utility method that uses it

src/Simplex/Chat/Store.hs
```diff
@@ -380,81 +394,99 @@ getGroup st User {userId, userContactId} localDisplayName =
...
+createMemberConnection st userId groupMemberId agentConnId =
+  liftIO . withTransaction st $ \db -> createMemberConnection_ db userId groupMemberId agentConnId
 
-createMember_ :: DB.Connection -> Int64 -> Int64 -> GroupMemberRole -> GroupMemberStatus -> Maybe Int64 -> ByteString -> IO ()
-createMember_ db groupId contactId memberRole memberStatus invitedBy memberId =
-  DB.executeNamed
+updateGroupMemberStatus :: MonadUnliftIO m => SQLiteStore -> UserId -> Int64 -> GroupMemberStatus -> m ()
+updateGroupMemberStatus _st _userId _groupMemberId _memberStatus = pure ()
+
+createMemberConnection_ :: DB.Connection -> UserId -> Int64 -> ConnId -> IO ()
```
efim-poberezkin, line 453: same here, move it above `update`?

src/Simplex/Chat/Types.hs
```diff
@@ -77,6 +89,24 @@ instance ToJSON GroupProfile where toEncoding = J.genericToEncoding J.defaultOpt
...
 
+data GroupInvitation = GroupInvitation
+  { fromMember :: MemberInfo,
+    invitedMember :: MemberInfo,
+    queueInfo :: SMPQueueInfo,
+    groupProfile :: GroupProfile
+  }
+  deriving (Eq, Show)
+
+data ReceivedGroupInvitation = ReceivedGroupInvitation
```
efim-poberezkin, line 100: why not newtype ReceivedGroupInvitation? these two seem the same

tests/ChatTests.hs
```diff
@@ -37,30 +44,55 @@ testAddContact = describe "add chat contact" $
...
       alice #> "@bob_1 hello"
       bob <# "alice_1> hello"
       bob #> "@alice_1 hi"
       alice <# "bob_1> hi"
       -- test deleting contact
       alice ##> "/d bob_1"
       alice <## "bob_1 is deleted"
-      chatCommand alice "@bob_1 hey"
+      alice #:> "@bob_1 hey"
       alice <## "no contact bob_1"
```
efim-poberezkin, line 57: might be an overkill, but maybe send one more message to `@bob` to verify he wasn't deleted?

tests/ChatTests.hs
```diff
@@ -37,30 +44,55 @@ testAddContact = describe "add chat contact" $
...
+      alice #:> "/g #team"
+      -- TODO this occasionally fails in case getWindow is run before the command above is printed
+      alice <## "use /a #team <name> to add members"
+      alice ##> "/a #team bob admin"
+      alice <## "invitation to join the group #team sent to bob"
+      bob <## "use /j #team to accept"
+      bob ##> "/j #team"
+      concurrently_
+        (alice <## "bob joined the group #team")
+        (bob <## "you joined the group #team")
```
efim-poberezkin, line 73: very cool

src/Simplex/Chat.hs
efim-poberezkin: the description `"insufficient role for this group command"` also clearly implies it's user's permissions that are lacking

efim-poberezkin (changes requested):
minor fixes and refactoring

src/Simplex/Chat/Store.hs
epoberezkin: I did it on purpose to not depend on the decision that group invitations are stored as groups - internally a different record is used. Hence the check here.

src/Simplex/Chat/Types.hs
epoberezkin: newtype can only be used when there one constructor and it has exactly one field - that is the type can have the same internal representation as said field, and therefore can be completely erased and not create an extra wrapper. Here we have many fields.

tests/ChatTests.hs
epoberezkin: it is indeed :)

tests/ChatTests.hs
epoberezkin: it's just one connection so far...

src/Simplex/Chat/Store.hs
epoberezkin: the idea was to have all non-exported methods in the bottom together

src/Simplex/Chat/Store.hs
epoberezkin: same here

src/Simplex/Chat/Store.hs
epoberezkin: so it was wrong to remove extra equality - it was `c.group_member_id = m.group_member_id` I think

how it even works?

src/Simplex/Chat.hs
epoberezkin: CEGroupHostRole ?

src/Simplex/Chat/Store.hs
efim-poberezkin: okay

src/Simplex/Chat.hs
epoberezkin: yes - this is not complete

PR #78, opened 2021-07-15
core chat groups protocol for adding members

epoberezkin:


src/Simplex/Chat.hs
```diff
@@ -231,48 +231,120 @@ processAgentMessage user@User {userId, profile} agentConnId agentMessage = do
...
-              if Just (invitedBy membership) == (IBContact <$> memberContactId m)
-                then do
-                  -- sender invited the current user
-                  withStore $ \st -> updateGroupMemberStatus st userId (groupMemberId m) GSMemConnected
-                  showUserConnectedToGroup gName
-                  pure ()
-                else do
-                  showConnectedGroupMember gName $ displayName (memberProfile m :: Profile)
+              mIds <- withStore $ \st -> createIntroductions st m
+              sendGroupMessage members . (`XGrpMemNew` mIds) $ memberInfo m
```
efim-poberezkin, line 258: TODO update status to "announced"

src/Simplex/Chat.hs
```diff
@@ -231,48 +231,120 @@ processAgentMessage user@User {userId, profile} agentConnId agentMessage = do
...
-            XGrpMemNew memId memRole memProfile -> do
-              Group {membership, members} <- withStore $ \st -> getGroup st user gName
-              when (memberId membership /= memId && isNothing (find ((== memId) . memberId . fst) members)) $
-                withStore $ \st -> pure () -- add new member as GSMemAccepted
+            XGrpMemNew memInfo@(MemberInfo memId _ _) _mIds -> do
+              group@Group {membership} <- withStore $ \st -> getGroup st user gName
+              case memberCategory m of
+                GCHostMember -> do
+                  -- TODO add listed members to some list of expected members - group_members table is not very good as it will make many fields nullable
+                  -- TODO maybe we should not set user to GSMemAnnounced to avoid inconsistency when for other member it means not connected yet...
```
efim-poberezkin, line 284: it wouldn't mean user is connected to other members for the user himself as well.. I think it's okay to assign this status here

src/Simplex/Chat.hs
```diff
@@ -231,48 +231,120 @@ processAgentMessage user@User {userId, profile} agentConnId agentMessage = do
...
+                  -- (currently user is GSMemConnected)
+                  when (memberId membership == memId) $
```
efim-poberezkin, line 286: isn't this check semantically duplicates the category pattern match that the message is from Host to member?

efim-poberezkin: or it's in case Host invites another one I guess

src/Simplex/Chat.hs
epoberezkin: I decided to use announced only for members that are not connected yet (that is neither host nor invitee)

PR #86, opened 2021-07-25
merge profiles using contact probe

epoberezkin:


migrations/20210612_initial.sql
```diff
@@ -43,6 +45,30 @@ CREATE TABLE contacts (
...
+  user_id INTEGER NOT NULL REFERENCES users,
+  UNIQUE (user_id, probe)
+);
+
+CREATE TABLE sent_probe_hashes (
+  sent_probe_hash_id INTEGER PRIMARY KEY,
+  sent_probe_id INTEGER NOT NULL REFERENCES sent_probes ON DELETE CASCADE,
+  contact_id INTEGER NOT NULL REFERENCES contacts ON DELETE CASCADE,
+  user_id INTEGER NOT NULL REFERENCES users,
+  UNIQUE (sent_probe_id, contact_id)
```
efim-poberezkin, line 61: would `UNIQUE (sent_probe_id, user_id)` be redundant?

src/Simplex/Chat/Store.hs
```diff
@@ -293,6 +302,126 @@ toMaybeConnection (Just connId, Just agentConnId, Just connLevel, viaContact, Ju
   Just $ toConnection (connId, agentConnId, connLevel, viaContact, connStatus, connType, contactId, groupMemberId, createdAt)
 toMaybeConnection _ = Nothing
 
+getExistingContacts :: MonadUnliftIO m => SQLiteStore -> UserId -> Contact -> m [Contact]
```
efim-poberezkin, line 305: getMatchingContacts?

efim-poberezkin (approved):

migrations/20210612_initial.sql
epoberezkin: I think so, but we could make foreign keys include user_id as we spoke... It's a separate change though.

PR #87, opened 2021-07-27
group commands (remove member, leave group, delete group)

epoberezkin:
- [x] remove member
- [x] leave group
- [x] re-add member that left of was removed
- [x] delete group while user is in the group (other members will delete it too)
- [x] delete group after user is removed or left the group (it will only be deleted locally)
- [x] prevent deletion of contact that is a group member

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - v4
     tags:
       - "v*"
   pull_request:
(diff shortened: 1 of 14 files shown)


ixzh:
are these command available? I'm looking for a way to rename a group, but -h doesn't have any group related command except fot /g to create one.

Can I have group name with space？ like `cosmic\ bebop` becomes `cosmic\`  :(

PR #94, opened 2021-08-23
send files to contacts

epoberezkin:


src/Simplex/Chat.hs
```diff
@@ -260,6 +302,26 @@ processChatCommand user@User {userId, profile} = \case
     contactMember Contact {contactId} =
       find $ \GroupMember {memberContactId = cId, memberStatus = s} ->
         cId == Just contactId && s /= GSMemRemoved && s /= GSMemLeft
+    getRcvFilePath :: Maybe FilePath -> String -> m FilePath
+    getRcvFilePath filePath fileName = case filePath of
+      Nothing -> getTemporaryDirectory >>= (`uniqueCombine` fileName)
```
efim-poberezkin, line 307: make `<user_dir>/Downloads/simplex` default instead of `/tmp`?

src/Simplex/Chat/Store.hs
```diff
@@ -334,9 +351,58 @@ getContact_ db userId localDisplayName = do
...
+            SELECT DISTINCT f.file_id
+            FROM files f
+            JOIN snd_files s
+            WHERE f.user_id = ? AND s.file_status IN (?, ?, ?)
+          |]
+          (userId, FSNew, FSAccepted, FSConnected)
+    concatMap (filter liveTransfer) . rights <$> mapM (getSndFileTransfers_ db userId) fileIds
+  where
+    liveTransfer :: SndFileTransfer -> Bool
+    liveTransfer SndFileTransfer {fileStatus} = fileStatus `elem` [FSNew, FSAccepted, FSConnected]
```
efim-poberezkin, line 374: don't you filter twice this way?

epoberezkin: Not really, because the first filter filters the file IDs that have some active transfers (there are multiple transfers per ID), and the second filter out live transfers.
The alternative would be to just write a completely new query and not use the function.

PR #95, opened 2021-08-28
subscribe pending connections on chat start

efim-poberezkin:


src/Simplex/Chat.hs
```diff
@@ -276,9 +276,14 @@ agentSubscriber = do
 subscribeUserConnections :: (MonadUnliftIO m, MonadReader ChatController m) => m ()
 subscribeUserConnections = void . runExceptT $ do
   user <- readTVarIO =<< asks currentUser
+  subscribeNonActivatedConnections user
```
epoberezkin, line 279: Pending?

epoberezkin: also - let's put it in the end, so it subscribes contacts first (kind of starts faster a bit :)

src/Simplex/Chat.hs
```diff
@@ -276,9 +276,14 @@ agentSubscriber = do
...
   subscribeContacts user
   subscribeGroups user
   where
+    subscribeNonActivatedConnections user = do
+      connections <- withStore (`getNullContactConnections` user)
```
epoberezkin, line 284: Pending

src/Simplex/Chat.hs
```diff
@@ -276,9 +276,14 @@ agentSubscriber = do
...
+      forM_ connections $ \conn ->
```
epoberezkin, line 285: You can also have `Connection {agentConnId}`

epoberezkin: and you need `catchError` as below, even if the handler does nothing (I don't think there is a meaningful error message here)

src/Simplex/Chat/Store.hs
```diff
@@ -336,6 +337,24 @@ getUserContacts st User {userId} =
     contactNames <- liftIO $ map fromOnly <$> DB.query db "SELECT local_display_name FROM contacts WHERE user_id = ?" (Only userId)
     rights <$> mapM (runExceptT . getContact_ db userId) contactNames
 
+getNullContactConnections :: StoreMonad m => SQLiteStore -> User -> m [Connection]
```
epoberezkin, line 340: It can be just MonadUnliftIO - it doesn't fail

src/Simplex/Chat/Store.hs
```diff
@@ -336,6 +337,24 @@ getUserContacts st User {userId} =
...
+getNullContactConnections st User {userId} =
+  liftIOEither . withTransaction st $ \db ->
```
epoberezkin, line 342: ... and just liftIO here

src/Simplex/Chat/Store.hs
```diff
@@ -336,6 +337,24 @@ getUserContacts st User {userId} =
...
+          SELECT c.connection_id, c.agent_conn_id, c.conn_level, c.via_contact,
+            c.conn_status, c.conn_type, c.contact_id, c.group_member_id, c.created_at
+          FROM connections c
+          WHERE c.user_id = :user_id
+            AND c.conn_type = 'contact'
+            AND c.contact_id IS NULL
+        |]
+        [":user_id" := userId]
+  where
+    connections rows = Right $ map toConnection rows
```
epoberezkin, line 356: ... and you won't need Right here, as it is never Left - so you can just put `map toConnection` above without creating this function.

epoberezkin (changes requested):

src/Simplex/Chat.hs
```diff
@@ -296,6 +297,10 @@ subscribeUserConnections = void . runExceptT $ do
             forM_ connectedMembers $ \(GroupMember {localDisplayName = c}, cId) ->
               subscribe cId `catchError` showMemberSubError g c
             showGroupSubscribed g
+    subscribePendingConnections user = do
+      connections <- withStore (`getPendingConnections` user)
+      forM_ connections $ \Connection {agentConnId} ->
+        subscribe agentConnId `catchError` \_ -> liftIO $ pure ()
```
epoberezkin, line 303: you don't need liftIO here

src/Simplex/Chat/Store.hs
```diff
@@ -336,6 +337,22 @@ getUserContacts st User {userId} =
...
+  liftIO . withTransaction st $ \db ->
+    map toConnection
+      <$> DB.queryNamed
+        db
+        [sql|
+          SELECT c.connection_id, c.agent_conn_id, c.conn_level, c.via_contact,
+            c.conn_status, c.conn_type, c.contact_id, c.group_member_id, c.created_at
+          FROM connections c
+          WHERE c.user_id = :user_id
+            AND c.conn_type = 'contact'
```
epoberezkin, line 351: better to pass it as a parameter (ConnContact), not to use the mapping in two places

epoberezkin (approved):

PR #127, opened 2021-11-27
:sparkles: add script installer

UltiRequiem:


install.sh
```diff
@@ -0,0 +1,48 @@
...
+	FILE_TO_EXPORT=".bashrc"
+else
+	echo "Your shell is not suported, try with bash/zsh."
+	exit 1
+fi
+######################
+
+# Check if the directory exists
+SIMPLEX_CHAT_DIR="$HOME/.config/$SIMPLEX_CHAT"
+[ ! -d $SIMPLEX_CHAT_DIR ] && mkdir -p $SIMPLEX_CHAT_DIR
```
efim-poberezkin, line 36: Hi! Thanks for the PR, a couple of comments from my side.

I think we should switch this to some `bin` folder - you're using it later as a target for downloading the binary, not storing the app's configuration. We currently suggest `~/.local/bin` in the README, and after a brief research into unix directory structure it looks okay, but if you would suggest a better alternative it would be fine as well.

```suggestion
TARGET_DIR="$HOME/.local/bin"
[ ! -d $TARGET_DIR ] && mkdir -p $TARGET_DIR
```

install.sh
```diff
@@ -0,0 +1,48 @@
+# Constants
+VERSION="v0.4.2"
+SIMPLEX_CHAT="simplex-chat"
```
efim-poberezkin, line 3: Not sure this should be a constant - it's used in different contexts where entities coincidentally have the same name (org, repo, bin name).. Any one gets renamed the constant has to be replaced in some places but not the others. Maybe better to remove it and just hardcode in the `PLATFORM_BIN` and `URL` variables

install.sh
```diff
@@ -0,0 +1,48 @@
...
+######################
+
+######################
+# Detect the shell and the file where to export the path
+FILE_TO_EXPORT=""
+
+if [ -n "$($SHELL -c 'echo $ZSH_VERSION')" ]; then
+	FILE_TO_EXPORT=".zshrc"
+elif [ -n "$($SHELL -c 'echo $BASH_VERSION')" ]; then
+	FILE_TO_EXPORT=".bashrc"
```
efim-poberezkin, line 27: I think `~/.profile` would be a more appropriate file for expanding PATH, see an explanation here [https://unix.stackexchange.com/questions/26047/how-to-correctly-add-a-path-to-path/26059#26059](https://unix.stackexchange.com/questions/26047/how-to-correctly-add-a-path-to-path/26059#26059) "Where to put it". Probably this part doesn't even have to be shell-dependent, but better double check.

install.sh
```diff
@@ -0,0 +1,48 @@
...
+[ ! -d $SIMPLEX_CHAT_DIR ] && mkdir -p $SIMPLEX_CHAT_DIR
+
+# Build the url
+URL="https://github.com/$SIMPLEX_CHAT/$SIMPLEX_CHAT/releases/download/$VERSION/$PLATAFORM_BIN"
+
+# Download the binary and make it executable
+wget -O $SIMPLEX_CHAT_DIR/simplex-chat $URL && chmod +x $SIMPLEX_CHAT_DIR/simplex-chat
+
+# Export the path to the binary
+echo "export PATH=\$PATH:$SIMPLEX_CHAT_DIR" >>$HOME/$FILE_TO_EXPORT
```
efim-poberezkin, line 45: Also no need to `export` it since the PATH would already be in the environment for our use case (see another explanation in the same post from the previous comment), just

```suggestion
# Update PATH
echo "PATH=\$PATH:$TARGET_DIR" >> $HOME/.profile
```

(we probably need to first check that TARGET_DIR is not already there in the PATH, since if we switch it to `~/.local/bin` there's a chance it is)

install.sh
```diff
@@ -0,0 +1,48 @@
+# Constants
+VERSION="v0.4.2"
+SIMPLEX_CHAT="simplex-chat"
+
+######################
+PLATAFORM=""
```
efim-poberezkin, line 6:
```suggestion
PLATFORM=""
```

a small typo here and in some other places

UltiRequiem:
What do I have to do to prevent the CLA Assistant from failing?

install.sh
UltiRequiem: Fixed, thanks!

install.sh
UltiRequiem: I think we can use `~/bin` instead of `~/.local/bin`, which in my case, Archlinux User, didn't even exist until now.

If we use `~/bin` we don't have to export anything to the path because its automatically done, At least in bash, dash and zsh.

https://unix.stackexchange.com/questions/26047/how-to-correctly-add-a-path-to-path/26059#26059
> Some distributions automatically put ~/bin in your PATH if it exists, by the way.

install.sh
UltiRequiem: Now I'm not exporting anything so I don't need to know the shell.

install.sh
UltiRequiem: https://github.com/simplex-chat/simplex-chat/pull/127#discussion_r757816962

install.sh
UltiRequiem: I changed the name of the variables so it makes more sense to make it constants.

epoberezkin:
> What do I have to do to prevent the CLA Assistant from failing?

Looking into it, it just stopped working for some reason :)
If it's not obvious how to fix will should just remove it...

install.sh
epoberezkin: @efim-poberezkin the previous version was better, as zsh (default on Mac) does not use .profile, it uses .zprofile instead, and most tools I know add lines to .zshrc and .bashrc (not sure why, but maybe they are reserving .profile/.zprofile to what is edited by a user manually...)

@UltiRequiem I will revert the commit that changed if it's ok?

install.sh
```diff
@@ -0,0 +1,19 @@
...
+APP_NAME="simplex-chat"
+TARGET_DIR="$HOME/bin"
+PLATFORM="$(uname)"
+
+if [ $PLATFORM == "Darwin" ]; then
+	PLATFORM="macos-x86-64"
+elif [ $PLATFORM == "Linux" ]; then
+	PLATFORM="ubuntu-20_04-x86-64"
+else
+	echo "Your platform is not suported, try with macos/linux."
```
epoberezkin, line 11:
```suggestion
	echo "Scripted installation on your platform is not supported."
	echo "See compiled binaries in the latest release: https://github.com/simplex-chat/simplex-chat/releases/latest
```

install.sh
```diff
@@ -0,0 +1,19 @@
...
+	exit 1
+fi
+
+[ ! -d $TARGET_DIR ] && mkdir -p $TARGET_DIR
+
+wget -O $TARGET_DIR/$APP_NAME "https://github.com/$APP_NAME/$APP_NAME/releases/download/v$VERSION/$PLATFORM" && chmod +x $TARGET_DIR/$APP_NAME
```
epoberezkin, line 17:
```suggestion
wget -O $TARGET_DIR/$APP_NAME "https://github.com/$APP_NAME/$APP_NAME/releases/latest/download/$APP_NAME-$PLATFORM" && chmod +x $TARGET_DIR/$APP_NAME
```

This change switches to the latest version, so we don't have to update this file for the new releases.
It also fixes the binary name that includes simplex-chat

install.sh
```diff
@@ -0,0 +1,19 @@
...
+
+echo "$APP_NAME was installed sucesfully!"
```
epoberezkin, line 19:
```suggestion
echo "$APP_NAME was installed sucesfully!"
echo "Type simplex-chat in your terminal to run it"
```

install.sh
```diff
@@ -0,0 +1,19 @@
+VERSION="0.4.2"
```
epoberezkin, line 1:
```suggestion
```

see the change below - I switched to the latest release

epoberezkin:
@UltiRequiem I see that writing to .profile is removed now...

I understand that the assumption is that ~/bin is on PATH but it's not guaranteed, in fact ~/bin may not even exist (I don't have it for example - although it's created, it's an uncommon location for binaries)...

On Mac I have ~/.local/bin - I suggest we revert to it.
I am not sure what's common on linux, but if it is indeed ~/bin, then we can make it conditional, in the same IF that chooses the binary.

epoberezkin:
recheck

epoberezkin:
> What do I have to do to prevent the CLA Assistant from failing?

@UltiRequiem apparently it works - you now need to type "I have read the CLA Document and I hereby sign the CLA" to make it happy :)

epoberezkin:
also @UltiRequiem I made the change in this branch (the last commit) - https://github.com/simplex-chat/simplex-chat/pull/129 - I cannot push to your repo to update it... Probably better if we could merge it here with the conversation, not sure what's the best way, maybe you can copy/paste?

epoberezkin:
ok, closing it, #129 is merged

@UltiRequiem - thank you very much - we've been talking about install script for ages but never came around to doing it.

Remaining improvements:

- [ ] make it work on Windows in git-bash (I never bothered using WSL with windows  and VMs are slow, I think many people just use git-bash)
- [ ] add to readme

@UltiRequiem - please let me know if you want to pick up either - and thank you!

UltiRequiem:
> * make it work on Windows in git-bash (I never bothered using WSL with windows  and VMs are slow, I think many people just use git-bash)

I would like to take this, but I don't have any Windows device   to try, when I used Windows I always used WSL, and actually of people I know its weird to use git-bash.

> [ ] add to readme

I can do this

PR #152, opened 2021-12-11
simple welcome message

epoberezkin:


src/Simplex/Chat/Store.hs
```diff
@@ -1586,6 +1591,21 @@ getSndFileTransfers_ db userId fileId =
         Just recipientDisplayName -> Right SndFileTransfer {..}
         Nothing -> Left $ SESndFileInvalid fileId
 
+getOnboarding :: MonadUnliftIO m => SQLiteStore -> UserId -> m Onboarding
```
efim-poberezkin, line 1594: would be kind of counterintuitive that it doesn't really remember your onboarding progress - I can create contact, then delete it and it will be marked as incomplete it seems. Not a big deal though

efim-poberezkin (approved):

src/Simplex/Chat/Store.hs
epoberezkin: Yeah I thought for about 10 sec about making a separate table with this state :)

PR #154, opened 2021-12-16
readme

vsevolod-mineev:
Added the network topology image to the Readme.


epoberezkin:
sorry, man, this is definitely no the style we can have in our technical docs - it's not our network topology...

vsevolod-mineev:
Fair enough, it was a long shot. 

I was actually thinking about that. Is it not? It is if you pretent the rave at the bottom left is a server.

PR #217, opened 2022-01-21
groups: delayed delivery of messages and introductions to announced members

efim-poberezkin:


src/Simplex/Chat.hs
```diff
@@ -942,13 +942,14 @@ processAgentMessage toView user@User {userId, profile} agentConnId agentMessage
...
             Nothing -> messageError "x.grp.mem.inv error: referenced member does not exists"
-            Just reMember -> do
-              intro <- withStore $ \st -> saveIntroInvitation st reMember m introInv
-              case activeConn (reMember :: GroupMember) of
-                Nothing -> pure () -- this is not an error, introduction will be forwarded once the member is connected
+            Just reMember@GroupMember {groupMemberId = reMemberId} -> do
+              GroupMemberIntro {introId} <- withStore $ \st -> saveIntroInvitation st reMember m introInv
+              (msgId, msgBody) <- createSndMessage $ XGrpMemFwd (memberInfo m) introInv
+              case memberConn reMember of
+                Nothing -> withStore $ \st -> createPendingGroupMessage st reMemberId msgId (Just introId)
```
epoberezkin, line 949: I don't think it is right... You need only current members here, not all members (that would include those that are invited to the group by the current user but did not accept yet).

src/Simplex/Chat.hs
```diff
@@ -656,9 +656,9 @@ processAgentMessage toView user@User {userId, profile} agentConnId agentMessage
             showToast ("#" <> gName) $ "member " <> localDisplayName (m :: GroupMember) <> " is connected"
             intros <- withStore $ \st -> createIntroductions st group m
             sendGroupMessage members . XGrpMemNew $ memberInfo m
-            forM_ intros $ \intro -> do
+            forM_ intros $ \intro@GroupMemberIntro {introId} -> do
               sendDirectMessage conn . XGrpMemIntro . memberInfo $ reMember intro
-              withStore $ \st -> updateIntroStatus st intro GMIntroSent
+              withStore $ \st -> updateIntroStatus st introId GMIntroSent
```
epoberezkin, line 661: why?

src/Simplex/Chat.hs
```diff
@@ -1143,17 +1148,27 @@ deliverMessage Connection {connId, agentConnId} msgBody msgId = do
...
 sendGroupMessage :: ChatMonad m => [GroupMember] -> ChatMsgEvent -> m ()
 sendGroupMessage members chatMsgEvent = do
-  let msgBody = directMessage chatMsgEvent
-      newMsg = NewMessage {direction = MDSnd, chatMsgEventType = toChatEventTag chatMsgEvent, msgBody}
-  msgId <- withStore $ \st -> createNewMessage st newMsg
-  -- TODO once scheduled delivery is implemented memberActive should be changed to memberCurrent
-  forM_ (map memberConn $ filter memberActive members) $
-    traverse (\conn -> deliverMessage conn msgBody msgId)
+  (msgId, msgBody) <- createSndMessage chatMsgEvent
+  for_ (filter memberCurrent members) $ \m@GroupMember {groupMemberId} ->
```
epoberezkin, line 1152: here it does filter current, but not above. That's why I suggested refactoring - sending to the group would also include filtering current members.

src/Simplex/Chat.hs
```diff
@@ -1143,17 +1148,27 @@ deliverMessage Connection {connId, agentConnId} msgBody msgId = do
...
+    case memberConn m of
+      Nothing -> withStore $ \st -> createPendingGroupMessage st groupMemberId msgId Nothing
+      Just conn -> deliverMessage conn msgBody msgId
+
+sendPendingGroupMessages :: ChatMonad m => GroupMember -> Connection -> m ()
+sendPendingGroupMessages GroupMember {groupMemberId, localDisplayName} conn = do
+  pendingMessages <- withStore $ \st -> getPendingGroupMessages st groupMemberId
+  -- TODO ensure order
+  for_ pendingMessages $ \(msgId, msgBody, chatMsgEventTag, mIntroId) -> do
```
epoberezkin, line 1161: looks like it needs the type. I usually don't go beyond 3 with ad-hoc tuples...

src/Simplex/Chat/Types.hs
```diff
@@ -629,12 +630,97 @@ serializeIntroStatus = \case
 
 data NewMessage = NewMessage
   { direction :: MsgDirection,
-    chatMsgEventType :: Text,
+    chatMsgEventTag :: CMEventTag,
```
epoberezkin, line 633: why not just move message to protocol instead?

src/Simplex/Chat/Types.hs
```diff
@@ -629,12 +630,97 @@ serializeIntroStatus = \case
...
     msgBody :: MsgBody
   }
 
 type MessageId = Int64
 
+data CMEventTag
```
epoberezkin, line 639: this type definitely does not belong here, it should stay in the protocol

PR #310, opened 2022-02-15
asynchronously subscribe to user connections

efim-poberezkin:


src/Simplex/Chat/Mobile.hs
```diff
@@ -49,7 +49,8 @@ mobileChatOpts =
   ChatOpts
     { dbFilePrefix = "simplex_v1", -- two database files will be created: simplex_v1_chat.db and simplex_v1_agent.db
       smpServers = defaultSMPServers,
-      logging = False
+      logConnections = False,
```
efim-poberezkin, line 52: `logSubscriptions` might've been more descriptive, but it's okay to keep as is
```suggestion
      logSubscriptions = False,
```

src/Simplex/Chat/Options.hs
```diff
@@ -55,9 +56,14 @@ chatOpts appDir =
           <> value defaultSMPServers
       )
     <*> switch
-      ( long "log"
+      ( long "connections"
+          <> short 'c'
+          <> help "Log every contact and group connection on start"
```
efim-poberezkin, line 61: same here
```suggestion
      ( long "subscriptions"
          <> short 'c'
          <> help "Log every contact and group subscription on start"
```

src/Simplex/Chat/Options.hs
```diff
@@ -21,7 +21,8 @@ import System.FilePath (combine)
 data ChatOpts = ChatOpts
   { dbFilePrefix :: String,
     smpServers :: NonEmpty SMPServer,
-    logging :: Bool
+    logConnections :: Bool,
```
efim-poberezkin, line 24:
```suggestion
    logSubscriptions :: Bool,
```

src/Simplex/Chat/Options.hs
epoberezkin: subscription is our internal term, the current message in UI is "contact is connected"

PR #335, opened 2022-02-18
chat layout

efim-poberezkin:


apps/android/app/src/main/java/chat/simplex/app/model/ChatModel.kt
```diff
@@ -498,8 +500,14 @@ class CIMeta (
...
-// TODO
-fun getTimestampText(d: Instant): String = ""
+// TODO use old api?
+// TODO date for older timestamps
+@SuppressLint("NewApi")
+fun getTimestampText(d: Instant): String {
+  val dt = d.toLocalDateTime(TimeZone.currentSystemDefault()).toJavaLocalDateTime()
+  val formatter = DateTimeFormatter.ofPattern("HH:mm")
+  return dt.format(formatter)
+}
```
epoberezkin, line 510: we need to unify it with the time/date logic in the chat previews that @IanRDavies did - I will do that later to match the logic in iOS app.

In iOS app it shows the time if it is today OR if it is (yesterday AND the time is after 6pm AND the current time is before 12pm) - so, effectively, in the morning you know if it was the evening.

We can revise it to something else.

apps/android/app/src/main/java/chat/simplex/app/views/chat/ChatView.kt
```diff
@@ -1,55 +1,92 @@
...
-              type = cInfo.chatType,
-              id = cInfo.apiId,
-              mc = MsgContent.MCText(msg)
-            )
-            // hide "in progress"
-            if (newItem != null) chatModel.addChatItem(cInfo, newItem.chatItem)
-          }
+    ChatLayout(chat, chatModel.chatItems, back = { nav.popBackStack() }, sendMessage = { msg ->
+      GlobalScope.launch {
+        withContext(Dispatchers.Main) {
```
epoberezkin, line 33:
```suggestion
      withApi {
```

epoberezkin: + import from Util and matching braces below

apps/android/app/src/main/java/chat/simplex/app/views/chat/item/TextItemView.kt
```diff
@@ -1,10 +1,76 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import chat.simplex.app.model.*
+import chat.simplex.app.ui.theme.SimpleXTheme
+import kotlinx.datetime.*
+
+val SentColorLight = Color(0x1E45B8FF)
+val ReceivedColorLight = Color(0x1EF1F0F5)
```
epoberezkin, line 17: TODO move to theme, but after we ship v0.1

PR #341, opened 2022-02-21
android: UserProfileView

efim-poberezkin:


apps/android/app/src/main/java/chat/simplex/app/model/ChatModel.kt
```diff
@@ -23,6 +19,13 @@ class ChatModel(val controller: ChatController, val alertManager: SimplexApp.Ale
   // set when app is opened via contact or invitation URI
   var appOpenUrl = mutableStateOf<Uri?>(null)
 
+  fun updateUserProfile(profile: Profile) {
+    val user = currentUser.value
+    if (user != null) {
+      currentUser = mutableStateOf(user.copy(profile = profile))
```
epoberezkin, line 25:
```suggestion
      currentUser.value = user.copy(profile = profile)
```

I think it should work like this, but not sure

apps/android/app/src/main/java/chat/simplex/app/views/usersettings/SettingsView.kt
```diff
@@ -5,54 +5,64 @@ import androidx.compose.material.*
...
 import androidx.compose.ui.unit.dp
-import chat.simplex.app.model.Profile
+import androidx.navigation.NavController
+import chat.simplex.app.Pages
+import chat.simplex.app.model.ChatModel
 
 //@Preview(showBackground = true)
 @Composable
-fun SettingsView(profile: Profile) {
+fun SettingsView(chatModel: ChatModel, nav: NavController) {
```
epoberezkin, line 14: split logic from layout

epoberezkin: it would also remove most of the diff ;)

apps/android/app/src/main/java/chat/simplex/app/views/usersettings/UserProfileView.kt
```diff
@@ -0,0 +1,58 @@
...
+import androidx.compose.foundation.layout.Row
+import androidx.compose.material.*
+import androidx.compose.runtime.*
+import androidx.navigation.NavController
+import chat.simplex.app.model.ChatModel
+import chat.simplex.app.model.Profile
+import chat.simplex.app.views.helpers.withApi
+
+@Composable
+fun UserProfileView(chatModel: ChatModel, nav: NavController) {
```
epoberezkin, line 13: also split logic from layout

PR #343, opened 2022-02-21
terminal: make input responsible for echo to keep commands synchronous (as in mobile) and avoid echo delays

epoberezkin:


src/Simplex/Chat/Terminal/Input.hs
```diff
@@ -28,9 +32,23 @@ getKey =
 runInputLoop :: ChatTerminal -> ChatController -> IO ()
 runInputLoop ct cc = forever $ do
   s <- atomically . readTBQueue $ inputQ cc
-  r <- runReaderT (execChatCommand . encodeUtf8 $ T.pack s) cc
+  let bs = encodeUtf8 $ T.pack s
+      cmd = parseAll chatCommandP $ B.dropWhileEnd isSpace bs
+  unless (isMessage cmd) $ printToTerminal ct [plain s]
```
efim-poberezkin, line 37:
```suggestion
  unless (isMessage cmd) $ echo s
```

efim-poberezkin (approved):

PR #344, opened 2022-02-21
android: improve chat, chat info, console

epoberezkin:


apps/android/app/src/main/java/chat/simplex/app/views/TerminalView.kt
```diff
@@ -1,49 +1,82 @@
...
+  val listState = rememberLazyListState()
+  val scope = rememberCoroutineScope()
+  val df = DateTimeFormatter.ofPattern("HH:mm:ss")
+  LazyColumn(state = listState) {
     items(terminalItems) { item ->
-      ClickableText(
-        AnnotatedString(item.label),
-        onClick = { navController.navigate("details/${item.id}") }
-      )
+        Text("${item.date.toString().subSequence(11, 19)} ${item.label}",
```
efim-poberezkin, line 70: DateTimeFormatter seems to be unused and `item.date.toString().subSequence(11, 19)` is somewhat hacky, wat the intention to use the formatter for the text?

epoberezkin: removed df, tacky will do in console

PR #357, opened 2022-02-23
update README by adding "Featured on"

vsevolod-mineev:
- update README by adding "Featured on"


epoberezkin:
sorry, I don't think we should promote all external references...

PR #364, opened 2022-02-23
ios: display name validation

IanRDavies:
My first swift, be kind on review!

### DIFF
--- a/apps/ios/Shared/Views/WelcomeView.swift
+++ b/apps/ios/Shared/Views/WelcomeView.swift
@@ -7,10 +7,15 @@
 
 import SwiftUI
 
+func isValidDisplayName(_ name: String) -> Bool {
+    return !(self.rangeOfCharacter(from: .whitespacesAndNewlines) != nil)
+}
+
 struct WelcomeView: View {
     @EnvironmentObject var chatModel: ChatModel
     @State var displayName: String = ""
     @State var fullName: String = ""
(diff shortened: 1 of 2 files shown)


apps/ios/Shared/Views/WelcomeView.swift
```diff
@@ -7,10 +7,15 @@
...
 
+func isValidDisplayName(_ name: String) -> Bool {
+    return !(self.rangeOfCharacter(from: .whitespacesAndNewlines) != nil)
+}
+
 struct WelcomeView: View {
     @EnvironmentObject var chatModel: ChatModel
     @State var displayName: String = ""
     @State var fullName: String = ""
+    @State var errorText: String = ""
```
epoberezkin, line 18:
```suggestion
```

apps/ios/Shared/Views/WelcomeView.swift
```diff
@@ -7,10 +7,15 @@
 
 import SwiftUI
 
+func isValidDisplayName(_ name: String) -> Bool {
+    return !(self.rangeOfCharacter(from: .whitespacesAndNewlines) != nil)
+}
```
epoberezkin, line 12:
```suggestion
func invalidDisplayName(_ name: String) -> Bool {
    return self.rangeOfCharacter(from: .whitespacesAndNewlines) != nil
}
```

I also think self is not needed here?

PR #368, opened 2022-02-24
core: add fks to messages

efim-poberezkin:


src/Simplex/Chat/Migrations/M20220224_messages_fks.hs
```diff
@@ -0,0 +1,22 @@
...
+ALTER TABLE messages ADD COLUMN group_id INTEGER DEFAULT NULL REFERENCES groups ON DELETE CASCADE;
+
+UPDATE messages
+SET connection_id = (
+  SELECT connection_id
+  FROM msg_deliveries
+  WHERE message_id = messages.message_id
+);
+
+DELETE FROM messages WHERE connection_id IS NULL;
```
epoberezkin, line 21: I don't think this is safe to delete it like that

src/Simplex/Chat/Migrations/M20220224_messages_fks.hs
```diff
@@ -0,0 +1,22 @@
...
+module Simplex.Chat.Migrations.M20220224_messages_fks where
+
+import Database.SQLite.Simple (Query)
+import Database.SQLite.Simple.QQ (sql)
+
+m20220224_messages_fks :: Query
+m20220224_messages_fks =
+  [sql|
+ALTER TABLE messages ADD COLUMN connection_id INTEGER DEFAULT NULL REFERENCES connections ON DELETE CASCADE;
+ALTER TABLE messages ADD COLUMN group_id INTEGER DEFAULT NULL REFERENCES groups ON DELETE CASCADE;
```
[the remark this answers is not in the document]
efim-poberezkin: We could. We also need these for cascade delete on user deletion though. It seemed safer/easier to rely on db for this rather than think about different cases..

src/Simplex/Chat/Migrations/M20220224_messages_fks.hs
efim-poberezkin: Why? It's being populated in the previous query. Worst that can happen is some "unsent" (i.e. not taken by agent for delivery, so not really critical) messages can get deleted, but even that is quite low probability

PR #392, opened 2022-03-02
ios, android: configurable smp servers (only model and api for android)

efim-poberezkin:


apps/ios/Shared/Views/UserSettings/SMPServers.swift
```diff
@@ -0,0 +1,180 @@
...
+                            isUserSMPServers = false
+                            userSMPServersStr = ""
+                        }
+                    }
+                }
+                .padding(.bottom)
+                .alert(isPresented: $showResetServersAlert) {
+                    Alert(
+                        title: Text("Use SimpleX Chat servers?"),
+                        message: Text("Saved SMP servers will be forgotten"),
```
epoberezkin, line 45:
```suggestion
                        message: Text("Saved SMP servers will be removed"),
```

PR #394, opened 2022-03-02
message replies and chat item references

epoberezkin:


src/Simplex/Chat.hs
```diff
@@ -264,6 +265,19 @@ processChatCommand = \case
     contactId <- withStore $ \st -> getContactIdByName st userId cName
     let mc = MCText $ safeDecodeUtf8 msg
     processChatCommand $ APISendMessage CTDirect contactId mc
+  SendReply cName replyToMsg msg -> withUser $ \user@User {userId} -> do
```
efim-poberezkin, line 268: For the purposes of mobile apps we could also add `APISendReply` ChatCommand that would be using IDs directly, similar to `APISendMessage`, and similarly route `SendReply` and `SendGroupReply` through that. Or is there a reason not to do so in this case?

epoberezkin: I was thinking that it would send JSON that includes the quote, as in this test: https://github.com/simplex-chat/simplex-chat/pull/394/files#diff-725b543c03566e1db2facb640d232fd22123b2b13bd669a3fb790da1aece838aR133

epoberezkin: but possibly it is better to have a separate API for sending replies, as in this case the format that exists only in the protocol (MsgContainer) would not be used in the API

PR #451, opened 2022-03-19
message update and delete

epoberezkin:


src/Simplex/Chat.hs
```diff
@@ -1397,7 +1445,7 @@ mkChatItem :: MsgDirectionI d => ChatDirection c d -> ChatItemId -> CIContent d
 mkChatItem cd ciId content quotedItem sharedMsgId itemTs createdAt = do
   tz <- getCurrentTimeZone
   let itemText = ciContentToText content
-      meta = mkCIMeta ciId itemText ciStatusNew sharedMsgId tz itemTs createdAt
+      meta = mkCIMeta ciId itemText ciStatusNew sharedMsgId False False tz itemTs createdAt
```
epoberezkin, line 1448: I don't think False's need to be passed, there is no case you need it set to True

src/Simplex/Chat.hs
```diff
@@ -1000,6 +1034,13 @@ processAgentMessage (Just user@User {userId, profile}) agentConnId agentMessage
       showMsgToast (c <> "> ") content formattedText
       setActive $ ActiveC c
 
+    messageUpdate :: Contact -> SharedMsgId -> MsgContent -> RcvMessage -> MsgMeta -> m ()
+    messageUpdate ct@Contact {contactId, localDisplayName = c} sharedMsgId mc RcvMessage {msgId} msgMeta = do
+      updCi <- withStore $ \st -> updateDirectChatItemContentBySharedMsgId st userId contactId sharedMsgId (CIRcvMsgContent mc) msgId
```
epoberezkin, line 1039: updateDirectChatItem would do

src/Simplex/Chat/Migrations/M20220321_chat_item_edited.hs
```diff
@@ -0,0 +1,12 @@
...
+
+module Simplex.Chat.Migrations.M20220321_chat_item_edited where
+
+import Database.SQLite.Simple (Query)
+import Database.SQLite.Simple.QQ (sql)
+
+m20220321_chat_item_edited :: Query
+m20220321_chat_item_edited =
+  [sql|
+ALTER TABLE chat_items ADD COLUMN item_edited INTEGER; -- 1 for edited
```
epoberezkin, line 11: + item_deleted

src/Simplex/Chat/Store.hs
```diff
@@ -190,7 +195,8 @@ schemaMigrations =
     ("20220224_messages_fks", m20220224_messages_fks),
     ("20220301_smp_servers", m20220301_smp_servers),
     ("20220302_profile_images", m20220302_profile_images),
-    ("20220304_msg_quotes", m20220304_msg_quotes)
+    ("20220304_msg_quotes", m20220304_msg_quotes),
+    ("20220321_chat_item_edited", m20220321_chat_item_edited)
```
epoberezkin, line 199: 20220321_chat_item_updated (or 20220321_chat_item_changes to include deletions)

src/Simplex/Chat.hs
epoberezkin: and updateDirectChatItemViaMsgId

src/Simplex/Chat.hs
epoberezkin: there are store methods

epoberezkin:
It needs the same change as the last commit in #454

PR #495, opened 2022-04-02
ios: update image picker

IanRDavies:
Update the image picker to the newer and more private [`PHPicker`](https://developer.apple.com/videos/play/wwdc2020/10652/)

This then separates camera image selection and image selection from gallery.

Note: This also improves privacy as the `PHPicker` runs in a separate process and only shares selected images with the SimpleX app rather than the full library of images.

### DIFF
--- a/apps/ios/Shared/Views/Helpers/ImagePicker.swift
+++ b/apps/ios/Shared/Views/Helpers/ImagePicker.swift
@@ -31,12 +31,13 @@ struct FromGalleryImagePicker: UIViewControllerRepresentable {
             guard !results.isEmpty else {
                 return
             }
-            // TODO dont force unwrap
-            let chosenImageProvider = results.first!.itemProvider
-            if chosenImageProvider.canLoadObject(ofClass: UIImage.self) {
-                chosenImageProvider.loadObject(ofClass: UIImage.self)  { [weak self] image, error in
-                    DispatchQueue.main.async {
-                        self?.loadImage(object: image, error: error)
+
+            if let chosenImageProvider = results.first?.itemProvider {
+                if chosenImageProvider.canLoadObject(ofClass: UIImage.self) {
+                    chosenImageProvider.loadObject(ofClass: UIImage.self)  { [weak self] image, error in
+                        DispatchQueue.main.async {
+                            self?.loadImage(object: image, error: error)
+                        }
                     }
                 }
             }
--- a/apps/ios/Shared/Views/Helpers/ImagePicker.swift
+++ b/apps/ios/Shared/Views/Helpers/ImagePicker.swift
@@ -10,7 +10,7 @@ import SwiftUI
 import PhotosUI
 
 enum ImageSelectionMethod {
-    case library
+    case imageLibrary
     case camera
 }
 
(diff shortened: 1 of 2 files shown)


apps/ios/Shared/Views/Helpers/ImagePicker.swift
```diff
@@ -7,44 +7,102 @@
...
 import SwiftUI
+import PhotosUI
 
-struct ImagePicker: UIViewControllerRepresentable {
+enum ImageSelectionMethod: Equatable {
+    case gallery
+    case camera
+}
+
+struct FromGalleryImagePicker: UIViewControllerRepresentable {
```
epoberezkin, line 17:
```suggestion
struct GalleryImagePicker: UIViewControllerRepresentable {
```

and below...

apps/ios/Shared/Views/Helpers/ImagePicker.swift
```diff
@@ -7,44 +7,102 @@
 //
 
 import SwiftUI
+import PhotosUI
 
-struct ImagePicker: UIViewControllerRepresentable {
+enum ImageSelectionMethod: Equatable {
+    case gallery
```
epoberezkin, line 13: they call it photoLibrary - should we?

apps/ios/Shared/Views/Helpers/ImagePicker.swift
```diff
@@ -7,44 +7,102 @@
...
-    var source: UIImagePickerController.SourceType
     @Binding var image: UIImage?
-    @Binding var imageUrl: URL?
-    
+
     class Coordinator: NSObject, UINavigationControllerDelegate, UIImagePickerControllerDelegate {
-        let parent: ImagePicker
-        
-        init(_ parent: ImagePicker) {
+        let parent: FromCameraImagePicker
```
epoberezkin, line 78:
```suggestion
        let parent: CameraImagePicker
```

apps/ios/Shared/Views/Helpers/ImagePicker.swift
```diff
@@ -7,44 +7,102 @@
 //
 
 import SwiftUI
+import PhotosUI
 
-struct ImagePicker: UIViewControllerRepresentable {
+enum ImageSelectionMethod: Equatable {
```
epoberezkin, line 12: ImageSource ?

apps/ios/Shared/Views/UserSettings/UserProfile.swift
```diff
@@ -84,14 +83,20 @@ struct UserProfile: View {
...
             }
             Button("Choose from library") {
-                imageSource = .photoLibrary
+                imageSource = .gallery
                 showImagePicker = true
             }
         }
         .sheet(isPresented: $showImagePicker) {
-            ImagePicker(source: imageSource, image: $pickedImage, imageUrl: $tmpImageUrl)
+            if imageSource == .gallery {
```
epoberezkin, line 91: better use switch for enum (you won't need Equatable then, it it checks for completeness)

apps/ios/Shared/Views/Helpers/ImagePicker.swift
IanRDavies: I'm tempted to leave it as library or gallery to differentiate it from the iOS versions since it is a different (self-rolled) API

PR #503, opened 2022-04-04
ios: link previews

IanRDavies:
Adding image previews to iOS app.

Note that we load the preview on the sender's phone and send the preview details in message to avoid unprovoked network calls on receipt.

There are some errors in the logs. Searching seems to suggest they can [be](https://stackoverflow.com/questions/65162287/wkwebview-is-trying-to-run-on-background-although-its-closed-and-should-be-dis) [ignored](https://developer.apple.com/forums/thread/112095)

There are some [potential fixes](https://stackoverflow.com/questions/44585980/com-apple-webkit-webcontent-drops-113-error-could-not-find-specified-service) which so far haven't worked.


apps/ios/Shared/Views/Chat/ComposeMessage/SendMessageView.swift
```diff
@@ -92,6 +92,7 @@ struct SendMessageView_Previews: PreviewProvider {
         @State var editingOn: Bool = true
         @State var item: ChatItem? = ChatItem.getSample(1, .directSnd, .now, "hello")
         @State var nilItem: ChatItem? = nil
+        @State var hyperlink: String? = nil
```
IanRDavies, line 95: I think this could probably be removed

apps/ios/Shared/Views/Chat/ComposeMessage/ComposeView.swift
```diff
@@ -7,6 +7,7 @@
 //
 
 import SwiftUI
+import LinkPresentation
```
IanRDavies, line 10: I believe this import is now not needed

apps/ios/Shared/Views/Helpers/ComposeLinkView.swift
```diff
@@ -0,0 +1,91 @@
...
+//  Created by Ian Davies on 04/04/2022.
+//  Copyright © 2022 SimpleX Chat. All rights reserved.
+//
+
+import SwiftUI
+import LinkPresentation
+
+
+func getLinkPrivew(url: URL, cb: @escaping (LinkPreview?) -> Void) {
+    logger.debug("getLinkMetadata: fetching URL preview")
```
IanRDavies, line 14: Do we still need this line?

apps/ios/Shared/Views/Helpers/ImagePicker.swift
```diff
@@ -9,6 +9,64 @@
...
+        let vScale = newSize.width / size.width
+        let scale = max(hScale, vScale) // scaleToFill
+        let resizeSize = CGSize(width: size.width * scale, height: size.height * scale)
+        var middle = CGPoint.zero
+        if resizeSize.width > newSize.width {
+            middle.x -= (resizeSize.width - newSize.width) / 2
+        } else if resizeSize.height > newSize.height {
+            middle.y -= (resizeSize.height - newSize.height) / 2
+        }
+        image.draw(in: CGRect(origin: middle, size: resizeSize))
```
IanRDavies, line 36: nit: I find origin being the middle confusing. Does the image span -width/2, width/2?
I would have origin as bottom left corner but I may be silly

apps/ios/Shared/Views/Helpers/ImagePicker.swift
```diff
@@ -9,6 +9,64 @@
...
+    return resizeAndCrop(image, to: CGSize(width: side, height: side))
+}
+
+func resizeImageToDataSize(_ image: UIImage, maxSize: Int) -> String? {
+    let size = image.size
+    var imageStr = compressImage(image)
+    var resized = image
+    var ratio: CGFloat = 1
+    var dataSize = imageStr?.count ?? 0
+    logger.debug("resizeImageToDataSize: initial size \(String(describing: size)), data size \(dataSize)")
```
IanRDavies, line 51: Are these debug calls hangovers from debugging? keep them?

apps/ios/Shared/Views/Helpers/ComposeLinkView.swift
epoberezkin: I don't think there is any harm in keeping debug logs

apps/ios/Shared/Views/Helpers/ImagePicker.swift
epoberezkin: it works, I'd rather keep as is :)

PR #510, opened 2022-04-08
android: link previews

IanRDavies:


apps/ios/Shared/Views/Chat/ComposeMessage/ComposeView.swift
```diff
@@ -106,9 +66,52 @@ struct ComposeView: View {
...
+            })
+            if let uri = uri { return URL(string: uri.text) }
+            else { return nil }
+        } catch {
+            logger.error("apiParseMarkdown error: \(error.localizedDescription)")
+            return nil
+        }
+    }
+
+    private func isSimplexLink(_ link: String) -> Bool {
```
IanRDavies, line 102: This is a redefinition of the same function in SendMessageView.kt

epoberezkin: probably forgot to remove it in one place - I just moved them all together

epoberezkin: ah - this is in swift

PR #530, opened 2022-04-14
core: support files folder for mobile, delete files, chat item in CRRcvFileComplete

jr-simplex:


src/Simplex/Chat.hs
```diff
@@ -1815,6 +1869,7 @@ chatCommandP =
   ("/user " <|> "/u ") *> (CreateActiveUser <$> userProfile)
     <|> ("/user" <|> "/u") $> ShowActiveUser
     <|> "/_start" $> StartChat
+    <|> "/_set_files_folder " *> (SetFilesFolder <$> filePath)
```
epoberezkin, line 1872:
```suggestion
    <|> "/_files_folder " *> (SetFilesFolder <$> filePath)
```

src/Simplex/Chat.hs
```diff
@@ -64,7 +64,7 @@ import System.IO (Handle, IOMode (..), SeekMode (..), hFlush, openFile, stdout)
 import Text.Read (readMaybe)
 import UnliftIO.Async
 import UnliftIO.Concurrent (forkIO, threadDelay)
-import UnliftIO.Directory (doesDirectoryExist, doesFileExist, getFileSize, getHomeDirectory, getTemporaryDirectory)
+import UnliftIO.Directory (createDirectoryIfMissing, doesDirectoryExist, doesFileExist, getFileSize, getHomeDirectory, getTemporaryDirectory, removeFile, removePathForcibly)
```
epoberezkin, line 67:
```suggestion
import UnliftIO.Directory
```
or as qualified if you want...

src/Simplex/Chat.hs
```diff
@@ -172,6 +173,11 @@ processChatCommand = \case
     asks agentAsync >>= readTVarIO >>= \case
       Just _ -> pure CRChatRunning
       _ -> startChatController user $> CRChatStarted
+  SetFilesFolder filesFolder' -> withUser' $ \_ -> do
```
epoberezkin, line 176:
```suggestion
  SetFilesFolder filesFolder' -> withUser' $ \_ -> do
```

shouldn't it be `withUser`? Folder should be set once it's started, right?

src/Simplex/Chat.hs
```diff
@@ -660,6 +675,33 @@ processChatCommand = \case
...
 
+-- mobile clients use file paths relative to app directory (e.g. for the reason ios app directory changes on updates),
+-- so we have to differentiate between the file path stored in db and communicated with frontend, and the file path
+-- used during file transfer for actual operations with file system
+toFSFilePath :: ChatMonad m => FilePath -> m FilePath
+toFSFilePath f = do
+  ff <- asks filesFolder
+  readTVarIO ff >>= \case
+    Nothing -> pure f
+    Just filesFolder -> pure $ filesFolder <> "/" <> f
```
epoberezkin, line 686:
```
toFSFilePath f =
  maybe f (<> "/" <> f) <$> (readTVarIO =<< asks filesFolder)
```

src/Simplex/Chat.hs
```diff
@@ -660,6 +675,33 @@ processChatCommand = \case
...
+    Just filesFolder ->
+      forM_ files $ \(fileId, filePath_, status) -> do
+        case status of
+          AFS _ CIFSRcvTransfer -> closeFileHandle fileId rcvFiles
+          _ -> pure ()
+        case filePath_ of
+          Just filePath -> do
+            let fsFilePath = filesFolder <> "/" <> filePath
+            removeFile fsFilePath `E.catch` \(_ :: E.SomeException) ->
+              removePathForcibly fsFilePath `E.catch` \(_ :: E.SomeException) -> pure ()
```
epoberezkin, line 702: why is it needed?

src/Simplex/Chat.hs
```diff
@@ -326,8 +338,10 @@ processChatCommand = \case
       ct@Contact {localDisplayName} <- withStore $ \st -> getContact st userId chatId
       withStore (\st -> getContactGroupNames st userId ct) >>= \case
         [] -> do
+          files <- withStore $ \st -> getContactFiles st userId ct
```
epoberezkin, line 341: this is wrong. You are loading files unconditionally, and then, further down the line, you check whether the files should be deleted (if folder is set). Instead you should check the folder being set here and only load files/call deletion if it is set.

src/Simplex/Chat/Mobile.hs
```diff
@@ -48,7 +49,7 @@ cChatRecvMsg cc = deRefStablePtr cc >>= chatRecvMsg >>= newCAString
 mobileChatOpts :: ChatOpts
 mobileChatOpts =
   ChatOpts
-    { dbFilePrefix = "simplex_v1", -- two database files will be created: simplex_v1_chat.db and simplex_v1_agent.db
+    { dbFilePrefix = undefined,
```
epoberezkin, line 52: why is that? I am fighting `undefined`s in config, it's a bad hack, there should be a better way

src/Simplex/Chat.hs
jr-simplex: correct

src/Simplex/Chat.hs
jr-simplex: just in case file couldn't be deleted normally
https://hackage.haskell.org/package/directory-1.3.7.0/docs/System-Directory.html#v:removePathForcibly
e.g. if for some reason the file permissions were changed to read only it would still attempt to remove it.

PR #549, opened 2022-04-22
ios: show local notification when checking messages having received background notification

epoberezkin:


apps/ios/Shared/Model/NtfManager.swift
```diff
@@ -108,6 +110,13 @@ class NtfManager: NSObject, UNUserNotificationCenterDelegate, ObservableObject {
                 actions: [],
                 intentIdentifiers: [],
                 hiddenPreviewsBodyPlaceholder: NSLocalizedString("New message", comment: "notifications")
+            ),
+            // TODO remove
+            UNNotificationCategory(
+                identifier: ntfCategoryMessageReceived,
```
jr-simplex, line 116: should it be a different category?
```suggestion
                identifier: ntfCategoryCheckingMessages,
```

apps/ios/Shared/Model/NtfManager.swift
```diff
@@ -167,7 +176,16 @@ class NtfManager: NSObject, UNUserNotificationCenterDelegate, ObservableObject {
 //            userInfo: ["chatId": cInfo.id, "chatItemId": cItem.id]
         )
     }
-    
+
+    // TODO remove
+    func notifyCheckingMessages() {
+        logger.debug("NtfManager.notifyCheckingMessages")
+        addNotification(
+            categoryIdentifier: ntfCategoryMessageReceived,
```
jr-simplex, line 184: and here
```suggestion
            categoryIdentifier: ntfCategoryCheckingMessages,
```

PR #553, opened 2022-04-22
core: api for pending contact connections

epoberezkin:


src/Simplex/Chat/Types.hs
```diff
@@ -683,6 +665,17 @@ instance ToJSON Connection where
   toJSON = J.genericToJSON J.defaultOptions {J.omitNothingFields = True}
   toEncoding = J.genericToEncoding J.defaultOptions {J.omitNothingFields = True}
 
+data PendingContactConnection = PendingContactConnection
```
jr-simplex, line 668: after pending connection is sent to front in CRNewContactConnection and then contact is created, how will the contact replace the connection? should PendingContactConnection have some entityId field like Connection did?

epoberezkin: there is no contact ID at this point, it's the other way around - contact has connection ID

jr-simplex (approved):

PR #554, opened 2022-04-23
Support Markdown Composition 

Reep236:
Important notes to start:

1. This has only been tested for the terminal. I don't see any reason at the style-level for this to behave strangely on other devices, but it's worth an overview from someone more familiar with the project as of now. 

2. This may have an inconsistent style with the rest of the project. Right off the bat, the explicit exports mean it could be desirable to drop the `where` clause if it is deemed unnecessary. Personally, I just find that some of the function names lose meaning outside of the context. Additionally, the use of lambda cases can be interchanged with equations if desired. 

3. It's very evident that this is a tacked-on solution. It compensates greatly for the existing structure, and while this is great for compatibility and minimizing conflicts, it leaves the code feeling out-of-place. While I wouldn't consider it unintuitive in its present form, it appears to fight rather than leverage the rest of the system in place. 

With that out of the way, it seems to work quite well at the moment. All of the elements laid out in `/markdown` seem to work as described and meet basic sanity checks. With cases like secret + color, it may be worth clarifying exactly how that should be have. Currently both operate about as the code would have you expect, black highlight and colored text (allowing it to be seen).

### DIFF
--- a/src/Simplex/Chat/Styled.hs
+++ b/src/Simplex/Chat/Styled.hs
@@ -40,14 +40,14 @@ styleMarkdown = \case
         merge :: StyledString -> StyledString 
         merge = \case 
             (Styled f1 s :<>: Styled f2 _) -> Styled (f1 ++ f2) s 
-            (Styled f s :<>: a) -> (<> Styled f s) . merge $ a 
+            (Styled f s :<>: a) -> merge . (<>) (Styled f s) . merge $ a 
             sing -> sing
 
 styleMarkdownList :: MarkdownList -> StyledString
 styleMarkdownList = \case 
     [] -> plain ""
     [FormattedText f s] -> styleMarkdown (Markdown f s)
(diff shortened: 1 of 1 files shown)


Reep236:
Also, it would appear that a commit got pulled in that shouldn't have, likely due to initially branching from the stable default. Shouldn't be of too great concern, but also not relevant.

epoberezkin:
hey - good start!

On code style - we use ormolu - you need to install Haskell extension in vscode and enable auto-formatting.

To merge it, it indeed needs to be updated in mobile apps. Is it something you want to try?

VictoriousFan:
In the SimpleX Android app Markdown is currently implemented, is this PR left open by accident?

epoberezkin:
@VictoriousFan this pr adds support for parsing nested markdown, but it’s not supported in mobile UIs

PR #565, opened 2022-04-24
ios: pending contact connections UI, core: delete connections on the server when deleting in UI/db

epoberezkin:


jr-simplex (dismissed):

apps/ios/Shared/Model/ChatModel.swift
```diff
@@ -476,20 +494,27 @@ struct ContactSubStatus: Decodable {
...
     var contactRequestId: Int64
     var localDisplayName: ContactName
     var profile: Profile
     var createdAt: Date
+    var updatedAt: Date
 
     var id: ChatId { get { "<@\(contactRequestId)" } }
     var apiId: Int64 { get { contactRequestId } }
-    var ready: Bool { get { true } }
+    var ready: Bool = true
```
epoberezkin, line 517:
```suggestion
    var ready: Bool { get { true } }
```

PR #590, opened 2022-05-01
core: webrtc calls api types

epoberezkin:


src/Simplex/Chat/Controller.hs
```diff
@@ -240,6 +248,12 @@ data ChatResponse
   | CRPendingSubSummary {pendingSubStatus :: [PendingSubStatus]}
   | CRSndFileSubError {sndFileTransfer :: SndFileTransfer, chatError :: ChatError}
   | CRRcvFileSubError {rcvFileTransfer :: RcvFileTransfer, chatError :: ChatError}
+  | CRCallInvitationSent {contact :: Contact}
```
jr-simplex, line 251: Is there a reason to use the specific response for invitation and not CRCmdOk like for other commands?

src/Simplex/Chat/Messages.hs
```diff
@@ -524,13 +537,42 @@ dbJsonCIContent = \case
...
   DBJCISndMsgContent mc -> ACIContent SMDSnd $ CISndMsgContent mc
   DBJCIRcvMsgContent mc -> ACIContent SMDRcv $ CIRcvMsgContent mc
   DBJCISndDeleted cidm -> ACIContent SMDSnd $ CISndDeleted cidm
   DBJCIRcvDeleted cidm -> ACIContent SMDRcv $ CIRcvDeleted cidm
+  DBJCISndCall {status, duration} -> ACIContent SMDSnd $ CISndCall status duration
+  DBJCIRcvCall {status, duration} -> ACIContent SMDRcv $ CIRcvCall status duration
+
+data CICallStatus
+  = CISCallPending
+  | CISCallMissed
```
jr-simplex, line 554: How is Missed status achieved in state machine? Is it when the call wasn't successfully negotiated? Would it be the same when the call couldn't be negotiated due to an error?

src/Simplex/Chat/Controller.hs
epoberezkin: callId was used internally in the beginning, can be replaced with Ok too

src/Simplex/Chat/Messages.hs
epoberezkin: When there was no response and it ended. Probably there should be one more status for the error

jr-simplex (approved):

PR #606, opened 2022-05-05
mobile: webrtc calls state machine

epoberezkin:


apps/ios/Shared/Views/ChatList/ChatListView.swift
```diff
@@ -53,6 +54,29 @@ struct ChatListView: View {
                     NewChatButton()
                 }
             }
+            .fullScreenCover(isPresented: $showCallView) {
```
jr-simplex, line 57: Just wondering what was the reason to place this fullScreenCover here and not in ChatView?

jr-simplex (approved):

apps/ios/Shared/Views/ChatList/ChatListView.swift
epoberezkin: You don’t have to be in chat to accept the call. Also, call is a singleton - there can’t be 2. It still works when you are in the chat.

PR #607, opened 2022-05-06
core: use filename in quote when there is no message text (MCText)

jr-simplex:


src/Simplex/Chat.hs
```diff
@@ -276,13 +276,25 @@ processChatCommand = \case
...
+          _ -> qmc
+        where
+          -- if the message we're quoting with is one of the "large" MsgContents
+          -- we replace the quote's content with MCText
+          replaceContent = case mc of
+            MCText _ -> False
+            MCLink {} -> True
+            MCImage {} -> True
+            MCFile _ -> False
+            MCUnknown {} -> True
```
epoberezkin, line 294:
```suggestion
            MCText _ -> False
            MCFile _ -> False
            MCLink {} -> True
            MCImage {} -> True
            MCUnknown {} -> True
```

PR #636, opened 2022-05-11
Managing battery optimization on Android 12+ via in-app UI

avently:
- in case of battery optimization is enabled a user will be asked to disable it if he wants to have a background service                                                                                
- when the service is enabled but the user don't want to disable the battery optimization, the service will be disabled with an alert for the user.

The behavior is the same as you shown on the mind map in our talk except the left bottom corner of the mind map.

### DIFF
--- a/apps/android/app/src/main/java/chat/simplex/app/SimplexApp.kt
+++ b/apps/android/app/src/main/java/chat/simplex/app/SimplexApp.kt
@@ -52,7 +52,6 @@ class SimplexApp: Application(), LifecycleEventObserver {
       } else {
         chatController.startChat(user)
         SimplexService.start(applicationContext)
-        chatController.showBackgroundServiceNotice()
       }
     }
   }
@@ -65,6 +64,9 @@ class SimplexApp: Application(), LifecycleEventObserver {
           if (!chatController.getRunServiceInBackground()) SimplexService.stop(applicationContext)
         Lifecycle.Event.ON_START ->
           SimplexService.start(applicationContext)
+        Lifecycle.Event.ON_RESUME ->
+          if (chatModel.onboardingStage.value == OnboardingStage.OnboardingComplete)
+            chatController.showBackgroundServiceNoticeIfNeeded()
       }
     }
   }
--- a/apps/android/app/src/main/AndroidManifest.xml
+++ b/apps/android/app/src/main/AndroidManifest.xml
@@ -16,6 +16,7 @@
   <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
   <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
   <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+  <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" />
 
   <application
       android:name="SimplexApp"
--- a/apps/android/app/src/main/java/chat/simplex/app/model/SimpleXAPI.kt
+++ b/apps/android/app/src/main/java/chat/simplex/app/model/SimpleXAPI.kt
@@ -1,9 +1,14 @@
 package chat.simplex.app.model
 
+import android.annotation.SuppressLint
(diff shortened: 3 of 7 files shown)


epoberezkin:
Great - thank you very much! I will work on it this week to merge – we will make a small release next week, hopefully it'll resolve the crashes on Android 12 we see!

avently:
> hopefully it'll resolve the crashes on Android 12 we see!

I don't think so. It can't fix all crashes (related to the background service) because the service can be started without looking at user's choice. For example, here is the service is trying to start without looking at any config value:
https://github.com/simplex-chat/simplex-chat/blob/master/apps/android/app/src/main/java/chat/simplex/app/MainActivity.kt#L61

Which means it WILL crash 100% once the timer fires. Maybe there is other places where the same can happen.

avently:
Also this info is important to understand: https://developer.android.com/training/monitoring-device-state/doze-standby.html#exemption-cases

Because I'm using [ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS](https://developer.android.com/reference/android/provider/Settings#ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS). There are some use-cases that is allowed and prohibited. In case of Simplex I think we can use this intent action

epoberezkin:
>  For example, here is the service is trying to start without looking at any config value:
https://github.com/simplex-chat/simplex-chat/blob/master/apps/android/app/src/main/java/chat/simplex/app/MainActivity.kt#L61

I meant once this is all taken into account, not with the current code.

epoberezkin:
> Also this info is important to understand: https://developer.android.com/training/monitoring-device-state/doze-standby.html#exemption-cases

> Because I'm using [ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS](https://developer.android.com/reference/android/provider/Settings#ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS). There are some use-cases that is allowed and prohibited. In case of Simplex I think we can use this intent action.

I got it now, it's definitely listed as allowed in this case - a messaging app that can't use FCM, so it's ok.

apps/android/app/src/main/java/chat/simplex/app/SimplexApp.kt
```diff
@@ -65,6 +64,9 @@ class SimplexApp: Application(), LifecycleEventObserver {
           if (!chatController.getRunServiceInBackground()) SimplexService.stop(applicationContext)
         Lifecycle.Event.ON_START ->
           SimplexService.start(applicationContext)
+        Lifecycle.Event.ON_RESUME ->
```
epoberezkin, line 67: Will it fire when the app is first started? I agree that for the users that already have the app running we want to move it here, so it shows on resume, and not in onCreate... But maybe it should stay in onCreate as well?

epoberezkin: I am thinking to change the logic in such a way that while the app is in the foreground, the service will not be running, user will only manage the setting whether it should run, and then when the app is STOPPED (goes into background), the service will be started (if the setting is on).

But: can a user just kill the app without it ever going into background? In which case the service would never start, even if it is configured to be on? In which case it may be better to keep it on even while the app is in foreground? Or is there some event that would fire and can be used when the app is killed? And, would it start on timer if the app is killed (via scheduled worker)?...

epoberezkin: also, what I see is that ON_START and ON_RESUME always fire together...

epoberezkin:
I pushed it here: https://github.com/simplex-chat/simplex-chat/pull/656

Wasn't sure how to add to that pr... Maybe I should have cloned your fork?

I changed notice conditions a bit, but what I couldn't figure out is how to check if ignore_optimisations was turned on  by the user after startActivity... Maybe I should do something with startActivityForResult?

epoberezkin:
closing here, merged via https://github.com/simplex-chat/simplex-chat/commit/36ef6df9fbbd3a102a0295dd66d0de3030f55237

PR #648, opened 2022-05-13
core: mark chat as read without specifying the items range, closes #647

### LINKED ISSUE #647: fr(mark-as-read): extend sdk to easily mark all messages in a chat as read
actual: /_read chat @x from=y to=z

wish list: /_read @chat @x

as discussed before, it would be convenient for app developers to not go through the the hassle of reading chat stats for @x... fetch it... iterate through chat item ids to find the latest id, aka to=z

this could easily be solved with a single sql update statement and would add much convenience.

epoberezkin:


tests/ChatTests.hs
```diff
@@ -152,6 +152,8 @@ testAddContact =
       -- read messages
       alice #$> ("/_read chat @2 from=1 to=100", id, "ok")
       bob #$> ("/_read chat @2 from=1 to=100", id, "ok")
+      alice #$> ("/_read chat @2", id, "ok")
+      bob #$> ("/_read chat @2", id, "ok")
```
jr-simplex, line 156: call /_read in testGroup

jr-simplex (approved):

PR #656, opened 2022-05-14
android: notification service and battery optimization on Android 12

epoberezkin:


epoberezkin:
@avently, could you please review if it all makes sense?

I've finally figured how it works - ON_RESUME picks up the execution after the power optimisation dialog - at first I thought it would only disable the service on the next app run.

I still added the second flag, so that the users who already have the app installed will be presented with the dialogue instead of having the service disabled.

Settings also work well.

I don't think it would crash now, the line you pointed to only restarts the service if it was previously running (it's yet one more flag:), that's why it doesn't check the setting... In any case, we will see if we have crashes after the release.

Thanks a lot!

avently:
https://github.com/simplex-chat/simplex-chat/pull/656/commits/d61babe92d5876a75eb92b299d0657af1f127147#diff-02c7df8f8b74f5f79ebb0e460c7db3a21a964f2577a07cb0450be51e32b132ebR60

Spelling error.

Replace this:
```
Для их использования, пожалуйста отключите оптимизацию батареи для <xliff:g id="appName">SimpleX</xliff:g> в слеующем диалоге. Иначе уведомления будут выключены.
```
With this:
```
Для их использования, пожалуйста, отключите оптимизацию батареи для <xliff:g id="appName">SimpleX</xliff:g> в следующем диалоге. Иначе уведомления будут выключены.
```

But the whole sentence it incorrect (you wrote "их использования") but at the same time in the same alert you're talking about the service which is "он", not "они". Т.е. ты объясняешь сначала про необходимость включить сервис, а потом говоришь "для их использования". Как-то нескладно выходит. 
```
Чтобы защитить ваши личные данные, вместо уведомлений от сервера приложение запускает <b>фоновый сервис <xliff:g id="appName">SimpleX</xliff:g></b>, который потребляет несколько процентов батареи в день.
```

here: https://github.com/simplex-chat/simplex-chat/pull/656/commits/d61babe92d5876a75eb92b299d0657af1f127147#diff-02c7df8f8b74f5f79ebb0e460c7db3a21a964f2577a07cb0450be51e32b132ebR58

Т.е. лучше бы рода слов согласовать. Либо прям указать конкретно, что разговор про уведомления, а не просто "Для их использования". Типа "Для использования приватных уведомлений..."

Don't you need to add a comma after word "please" here: https://github.com/simplex-chat/simplex-chat/pull/656/commits/d61babe92d5876a75eb92b299d0657af1f127147#diff-1a4f8f2996358e9a5bf137738c172de36e631a2c3d1151d6a2cbfb0f909c091cR60 ?

About `BackgroundServiceBatteryNoticeShown`. I would say that another setting just overcompicates the situation. Are you really need it? Harder to maintain such code that deeply depends on unexpected behavior with so many settings. My initial commit was quite small just for the reason of simplicity. The problem that you're trying to solve by adding another setting probably doesn't worth it. In initial commit a user will see the message that describes a problem and a solution. Which is easy to follow from my point of view. And at the same time the info alert describes why that user could have crashes in the previous versions and why he didn't have private notifications before.

Be aware that for now you have a small users base but this setting should be supported forever:)

So I would choose initial commit + your changes to translations (because I'm not super english speaker as you already know).

epoberezkin:
> Т.е. лучше бы рода слов согласовать.

👍

> Don't you need to add a comma after word "please" here: https://github.com/simplex-chat/simplex-chat/commit/d61babe92d5876a75eb92b299d0657af1f127147#diff-1a4f8f2996358e9a5bf137738c172de36e631a2c3d1151d6a2cbfb0f909c091cR60 ?

I think it's ok without comma here

> Be aware that for now you have a small users base but this setting should be supported forever:)

I think it's better both to migrate the existing users and for all other users who would be migrating from Android 11 to Android 12 in the future - I think there are much more A11 devices out there than 12...

Also, when I was testing on A12, I had notifications working even without disabling power optimisation (i.e., the app had power setting as "optimised", not as "unrestricted"), so without this extra option it can disable notifications for users for whom it was working, and they may not read the commit.

I think it's ok overall, not that much extra complexity

PR #657, opened 2022-05-14
mobile: hide app screen from recent apps

epoberezkin:


VictoriousFan:
This would be good functionality to have, Android and IOS save screenshots of the last state of the app when you go into recent-apps.

PR #659, opened 2022-05-16
terminal: support to connect via webrtc

epoberezkin:


src/Simplex/Chat.hs
```diff
@@ -2127,6 +2130,7 @@ chatCommandP =
     <|> "/_accept " *> (APIAcceptContact <$> A.decimal)
     <|> "/_reject " *> (APIRejectContact <$> A.decimal)
     <|> "/_call invite @" *> (APISendCallInvitation <$> A.decimal <* A.space <*> jsonP)
+    <|> ("/call @" <|> "/call ") *> (SendCallInvitation <$> displayName <*> pure defaultCallType)
```
jr-simplex, line 2133: Parse option to call with audio?

jr-simplex (approved):

src/Simplex/Chat.hs
epoberezkin: maybe later, it's for testing, UX is pretty horrible

PR #660, opened 2022-05-17
mobile: clear chat; allow to delete items deleted by sender

jr-simplex:


apps/ios/Shared/Views/ChatList/ChatListNavLink.swift
```diff
@@ -173,6 +188,17 @@ struct ChatListNavLink: View {
         )
     }
 
+    private func clearChatAlert() -> Alert {
+        Alert(
+            title: Text("Clear chat?"),
+            message: Text("All messages will be deleted - this cannot be undone! The messages will be deleted ONLY for you."),
```
epoberezkin, line 194:
```suggestion
            message: Text("All messages will be deleted for you ONLY - this cannot be undone!"),
```

maybe this?

apps/android/app/src/main/res/values/strings.xml
```diff
@@ -169,6 +169,11 @@
   <string name="accept_contact_button">Accept</string>
   <string name="reject_contact_button">Reject</string>
 
+  <!-- Clear Chat - ChatListNavLinkView.kt -->
+  <string name="clear_chat_question">Clear chat?</string>
+  <string name="clear_chat_warning">All messages will be deleted - this cannot be undone! The messages will be deleted ONLY for you.</string>
```
epoberezkin, line 174:
```suggestion
  <string name="clear_chat_warning">All messages will be deleted for you ONLY - this cannot be undone!</string>
```

PR #694, opened 2022-05-23
core: chat settings schema and types

epoberezkin:


src/Simplex/Chat/Migrations/M20220524_chat_settings.hs
```diff
@@ -0,0 +1,31 @@
...
+
+module Simplex.Chat.Migrations.M20220524_chat_settings where
+
+import Database.SQLite.Simple (Query)
+import Database.SQLite.Simple.QQ (sql)
+
+m20220524_chat_settings :: Query
+m20220524_chat_settings =
+  [sql|
+CREATE TABLE chat_settings (
```
jr-simplex, line 11: will all "users" have all settings and settings (not chat_setting_values) will be only added in migrations? I'm just wondering whether we need `user_id` field here

PR #698, opened 2022-05-25
android: update colors

epoberezkin:


apps/android/app/src/main/java/chat/simplex/app/views/chatlist/ChatListNavLinkView.kt
```diff
@@ -103,7 +104,8 @@ fun ContactMenuItems(chat: Chat, chatModel: ChatModel, showMenu: MutableState<Bo
     onClick = {
       clearChatDialog(chat.chatInfo, chatModel)
       showMenu.value = false
-    }
+    },
+    color = WarningOrange
```
jr-simplex, line 108: I'm not sure orange looks good here in dropdown menu.. it can be the only item / or this and "delete" in red. I'd keep it in regular color

jr-simplex (approved):

PR #701, opened 2022-05-27
mobile: call settings, request camera on iOS on call start

epoberezkin:


apps/android/app/src/main/assets/www/call.js
```diff
@@ -430,6 +434,10 @@ const processCommand = (function () {
             console.log(`no ${operation}`);
         }
     }
+    function getLocatMediaStream(mediaType, facingMode) {
```
jr-simplex, line 437:
```suggestion
    function getLocalMediaStream(mediaType, facingMode) {
```

PR #703, opened 2022-05-27
mobile: timeout call invitations, more android options

epoberezkin:


jr-simplex (approved):

apps/android/app/src/main/java/chat/simplex/app/views/usersettings/SettingsView.kt
```diff
@@ -135,6 +135,7 @@ fun SettingsLayout(
     Icon(
       Icons.Outlined.QrCode,
       contentDescription = stringResource(R.string.icon_descr_address),
+      tint = HighOrLowlight,
```
jr-simplex, line 138: Not sure about the color, I think default was better.. Maybe revert it?

PR #721, opened 2022-06-02
ios: fix authentication

epoberezkin:


apps/ios/Shared/ContentView.swift
```diff
@@ -48,28 +48,19 @@ struct ContentView: View {
...
                 }
             } else if prefPerformLA && laFailed {
-                retryAuthView()
-            }
-        }
-        .onChange(of: doAuthenticate) { doAuth in
-            if doAuth, authenticationExpired() {
-                runAuthenticate()
+                Button(action: runAuthenticate) { Label("Retry", systemImage: "arrow.counterclockwise") }
+            } else {
```
jr-simplex, line 52: Should `if prefPerformLA && laFailed` be split into separate ifs and its else branch only applied to `if laFailed`? Otherwise it's possible to offer authentication to a user who never set up the lock..

apps/ios/Shared/ContentView.swift
```diff
@@ -48,28 +48,19 @@ struct ContentView: View {
...
 
-    private func retryAuthView() -> some View {
-        Button {
-            laFailed = false
-            runAuthenticate()
-        } label: { Label("Retry", systemImage: "arrow.counterclockwise") }
-    }
-
     private func runAuthenticate() {
-        if !prefPerformLA {
```
jr-simplex, line 70: Why this had to be changed? If `prefPerformLA` is false we don't need other checks to set `userAuthorized = true`.. Then its else branch it could check for `if doAuthenticate && authenticationExpired()` separately

PR #739, opened 2022-06-16
core: fix dependencies, update nix

epoberezkin:


flake.nix
```diff
@@ -1,8 +1,11 @@
 {
   description = "nix flake for simplex-chat";
   inputs.nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable"; #angerman/nixpkgs/patch-1"; # based on 21.11, still need this, until everything is merged into 21.11.
-  inputs.haskellNix.url = "github:input-output-hk/haskell.nix?ref=angerman/try-no-libcharset";
-  inputs.haskellNix.inputs.nixpkgs.follows = "nixpkgs";
+  inputs.hackage = {
+    url = "github:input-output-hk/haskell.nix";
```
angerman, line 5: This should point to `hackage.nix`

flake.nix
```diff
@@ -1,8 +1,11 @@
 {
   description = "nix flake for simplex-chat";
   inputs.nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable"; #angerman/nixpkgs/patch-1"; # based on 21.11, still need this, until everything is merged into 21.11.
-  inputs.haskellNix.url = "github:input-output-hk/haskell.nix?ref=angerman/try-no-libcharset";
-  inputs.haskellNix.inputs.nixpkgs.follows = "nixpkgs";
```
angerman, line 5: We want to keep these.

angerman (changes requested):
We also want a line:

inputs.haskellNix.inputs.hackage = "hackage";

To ensure that haskell.nix knows about the new hackage.nix

flake.nix
epoberezkin: both lines?

PR #771, opened 2022-07-02
iOS: update call invitations when exiting background

epoberezkin:


src/Simplex/Chat.hs
```diff
@@ -870,9 +882,11 @@ processChatCommand = \case
           Just call@Call {contactId}
             | ctId == contactId -> do
               call_ <- action userId ct call
-              atomically $ case call_ of
-                Just call' -> TM.insert ctId call' calls
-                _ -> TM.delete ctId calls
+              case call_ of
+                Just call' -> atomically $ TM.insert ctId call' calls
```
epoberezkin, line 886: delete invitation when it is no longer an invitation here

src/Simplex/Chat.hs
```diff
@@ -1789,9 +1804,11 @@ processAgentMessage (Just user@User {userId, profile}) agentConnId agentMessage
           | contactId /= ctId' || callId /= callId' -> messageError $ eventName <> ": wrong contact or callId"
           | otherwise -> do
             (call_, aciContent_) <- action call
-            atomically $ case call_ of
-              Just call' -> TM.insert ctId' call' calls
-              _ -> TM.delete ctId' calls
+            case call_ of
+              Just call' -> atomically $ TM.insert ctId' call' calls
```
epoberezkin, line 1808: same here

src/Simplex/Chat/Call.hs
```diff
@@ -91,6 +113,22 @@ instance ToJSON CallId where
...
+  { contact :: Contact,
+    callType :: CallType,
+    sharedKey :: Maybe C.Key,
+    callTs :: UTCTime
+  }
+  deriving (Show, Generic)
+
+instance ToJSON RcvCallInvitation where
+  toJSON = J.genericToJSON J.defaultOptions
+  toEncoding = J.genericToEncoding J.defaultOptions
```
epoberezkin, line 130:
```suggestion
  toJSON = J.genericToJSON J.defaultOptions {J.omitNothingFields = True}
  toEncoding = J.genericToEncoding J.defaultOptions {J.omitNothingFields = True}
```

PR #776, opened 2022-07-04
ios: refresh call invitations and report call on start and activation; core: restore calls on activation

jr-simplex:


apps/ios/SimpleXChat/APITypes.swift
```diff
@@ -348,11 +353,12 @@ public enum ChatResponse: Decodable, Error {
             case let .sndFileCancelled(chatItem, _): return String(describing: chatItem)
             case let .sndFileRcvCancelled(chatItem, _): return String(describing: chatItem)
             case let .sndGroupFileCancelled(chatItem, _, _): return String(describing: chatItem)
-            case let .callInvitation(contact, callType, sharedKey, _): return "contact: \(contact.id)\ncallType: \(String(describing: callType))\nsharedKey: \(sharedKey ?? "")"
+            case let .callInvitation(inv): return "contact: \(inv.contact.id)\ncallType: \(String(describing: inv.callType))\nsharedKey: \(inv.sharedKey ?? "")"
```
epoberezkin, line 356:
```suggestion
            case let .callInvitation(inv): return String(describing: inv)
```

PR #806, opened 2022-07-12
core: groups api

jr-simplex:


src/Simplex/Chat.hs
```diff
@@ -574,6 +584,62 @@ processChatCommand = \case
         msgTs' = systemToUTCTime . (SMP.msgTs :: SMP.NMsgMeta -> SystemTime) <$> ntfMsgMeta
     connEntity <- withStore (\db -> Just <$> getConnectionEntity db user ntfConnId) `catchError` \_ -> pure Nothing
     pure CRNtfMessages {connEntity, msgTs = msgTs', ntfMessages}
+  APIAddMember groupId contactId memRole -> withUser $ \user@User {userId} -> withChatLock $ do
+    -- TODO for large groups: no need to load all members to determine if contact is a member
```
epoberezkin, line 588: maybe fix it? it's now easier to load group info instead of the whole group - there is a function for that.

epoberezkin (approved):

PR #816, opened 2022-07-15
mobile: group invitations ui

jr-simplex:


epoberezkin:
IMG_215F4BA28788-1

apps/android/app/src/main/java/chat/simplex/app/model/ChatModel.kt
```diff
@@ -428,6 +430,9 @@ class Contact(
   override val fullName get() = profile.fullName
   override val image get() = profile.image
 
+  val isIndirectContact: Boolean get() =
+    activeConn.connLevel > 0 || viaGroup != null
```
epoberezkin, line 434: I am not sure viaGroup is needed here, I think connLevel can be enough.

PR #831, opened 2022-07-22
readme: technical details

epoberezkin:


README.md
```diff
@@ -122,6 +122,27 @@ Only the client devices have information about users, their contacts and groups.
 
 See [SimpleX whitepaper](https://github.com/simplex-chat/simplexmq/blob/master/protocol/overview-tjr.md) for more information on platform objectives and technical design.
 
+## Privacy: technical details and limitations
```
jr-simplex, line 125: add to contents

README.md
```diff
@@ -122,6 +122,27 @@ Only the client devices have information about users, their contacts and groups.
...
+
+SimpleX Chat is a work in progress, we are releasing improvements as they are ready. You have to decide if the current state is good enough for your scenarios.
+
+What is already implemented:
+
+1. Instead of user profile identifiers used by all other platforms, even the most private, SimpleX uses pairwise per-queue identifiers (2 addresses for each message queue, 2 queues in each connection between the users) . It makes understanding of network graph more difficult on the application level, as for `n` users you have `n * (n-1)` message queues.
```
jr-simplex, line 131:
```suggestion
1. Instead of user profile identifiers used by all other platforms, even the most private, SimpleX uses pairwise per-queue identifiers (2 addresses for each message queue, 2 queues in each connection between the users). It makes understanding of network graph more difficult on the application level, as for a complete graph of `n` users you have `n * (n-1)` message queues.
```

README.md
```diff
@@ -122,6 +122,27 @@ Only the client devices have information about users, their contacts and groups.
...
+2. End-to-end encryption in each message queue using [NaCl cryptobox](https://nacl.cr.yp.to/box.html). This is added to allow redundancy in the future (passing each message via several servers), to avoid having the same ciphertext in different queues (that would only be visible to the attacker if TLS is compromised). The encryption keys used for this level are not rotated, instead we are planning to replace the queues. Curve25519 keys are used for key negotiation.
```
jr-simplex, line 132: "rotate" is better? "replace" sounds like one-time
```suggestion
2. End-to-end encryption in each message queue using [NaCl cryptobox](https://nacl.cr.yp.to/box.html). This is added to allow redundancy in the future (passing each message via several servers), to avoid having the same ciphertext in different queues (that would only be visible to the attacker if TLS is compromised). The encryption keys used for this level are not rotated, instead we are planning to rotate the queues. Curve25519 keys are used for key negotiation.
```

README.md
```diff
@@ -122,6 +122,27 @@ Only the client devices have information about users, their contacts and groups.
...
+4. Additional layer of encryption using NaCL cryptobox for the messages delivered from the server to the recipient. This layer avoids having any ciphertext in common between sent and received traffic of the server (and there are no identifiers in common as well) inside TLS.
+5. Several levels of content padding to the fixed size - server blocks inside TLS and message content inside each of 3 encrypted envelopes are padded to a constant size, to prevent any correlation by message size.
+6. Starting from v2 of SMP protocol (the current version is v4) all message metadata, including the time message was received by the server (rounded to a second) is inside encrypted envelope, so even if TLS is compromised it cannot be observed.
+7. Only TLS 1.3 is used, with two certificate to provide server identity verification, with tls-unique used as session - it is used in each client command and signed over by the per-queue ephemeral key.
```
jr-simplex, line 137:
```suggestion
7. Only TLS 1.3 is used, with two certificate chain to provide server identity verification, with tls-unique used as session - it is used in each client command and signed over by the per-queue ephemeral key.
```

README.md
```diff
@@ -122,6 +122,27 @@ Only the client devices have information about users, their contacts and groups.
...
+
+What we plan to provide soon:
+
+1. Access to messaging servers via Tor. Currently clients access servers via public Internet, and the servers can observe IP addresses of the clients. Depending which platform you use, you might be able to configure access via tor independently. The servers provided by SimpleX Chat do not log IP addresses, and you can use your own servers, but in some scenarios that may be not a sufficient level of privacy.
```
jr-simplex, line 141:
```suggestion
1. Access to messaging servers via Tor. Currently clients access servers via public Internet, and the servers can observe IP addresses of the clients. Depending on which platform you use, you might be able to configure access via tor independently. The servers provided by SimpleX Chat do not log IP addresses, and you can use your own servers, but in some scenarios that may be not a sufficient level of privacy.
```

README.md
epoberezkin: should be fully connected, but better "you can have up to"... will change

README.md
```diff
@@ -124,23 +125,24 @@ See [SimpleX whitepaper](https://github.com/simplex-chat/simplexmq/blob/master/p
...
 5. Several levels of content padding to the fixed size - server blocks inside TLS and message content inside each of 3 encrypted envelopes are padded to a constant size, to prevent any correlation by message size.
-6. Starting from v2 of SMP protocol (the current version is v4) all message metadata, including the time message was received by the server (rounded to a second) is inside encrypted envelope, so even if TLS is compromised it cannot be observed.
-7. Only TLS 1.3 is used, with two certificate to provide server identity verification, with tls-unique used as session - it is used in each client command and signed over by the per-queue ephemeral key.
+6. Starting from v2 of SMP protocol (the current version is v4) all message metadata, including the time message was received by the server (rounded to a second) is sent to the recipients inside an encrypted envelope, so even if TLS is compromised it cannot be observed.
+7. Only TLS 1.2/1.3 are allowed used for client-server connections, limited to cryptographic algorithms: CHACHA20POLY1305_SHA256, Ed25519/Ed448, Curve25519/Curve448.
```
jr-simplex, line 138:
```suggestion
7. Only TLS 1.2/1.3 are allowed for client-server connections, limited to cryptographic algorithms: CHACHA20POLY1305_SHA256, Ed25519/Ed448, Curve25519/Curve448.
```

README.md
```diff
@@ -124,23 +125,24 @@ See [SimpleX whitepaper](https://github.com/simplex-chat/simplexmq/blob/master/p
...
-2. Queue rotation. Currently the queues created between two users are used until the contact is deleted, providing a long-term pairwise identifiers of the conversation. We are planning to add queue rotation to make these identifiers termporary and rotate based on some schedule TBC (e.g., every X messages, or every X hours/days).
-3. Local database encryption. Currently the local chat database is not encrypted.
+1. Access to messaging servers via Tor. Currently clients access servers via public Internet, and the servers can observe IP addresses of the clients. Depending which platform you use, you might be able to configure access via Tor independently. The servers provided by SimpleX Chat do not do not correlate users by or log IP addresses, and you can use your own servers, but in some scenarios that may be not a sufficient level of privacy.
```
jr-simplex, line 143:
```suggestion
1. Access to messaging servers via Tor. Currently clients access servers via public Internet, and the servers can observe IP addresses of the clients. Depending which platform you use, you might be able to configure access via Tor independently. The servers provided by SimpleX Chat do not correlate users by or log IP addresses, and you can use your own servers, but in some scenarios that may be not a sufficient level of privacy.
```

PR #834, opened 2022-07-22
ios: add member ui wip

jr-simplex:


apps/ios/Shared/ContentView.swift
```diff
@@ -64,11 +63,11 @@ struct ContentView: View {
...
-        } else if showChatInfo {
-            showChatInfo = false
-            DispatchQueue.main.async {
-                justAuthenticate()
-            }
+//        } else if showChatInfo {
+//            showChatInfo = false
+//            DispatchQueue.main.async {
+//                justAuthenticate()
+//            }
```
jr-simplex, line 70: we're going to have more sheet views opening from chat view for groups - we should look for a less fiddly solution. maybe at least a single binding for all such views affecting them via onChange modifiers

apps/ios/Shared/Views/Chat/Group/AddGroupMembersView.swift
```diff
@@ -0,0 +1,102 @@
...
+                .font(.headline)
+                .padding()
+            if (unaddedContacts.isEmpty) {
+                Text("No unadded contacts")
+                    .foregroundColor(.secondary)
+            } else {
+                HStack {
+                    Button {
+                        selectedContactIds.forEach { contactId in
+                            Task {
```
epoberezkin, line 31: what's the reason to create multiple tasks here, instead of iterating inside a single task? I only see downsides.

epoberezkin: also, forEach should generally not be used other than for view construction - there is `for in` loop for that.

apps/ios/Shared/Views/Chat/Group/AddGroupMembersView.swift
```diff
@@ -0,0 +1,102 @@
...
+                                await addMember(groupId: groupId, contactId: contactId)
+                            }
+                            chatViewSheet = nil
+                        }
+                    } label: {
+                        Label("Add members", systemImage: "plus.circle.fill")
+                    }
+                    .padding()
+                    Spacer()
```
epoberezkin, line 40: do not use Spacer for alignment like that - it should be alignment parameter in HStack

apps/ios/Shared/Model/SimpleXAPI.swift
```diff
@@ -567,6 +587,12 @@ func apiLeaveGroup(groupId: Int64) async throws -> GroupInfo {
     throw r
 }
 
+func apiListMembers(groupId: Int64) -> [GroupMember] {
```
epoberezkin, line 590: we did discuss that this function must be async

epoberezkin: we only use blocking API calls on chat start I think, and in one more case where it's difficult to make UI work otherwise

apps/ios/Shared/Views/Chat/Group/AddGroupMembersView.swift
```diff
@@ -0,0 +1,102 @@
...
+            .compactMap{ $0.chatInfo.contact }
+            .filter{ !memberContactIds.contains($0.apiId) }
+            .sorted{ $0.displayName < $1.displayName }
+        return unaddedContacts
+    }
+
+    struct ContactCheckView: View {
+        var contact: Contact
+        @State var checked: Bool = false
+        @Binding var selectedContactIds: Set<Int64>
```
epoberezkin, line 69: I don't think it needs to be passed as parameter, it's in scope

apps/ios/Shared/Model/SimpleXAPI.swift
jr-simplex: I think it's from the attempts of reading it on button press or rather sheet opening.. I didn't manage to make it async there..

PR #850, opened 2022-07-28
android: groups ui

jr-simplex:


apps/android/app/src/main/java/chat/simplex/app/model/ChatModel.kt
```diff
@@ -563,6 +574,14 @@ class GroupMember (
   val memberContactId: Long? = null,
   var activeConn: Connection? = null
 ) {
+  val id: String get() = "#$groupId @$groupMemberId"
```
epoberezkin, line 577: Is this used? iOS uses it in notifications

jr-simplex: no - was just translating what was missing in types to not get distracted afterwards

PR #851, opened 2022-07-29
core: editing group profiles (no conflict resolution)

epoberezkin:


src/Simplex/Chat/Messages.hs
```diff
@@ -539,6 +541,7 @@ data RcvGroupEvent
   | RGEMemberDeleted {groupMemberId :: GroupMemberId, profile :: Profile} -- CRDeletedMember
   | RGEUserDeleted -- CRDeletedMemberUser
   | RGEGroupDeleted -- CRGroupDeleted
+  | RGEGroupProfileUpdated -- CRGroupProfileUpdated
```
jr-simplex, line 544: maybe have profile (or both old and new) in CIContent?

epoberezkin: added

apps/android/.idea/gradle.xml
```diff
@@ -13,6 +13,7 @@
             <option value="$PROJECT_DIR$/app" />
           </set>
         </option>
+        <option name="resolveModulePerSourceSet" value="false" />
```
epoberezkin, line 16:
```suggestion
```

PR #864, opened 2022-08-01
ci: script for downloading and unpacking prebuilt aarch64 libs for Android

avently:
Can be called from any directory like so:
```bash
path-to-simplex-source/scripts/android/download_libs_aarch64.sh https://something.com/job/something
```


ata-star:
can this be modified to use an armv7 lib?

avently:
@ata-star sorry for such late answer, didn't see your comment before. 
Script can be modified but the problem is that there is no armv7 lib built. As you can see here: https://github.com/simplex-chat/simplex-chat/blob/stable/flake.nix#L13 there is only x64 builds are made. You can probably try to change that and rebuild the lib yourself but it's not an easy task, so. You can start experimenting by executing this script: https://github.com/simplex-chat/simplex-chat/blob/stable/scripts/android/build-android.sh

If you will be able to build aarch64, then you can experiment with armv7 by modifying both scripts. I didn't try it and can't say if it works or not. Haskell for Android is kind of miracle and don't really easy to use.

PR #871, opened 2022-08-02
Removed gesture interception while long clicking on a chat bubble

avently:
- allowed to skip motion event consuming based on touch offset
- long clicking on a link copies it to a clipboard


apps/android/app/src/main/java/chat/simplex/app/views/chat/item/ChatItemView.kt
```diff
@@ -53,6 +54,13 @@ fun ChatItemView(
       .fillMaxWidth(),
     contentAlignment = alignment,
   ) {
+    val onLinkLongClick = { link: String ->
+      val clipboard = context.getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
+      val clip: ClipData = ClipData.newPlainText("link", link)
+      clipboard.setPrimaryClip(clip)
+      Toast.makeText(context, context.getString(R.string.copied_to_clipboard, link), Toast.LENGTH_SHORT).show()
+    }
```
epoberezkin, line 62:
```suggestion
```

I understand the idea, but I think we are overloading the function, and the is frequently the only thing there is in the message - so it overcomplicates UX without much reason... It is better not to have different functions on long press - it should open the message menu even if it is on the link.

avently: Done

PR #873, opened 2022-08-02
ios: advanced network settings

epoberezkin:


apps/ios/Shared/Views/UserSettings/AdvancedNetworkSettings.swift
```diff
@@ -0,0 +1,125 @@
...
+                        resetNetCfg()
+                        saveNetCfg()
+                    } label: {
+                        Text("Reset to defaults")
+                    }
+                    .disabled(currentNetCfg == NetCfg.defaults)
+
+                    timeoutSettingPicker("TCP connection timeout", selection: $netCfg.tcpConnectTimeout, values: [2_500000, 5_000000, 7_500000, 10_000000, 15_000000, 20_000000], label: secondsLabel)
+                    timeoutSettingPicker("Protocol timeout", selection: $netCfg.tcpTimeout, values: [1_500000, 3_000000, 5_000000, 7_000000, 10_000000, 15_000000], label: secondsLabel)
+                    timeoutSettingPicker("PING interval", selection: $netCfg.smpPingInterval, values: [120_000000, 300_000000, 600_000000, 1200_000000, 2400_000000], label: secondsLabel)
```
jr-simplex, line 38: maybe we could do these conversions somewhere in api (or even in backend but I'm not offering to move it there now).. just seems strange that top level ui needs to know about low level implementation scales

apps/ios/Shared/Views/UserSettings/SettingsView.swift
```diff
@@ -121,10 +121,10 @@ struct SettingsView: View {
                         Toggle("Show pending connections", isOn: $pendingConnections)
                     }
                     NavigationLink {
-                        SMPServers()
-                            .navigationTitle("Your SMP servers")
+                        NetworkAndServers()
+                            .navigationTitle("Network & servers")
                     } label: {
-                        settingsRow("server.rack") { Text("SMP servers") }
+                        settingsRow("point.3.filled.connected.trianglepath.dotted") { Text("Network & servers") }
```
jr-simplex, line 127: maybe this one?
```suggestion
                        settingsRow("externaldrive.connected.to.line.below") { Text("Network & servers") }
```

apps/ios/Shared/Views/UserSettings/NetworkAndServers.swift
```diff
@@ -0,0 +1,39 @@
...
+                            .navigationTitle("Your SMP servers")
+                    } label: {
+                        settingsRow("server.rack") { Text("SMP servers") }
+                    }
+
+                    NavigationLink {
+                        AdvancedNetworkSettings()
+                            .navigationTitle("Advanced settings")
+                    } label: {
+                        settingsRow("app.connected.to.app.below.fill") { Text("Advanced network settings") }
```
jr-simplex, line 27: ?
```suggestion
                        settingsRow("gearshape.2") { Text("Advanced network settings") }
```

apps/ios/Shared/Views/UserSettings/AdvancedNetworkSettings.swift
jr-simplex: I think at least we could make a local constant / a function to use in these lists, e.g. `toMicrosec = 1_000_000`? I just feel it's really easy to make an error here

apps/ios/Shared/Views/UserSettings/NetworkAndServers.swift
epoberezkin: don't like this one

apps/ios/Shared/Views/UserSettings/AdvancedNetworkSettings.swift
epoberezkin: then the numbers would have to be doubles. I tried map, tried different things, this seems the least bad...

PR #875, opened 2022-08-03
Button in notification that routes to settings for notification channel

avently:
Now users are able to hide a notification without finding the app in the settings. At least now they will be aware of such possibility. I could write `setup` instead of `hide` (which is closer to actual situation) but in this case users could misunderstand of what they will be trying to setup.


avently:
Changed the code because Android 12 introduced new rules for working with activities that are started from service. This implementation should be ideal.

PR #891, opened 2022-08-04
terminal: update active group when message is updated

epoberezkin:


jr-simplex (dismissed):

src/Simplex/Chat.hs
```diff
@@ -1731,6 +1731,7 @@ processAgentMessage (Just user@User {userId, profile}) agentConnId agentMessage
             then do
               updCi <- withStore $ \db -> updateGroupChatItem db user groupId itemId (CIRcvMsgContent mc) msgId
               toView . CRChatItemUpdated $ AChatItem SCTGroup SMDRcv (GroupChat gInfo) updCi
+              setActive $ ActiveG gInfo
```
epoberezkin, line 1734:
```suggestion
              let g = groupName' gInfo              
              setActive $ ActiveG g
```

PR #894, opened 2022-08-04
App icon chooser

avently:
- ability to choose an icon from a predefined list

Icons needs to be added:
- icon_dark_blue.png
- icon_foreground_dark_blue.png
- icon_round_dark_blue.png

`dark_blue` is an icon name. It's used inside `AndroidManifest.xml` as a part of a name of activity-alias and inside `AppIcon` enum. Default icon called `default`.

Be aware that since we use such mechanism of changing icons, we should be care with activity-aliases. One activity-alias should ALWAYS be enabled. Otherwise the app will not start.

Tested on Android 12.


avently:
I would not merge it immediatelly because I'm worring about probability to get the app without enabled activity-alias. I relaunched it so many times but still unsure. It's kind of stupid way to do things but there is no alternative

avently:
Force pushed and made the code much smarter than before. Now I like it.
Can be tested & merged.

apps/android/app/src/main/AndroidManifest.xml
```diff
@@ -38,7 +38,6 @@
         android:windowSoftInputMode="adjustResize"
         android:theme="@style/Theme.SimpleX">
       <intent-filter>
-        <action android:name="android.intent.action.MAIN" />
```
epoberezkin, line 41: why is it removed? I think something might depend on it, don't remember exactly

avently: Activity-aliases have this action. They receive it

PR #915, opened 2022-08-08
ios: update group members in model

jr-simplex:


apps/ios/Shared/Model/SimpleXAPI.swift
```diff
@@ -567,9 +567,10 @@ func apiNewGroup(_ p: GroupProfile) throws -> GroupInfo {
     throw r
 }
 
-func addMember(groupId: Int64, contactId: Int64, memberRole: GroupMemberRole) async {
+func addMember(groupInfo: GroupInfo, contactId: Int64, memberRole: GroupMemberRole) async {
```
epoberezkin, line 570: that's not a good idea in general to handle errors here, as you cannot show any alerts here - why can't it be handled at call site?

apps/ios/Shared/Views/Chat/Group/GroupChatInfoView.swift
```diff
@@ -33,6 +32,10 @@ struct GroupChatInfoView: View {
 
     var body: some View {
         NavigationView {
+            let members = chatModel.groupMembers
+                .filter { $0.memberStatus != .memLeft && $0.memberStatus != .memRemoved }
```
epoberezkin, line 36: .currentMember ?

apps/ios/Shared/Model/SimpleXAPI.swift
```diff
@@ -567,20 +567,18 @@ func apiNewGroup(_ p: GroupProfile) throws -> GroupInfo {
...
     if case .sentGroupInvitation = r { return }
     throw r
 }
 
+//func apiAddMember(_ groupId: Int64, _ contactId: Int64, _ memberRole: GroupMemberRole) async throws -> GroupMember {
+//    let r = await chatSendCmd(.apiAddMember(groupId: groupId, contactId: contactId, memberRole: memberRole))
+//    if case let .sentGroupInvitation(_, _, member) = r { return member }
+//    throw r
+//}
+
```
epoberezkin, line 581:
```suggestion
```

PR #920, opened 2022-08-09
Debug package suffix and ability to override Gradle variables

avently:
- now debug builds will have `.debug` suffix by default. It allows to have multiple app builds (debug and release) on the same device. If you don't need this, create a file local.properties and add `application_id_suffix=` into it
- now everyone can override some variables from top-level build.gradle file. For example, gradle_plugin_version, debuggable manifest attribute, and so on. Overriding Gradle plugin version is useful for those who uses Intellij IDEA with older Gradle plugin than in Android Studio.

There is no behaviour changes for release build and no changes in versions of plugins/libs. 
There is a change for debug build which can be easily returned to previous behavior by adding `application_id_suffix=` into local.properties file.

### DIFF
--- a/apps/android/build.gradle
+++ b/apps/android/build.gradle
@@ -1,14 +1,34 @@
 buildscript {
+    Properties localProperties = new Properties()
+    if (rootProject.file('local.properties').canRead()) {
+        localProperties.load(rootProject.file("local.properties").newDataInputStream())
+    }
+
     ext {
-        compose_version = '1.2.0-beta02'
+        compose_version = localProperties['compose_version'] ?: '1.2.0-beta02'
+        kotlin_version = localProperties['kotlin_version'] ?: '1.6.21'
+        gradle_plugin_version = localProperties['gradle_plugin_version'] ?: '7.2.0'
+
+        // Whether the app is debuggable or not. Specify `false` if you want good performance in debug builds
+        enable_debuggable = localProperties['debuggable'] ?: true
+        // Ending part of package name. Allows debug & release versions to coexist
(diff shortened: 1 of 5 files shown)


apps/android/build.gradle
```diff
@@ -1,25 +1,47 @@
...
+    }
+
     ext {
-        compose_version = '1.2.0-beta02'
+        compose_version = localProperties['compose_version'] ?: '1.2.0-beta02'
+        kotlin_version = localProperties['kotlin_version'] ?: '1.6.21'
+        gradle_plugin_version = localProperties['gradle_plugin_version'] ?: '7.2.0'
+
+        // Name that will be shown for debug build
+        app_name = localProperties['app_name'] ?: "SimpleX Debug"
```
epoberezkin, line 13:
```suggestion
        app_name = localProperties['app_name'] ?: "SimpleX"
```

apps/android/build.gradle
```diff
@@ -1,25 +1,47 @@
...
+        // Whether the app is debuggable or not. Specify `false` if you want good performance in debug builds
+        enable_debuggable = localProperties['debuggable'] ?: true
```
epoberezkin, line 15:
```suggestion
        enable_debuggable = localProperties['debuggable'] ?: false
```

apps/android/build.gradle
```diff
@@ -1,25 +1,47 @@
...
+        // Ending part of package name. Allows debug & release versions to coexist
+        application_id_suffix = localProperties['application_id_suffix'] ?: '.debug'
```
epoberezkin, line 17:
```suggestion
        application_id_suffix = localProperties['application_id_suffix'] ?: ''
```

PR #925, opened 2022-08-10
Endless scrolling in a chat view

avently:
- scroll position when you open keyboard/change screen orientation will remain the same
- scrolling to top will show messages from history
- unread messages will be positioned at the top of the screen

### DIFF
--- a/apps/android/app/src/main/java/chat/simplex/app/model/ChatModel.kt
+++ b/apps/android/app/src/main/java/chat/simplex/app/model/ChatModel.kt
@@ -211,25 +211,32 @@ class ChatModel(val controller: ChatController) {
     }
   }
 
-  fun markChatItemsRead(cInfo: ChatInfo) {
+  fun markChatItemsRead(cInfo: ChatInfo, range: CC.ItemRange? = null) {
+    // update current chat
+    var markedRead = 0
+    if (chatId.value == cInfo.id) {
+      var i = 0
+      while (i < chatItems.count()) {
+        val item = chatItems[i]
+        if (item.meta.itemStatus is CIStatus.RcvNew && (range == null || (range.from..range.to).contains(item.id))) {
+          chatItems[i] = item.withStatus(CIStatus.RcvRead())
+          markedRead++
+        }
+        i += 1
+      }
+    }
     // update preview
     val chatIdx = getChatIndex(cInfo.id)
     if (chatIdx >= 0) {
       val chat = chats[chatIdx]
       val lastId = chat.chatItems.lastOrNull()?.id
       if (lastId != null) {
-        chats[chatIdx] = chat.copy(chatStats = chat.chatStats.copy(unreadCount = 0, minUnreadItemId = lastId + 1))
-      }
-    }
-    // update current chat
-    if (chatId.value == cInfo.id) {
-      var i = 0
-      while (i < chatItems.count()) {
-        val item = chatItems[i]
-        if (item.meta.itemStatus is CIStatus.RcvNew) {
-          chatItems[i] = item.withStatus(CIStatus.RcvRead())
-        }
-        i += 1
+        chats[chatIdx] = chat.copy(
+          chatStats = chat.chatStats.copy(
+            unreadCount = if (range != null) chat.chatStats.unreadCount - markedRead else 0,
+            minUnreadItemId = if (range != null) kotlin.math.max(chat.chatStats.minUnreadItemId, range.to + 1) else lastId + 1
+          )
+        )
       }
     }
   }
--- a/apps/android/app/src/main/java/chat/simplex/app/model/ChatModel.kt
+++ b/apps/android/app/src/main/java/chat/simplex/app/model/ChatModel.kt
@@ -212,19 +212,7 @@ class ChatModel(val controller: ChatController) {
   }
 
   fun markChatItemsRead(cInfo: ChatInfo, range: CC.ItemRange? = null) {
-    // update current chat
-    var markedRead = 0
-    if (chatId.value == cInfo.id) {
-      var i = 0
-      while (i < chatItems.count()) {
-        val item = chatItems[i]
-        if (item.meta.itemStatus is CIStatus.RcvNew && (range == null || (range.from..range.to).contains(item.id))) {
-          chatItems[i] = item.withStatus(CIStatus.RcvRead())
-          markedRead++
-        }
-        i += 1
-      }
-    }
+    val markedRead = markItemsReadInCurrentChat(cInfo.id, range)
     // update preview
     val chatIdx = getChatIndex(cInfo.id)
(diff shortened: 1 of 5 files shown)


apps/android/app/src/main/java/chat/simplex/app/model/ChatModel.kt
```diff
@@ -211,27 +211,34 @@ class ChatModel(val controller: ChatController) {
     }
   }
 
-  fun markChatItemsRead(cInfo: ChatInfo) {
-    // update preview
```
epoberezkin, line 215: please revert the order change so I can see the diff

apps/android/app/src/main/java/chat/simplex/app/model/ChatModel.kt
```diff
@@ -211,27 +211,34 @@ class ChatModel(val controller: ChatController) {
...
-    }
+  fun markChatItemsRead(cInfo: ChatInfo, range: CC.ItemRange? = null) {
     // update current chat
+    var markedRead = 0
     if (chatId.value == cInfo.id) {
       var i = 0
       while (i < chatItems.count()) {
         val item = chatItems[i]
-        if (item.meta.itemStatus is CIStatus.RcvNew) {
+        if (item.meta.itemStatus is CIStatus.RcvNew && (range == null || (range.from..range.to).contains(item.id))) {
```
epoberezkin, line 221:
```suggestion
        if (item.meta.itemStatus is CIStatus.RcvNew && (range == null || (range.from > item.id && item.id < range.to)) {
```

or, if the range should be inclusive replace with >=, <= - there is no reason to construct the range here.

apps/android/app/src/main/java/chat/simplex/app/model/ChatModel.kt
```diff
@@ -211,27 +211,34 @@ class ChatModel(val controller: ChatController) {
...
+    // update preview
+    val chatIdx = getChatIndex(cInfo.id)
+    if (chatIdx >= 0) {
+      val chat = chats[chatIdx]
+      val lastId = chat.chatItems.lastOrNull()?.id
+      if (lastId != null) {
+        chats[chatIdx] = chat.copy(
+          chatStats = chat.chatStats.copy(
+            unreadCount = if (range != null) chat.chatStats.unreadCount - markedRead else 0,
+            minUnreadItemId = if (range != null) kotlin.math.max(chat.chatStats.minUnreadItemId, range.to + 1) else lastId + 1
```
epoberezkin, line 237: once this part is back where it was it will be clear what's changed... But it looks as potentially incorrect for groups, because of the different sorting order.

apps/android/app/src/main/java/chat/simplex/app/views/chat/ChatView.kt
```diff
@@ -20,60 +19,44 @@ import androidx.compose.ui.Modifier
...
 import com.google.accompanist.insets.ProvideWindowInsets
 import com.google.accompanist.insets.navigationBarsWithImePadding
 import kotlinx.coroutines.*
+import kotlinx.coroutines.flow.*
 import kotlinx.datetime.Clock
 
 @Composable
 fun ChatView(chatModel: ChatModel) {
-  val chat: Chat? = chatModel.chats.firstOrNull { chat -> chat.chatInfo.id == chatModel.chatId.value }
+  var activeChat by remember { mutableStateOf(chatModel.chats.firstOrNull { chat -> chat.chatInfo.id == chatModel.chatId.value }) }
```
epoberezkin, line 46: why does active chat needs to be mutable?

avently: https://github.com/simplex-chat/simplex-chat/pull/925/files#diff-8187df6fb047b579372e5e8df26f6a71db30123fad806cb7e0455bdd2d163a4dR116

apps/android/app/src/main/java/chat/simplex/app/views/chat/ChatView.kt
```diff
@@ -118,11 +101,22 @@ fun ChatView(chatModel: ChatModel) {
           }
         }
       },
-      openDirectChat = { contactId ->
+      showChat = { apiId, chatType, pagination ->
```
epoberezkin, line 104: remove this coupling please - separate function that loads the initial chat (as it was) and add something like `loadMoreItem` that loads additional items on top - using API parameters (pagination) to manage view is not a good approach.

apps/android/app/src/main/java/chat/simplex/app/views/chat/ChatView.kt
```diff
@@ -118,11 +101,22 @@ fun ChatView(chatModel: ChatModel) {
...
         val c = chatModel.chats.firstOrNull {
-          it.chatInfo is ChatInfo.Direct && it.chatInfo.contact.contactId == contactId
+          it.chatInfo.chatType == chatType && it.chatInfo.apiId == apiId
+        }
+        if (c != null) {
+          // If there are more unread messages than pagination.count wants, load all unread
+          if (pagination is ChatPagination.Last && c.chatStats.unreadCount > pagination.count) {
```
epoberezkin, line 110: this block of code actually shows why it is not approach - first you have to pass this end then to use if to decide what to do - it just needs two different functions.

apps/android/app/src/main/java/chat/simplex/app/views/chat/ChatView.kt
```diff
@@ -389,17 +423,49 @@ fun ChatItemsList(
...
       }
-    }
-    val len = chatItems.count()
-    if (len > 1 && (keyboardState != ciListState.value.keyboardState || !ciListState.value.scrolled || len != ciListState.value.itemCount)) {
-      scope.launch {
-        ciListState.value = CIListState(true, len, keyboardState)
-        listState.animateScrollToItem(len - 1)
+
+      if (cItem.isRcvNew) {
+        LaunchedEffect(cItem.id) {
```
epoberezkin, line 428: the logic around keyboardState appears to be lost here?

apps/android/app/src/main/java/chat/simplex/app/model/ChatModel.kt
avently: I didn't do it just because I like that order. 

https://github.com/simplex-chat/simplex-chat/pull/925/files#diff-680099d115984a292f562ee96bf0e78965557165b4dd49b2aa20c80383ad33aeR216

This line shows that I need to calculate `markedRead` first before using this variable here: https://github.com/simplex-chat/simplex-chat/pull/925/files#diff-680099d115984a292f562ee96bf0e78965557165b4dd49b2aa20c80383ad33aeR236

Since I need to calculate how many I marked as read. Now there is possibility to mark individual items as read not just the whole chat.

PR #929, opened 2022-08-11
Floating button with unread counter and go to bottom action

avently:
- shows floating button with down arrow when lazy column is scrolled to at least second message
- shows floating button with a counter when there are unread messages


apps/android/app/src/main/java/chat/simplex/app/views/chat/ChatView.kt
```diff
@@ -375,8 +379,8 @@ fun BoxWithConstraintsScope.ChatItemsList(
 
   Spacer(Modifier.size(8.dp))
 
+  val reversedChatItems by remember { derivedStateOf { chatItems.reversed() } }
```
epoberezkin, line 382: This will be a very slow operation on a long chat - I've seen it even on a much faster iPhone.

If the reversed list is really needed it has to be maintained, and not created on demand from the whole list.

apps/android/app/src/main/java/chat/simplex/app/views/helpers/GestureDetector.kt
```diff
@@ -176,3 +179,49 @@ private class PressGestureScopeImpl(
...
+      if (topLevelInteraction is PressInteraction.Release)
+        onClick()
+      Log.e(TAG, ex.stackTraceToString())
+    } catch (ex: Exception) {
+      // Should never be called
+      Log.e(TAG, ex.stackTraceToString())
+    }
+  }
+  return interactionSource
+}
```
epoberezkin, line 227:
```suggestion
}

```

apps/android/app/src/main/java/chat/simplex/app/model/SimpleXAPI.kt
```diff
@@ -1279,7 +1279,7 @@ sealed class ChatPagination {
 
   companion object {
     const val INITIAL_COUNT = 100
-    const val PRELOAD_COUNT = 20
+    const val PRELOAD_COUNT = 50
     const val UNTIL_PRELOAD_COUNT = 10
```
epoberezkin, line 1283:
```suggestion
    const val PRELOAD_COUNT = 100
    const val UNTIL_PRELOAD_COUNT = 50
```

because it's bumping into the end on long chats

epoberezkin: ... and the cost of query is not proportionally higher to the number of loaded items, the large part of the cost is a "set up cost" that depends on the chat size, so it's better to load more items, less frequently, and start loading earlier...

apps/android/app/src/main/java/chat/simplex/app/views/chat/ChatView.kt
epoberezkin: seems fast enough on Android

PR #936, opened 2022-08-15
Swipe to reply feature

avently:
- ability to reply by swiping on a message from right-to-left or left-to-right
- keyboard will be open automatically


apps/android/app/src/main/java/chat/simplex/app/views/helpers/Modifiers.kt
```diff
@@ -15,3 +23,22 @@ fun Modifier.badgeLayout() =
...
+  if (DismissDirection.StartToEnd in directions) anchors += swipeDistance to DismissValue.DismissedToEnd
+  if (DismissDirection.EndToStart in directions) anchors += -swipeDistance to DismissValue.DismissedToStart
+  return Modifier.swipeable(
+    state = state,
+    anchors = anchors,
+    thresholds = { _, _ -> FractionalThreshold(0.5f) },
+    orientation = Orientation.Horizontal,
+    reverseDirection = isRtl,
+  ).offset { IntOffset(state.offset.value.roundToInt(), 0) }
+}
```
epoberezkin, line 44:
```suggestion
}

```

elendil7:
Very good feature

PR #942, opened 2022-08-17
Fixes service issues

avently:
- no more crashes after start of a service with battery optimization enabled
- no more service restarts after the app exit with disabled private notifications


apps/android/app/src/main/java/chat/simplex/app/SimplexApp.kt
```diff
@@ -51,16 +54,6 @@ class SimplexApp: Application(), LifecycleEventObserver {
...
-    withApi {
-      val user = chatController.apiGetActiveUser()
-      if (user == null) {
-        chatModel.onboardingStage.value = OnboardingStage.Step1_SimpleXInfo
-      } else {
-        chatController.startChat(user)
-        SimplexService.start(applicationContext)
-        chatController.showBackgroundServiceNoticeIfNeeded()
-      }
-    }
```
epoberezkin, line 63: where does it start now?

PR #946, opened 2022-08-18
core: commands to set chat notification settings

epoberezkin:


src/Simplex/Chat/Types.hs
```diff
@@ -200,6 +202,17 @@ instance ToJSON GroupInfo where toEncoding = J.genericToEncoding J.defaultOption
 groupName' :: GroupInfo -> GroupName
 groupName' GroupInfo {localDisplayName = g} = g
 
+-- TODO when more settings are added we should create another type to allow partial setting updates (with all Maybe properties)
+data ChatSettings = ChatSettings
```
jr-simplex, line 206: I'm not sure about using the same type for contacts and groups, might backfire

PR #977, opened 2022-08-27
core: use async agent commands when establishing connections w/t user action

jr-simplex:


docs/rfcs/2022-08-26-group-connections-recovery.md
```diff
@@ -0,0 +1,43 @@
...
+
+- Add new connection type `NewConnection` / `NoQueueConnection`;
+
+- Add agent responses signalling these commands' asynchronous completion, e.g. `CREATE_SUCCESS`, `JOIN_SUCCESS` (?);
+
+- Return connection status on subscriptions;
+
+In chat:
+
+- Event-driven processing - for group connections use commands with synchronous responses, save intermediate connection state, process success responses;
```
epoberezkin, line 19:
```suggestion
- Event-driven processing - for group connections use commands with asynchronous responses, save intermediate connection state, process success responses;
```

jr-simplex: meant immediate synchronous response, e.g. conn id, and the "success response" is asynchronous

src/Simplex/Chat.hs
```diff
@@ -1675,13 +1736,34 @@ processAgentMessage (Just user@User {userId, profile}) agentConnId agentMessage
...
+          Just cmdConnId' ->
+            if connId == cmdConnId'
+              then
+                if aCommandTag agentMsg == commandExpectedResponse cmdFunction
+                  then do
+                    withStore' $ \db -> updateCommandStatus db user cmdId CSCompleted
+                    action cmdData
+                  else throwChatError $ CECommandError "unexpected response for correlated command"
+              else throwChatError $ CECommandError "correlated command's connection id doesn't match"
+          Nothing -> throwChatError $ CECommandError "correlated command doesn't have connection id"
```
jr-simplex, line 1753: not sure about re-using this error

src/Simplex/Chat/Store.hs
```diff
@@ -3846,6 +3891,125 @@ getCalls db User {userId} = do
...
+  DB.execute
+    db
+    [sql|
+      UPDATE connections
+      SET conn_req_inv = ?, updated_at = ?
+      WHERE user_id = ? AND connection_id = ?
+    |]
+    (connReq, updatedAt, userId, connId)
+
+getXGrpMemIntroContDataDirect :: DB.Connection -> User -> Contact -> IO (Maybe XGrpMemIntroContData)
```
jr-simplex, line 3957: wanted name to reflect it's continuation for specific message, but we can change

src/Simplex/Chat.hs
```diff
@@ -1362,24 +1367,45 @@ processAgentMessage (Just user@User {userId, profile}) agentConnId agentMessage
...
-          ackMsgDeliveryEvent conn meta
+          cmdId <- createAckCmd conn
+          _ <- saveRcvMSG conn (ConnectionId connId) meta msgBody cmdId
+          withAckMessage agentConnId cmdId meta $ pure ()
         SENT msgId ->
           -- ? updateDirectChatItemStatus
           sentMsgDeliveryEvent conn msgId
+        OK ->
+          -- [async agent commands] continuation on receiving OK
+          withCompletedCommand conn agentMsg $ \CommandData {cmdFunction, cmdId} ->
```
jr-simplex, line 1383: `withCompletedCommand` can be on higher level

src/Simplex/Chat.hs
epoberezkin: I would just combine all 4 conditions in one and throw one error otherwise.

src/Simplex/Chat.hs
```diff
@@ -1675,13 +1736,34 @@ processAgentMessage (Just user@User {userId, profile}) agentConnId agentMessage
...
+          Nothing -> throwChatError $ CECommandError "correlated command doesn't have connection id"
+
+    createAckCmd :: Connection -> m CommandId
+    createAckCmd Connection {connId} = do
+      withStore' $ \db -> createCommand db user (Just connId) CFAckMessage
+
+    withAckMessage :: ConnId -> CommandId -> MsgMeta -> m () -> m ()
+    withAckMessage cId cmdId MsgMeta {recipient = (msgId, _)} action =
+      -- [async agent commands] command should be asynchronous, continuation is ackMsgDeliveryEvent
+      action `E.finally` withAgent (\a -> ackMessageAsync a (aCorrId cmdId) cId msgId `catchError` \_ -> pure ())
```
epoberezkin, line 1762: ackMessageAsync is non-trowing, right?

src/Simplex/Chat.hs
```diff
@@ -1887,10 +1969,13 @@ processAgentMessage (Just user@User {userId, profile}) agentConnId agentMessage
...
           then
-            tryError (withAgent $ \a -> joinConnection a True fileConnReq . directMessage $ XOk) >>= \case
-              Right acId ->
-                withStore' $ \db -> createSndGroupFileTransferConnection db userId fileId acId m
-              Left e -> throwError e
+            tryError
+              -- [async agent commands] no continuation needed, but command should be asynchronous for stability
+              (joinAgentConnectionAsync user True fileConnReq . directMessage $ XOk)
+              >>= \case
+                Right connIds ->
```
epoberezkin, line 1976:
```suggestion
                Right connIds ->
```

why plural?

src/Simplex/Chat/Types.hs
```diff
@@ -915,3 +915,78 @@ type JSONString = String
...
+    cmdFunction :: CommandFunction,
+    cmdStatus :: CommandStatus
+  }
+  deriving (Show)
+
+-- ad-hoc type for data required for XGrpMemIntro continuation
+data XGrpMemIntroContData = XGrpMemIntroContData
+  { groupId :: GroupId,
+    groupMemberId :: GroupMemberId,
+    memberId :: MemberId,
```
epoberezkin, line 988: these things should be already in the context, no?

src/Simplex/Chat.hs
```diff
@@ -1462,7 +1492,22 @@ processAgentMessage (Just user@User {userId, profile}) agentConnId agentMessage
...
+      INV (ACR _ cReq) ->
+        -- [async agent commands] XGrpMemIntro continuation on receiving INV
+        withCompletedCommand conn agentMsg $ \_ ->
+          case cReq of
+            groupConnReq@(CRInvitationUri _ _) -> do
+              contData <- withStore' $ \db -> do
+                setConnConnReqInv db user connId cReq
+                getXGrpMemIntroContDataGroup db user m
+              forM_ contData $ \XGrpMemIntroContData {groupMemberId, memberId, connReq, hostConnId} ->
+                forM_ connReq $ \directConnReq -> do
```
epoberezkin, line 1505: all three (or at least two) conditions can be one case and then throw error if either is Nothing or wrong type?

epoberezkin: or maybe it should be no action as we are waiting for the second one

src/Simplex/Chat.hs
```diff
@@ -1362,24 +1367,45 @@ processAgentMessage (Just user@User {userId, profile}) agentConnId agentMessage
...
+                contData <- withStore' $ \db -> do
+                  setConnConnReqInv db user connId cReq
+                  getXGrpMemIntroContDataDirect db user ct
+                forM_ contData $ \XGrpMemIntroContData {groupId, groupMemberId, memberId, connReq, hostConnId} ->
+                  forM_ connReq $ \groupConnReq -> do
+                    hostConn <- withStore $ \db -> getConnectionById db user hostConnId
+                    let msg = XGrpMemInv memberId IntroInvitation {groupConnReq, directConnReq}
+                    void $ sendDirectMessage hostConn msg (GroupId groupId)
+                    withStore' $ \db -> updateGroupMemberStatusById db userId groupMemberId GSMemIntroInvited
+              CRContactUri _ -> throwChatError $ CECommandError "unexpected ConnectionRequestUri type"
```
epoberezkin, line 1404: this looks almost like below

src/Simplex/Chat/Types.hs
epoberezkin: probably at least groupId?

src/Simplex/Chat.hs
jr-simplex: it's command's and connection's ids

src/Simplex/Chat/Types.hs
jr-simplex: not in direct connection

src/Simplex/Chat.hs
jr-simplex: yes, I started refactoring it initially but it either requires signalling which connection connection request refers to (direct or group) if you start from Maybe XGrpMemIntroContData, or only refactoring last 4 lines (from reading `hostConn` to status update). Starting even higher is even more hassle since store method is different. Overall I'm not sure it's worth it

src/Simplex/Chat.hs
jr-simplex: I think it is, it reads connection and can throw CONN and CMD errors

PR #1009, opened 2022-09-02
android: Search in a list of chats

avently:


apps/android/app/src/main/java/chat/simplex/app/views/chatlist/ChatListView.kt
```diff
@@ -103,9 +107,45 @@ fun ChatListView(chatModel: ChatModel, setPerformLA: (Boolean) -> Unit, stopped:
...
 
 @Composable
-fun ChatListToolbar(chatModel: ChatModel, scaffoldCtrl: ScaffoldController, stopped: Boolean) {
+fun ChatListToolbar(chatModel: ChatModel, scaffoldCtrl: ScaffoldController, stopped: Boolean, onSearchValueChanged: (String) -> Unit) {
+  var showSearch by rememberSaveable { mutableStateOf(false) }
+  val hideSearchOnBack = { onSearchValueChanged(""); showSearch = false }
+  if (showSearch) {
+    BackHandler(onBack = hideSearchOnBack)
+  }
+  val barButtons = arrayListOf<@Composable RowScope.() -> Unit>()
```
epoberezkin, line 116: this is not a necessary abstraction, please remove it - there should be just a condition in the view.

PR #1012, opened 2022-09-02
ios: use SQLCipher

epoberezkin:


apps/ios/Shared/Views/Database/DatabaseEncryptionView.swift
```diff
@@ -0,0 +1,286 @@
...
+                primaryButton: .destructive(Text("Remove")) {
+                    if removeDatabaseKey() {
+                        storeDBPassphraseGroupDefault.set(false)
+                        storedKey = false
+                    } else {
+                        alert = .error(title: "Keychain error", error: "Failed to remove passphrase")
+                    }
+                },
+                secondaryButton: .cancel() {
+                    useKeychain = true
```
jr-simplex, line 185:
```suggestion
                    withAnimation { useKeychain = true }
```

apps/ios/Shared/Views/Database/DatabaseEncryptionView.swift
```diff
@@ -0,0 +1,286 @@
...
+                    !validKey(currentKey) ||
+                    !validKey(newKey)
+                )
+            } header: {
+                Text("")
+            } footer: {
+                VStack(alignment: .leading, spacing: 16) {
+                    if m.chatDbEncrypted == false {
+                        Text("Your chat database is not encrypted - set passphrase to encrypt it.")
+                    } else if useKeychain {
```
jr-simplex, line 101: use separate variable for determining text to show, so that when disabling keychain text doesn't change behind alert before confirmation? (the variable to be changed on confirm)

jr-simplex (approved):

PR #1053, opened 2022-09-14
android: Changed requirements for ignoring battery for Periodic notifications mode

avently:


apps/android/app/src/main/java/chat/simplex/app/SimplexApp.kt
```diff
@@ -86,7 +86,16 @@ class SimplexApp: Application(), LifecycleEventObserver {
...
+        Lifecycle.Event.ON_RESUME -> CoroutineScope(Dispatchers.Default).launch {
+          /**
+           * When the app starts [chatModel.onboardingStage] is null, because [Lifecycle.Event.ON_RESUME] called sooner than chat is
+           * able to start. This is why [withTimeoutOrNull] waits for the moment when the value is not null and runs required steps afterward
+           * */
+          withTimeoutOrNull(10_000) {
+            while (isActive && chatModel.onboardingStage.value == null) {
+              delay(1000)
+            }
+          }
```
epoberezkin, line 98: this seems a hack - isn't there a better way? like some event sent from the place where chatModel.onboardingStage.value is set?

apps/android/app/src/main/java/chat/simplex/app/views/usersettings/NotificationsSettingsView.kt
```diff
@@ -24,8 +25,11 @@ import chat.simplex.app.views.helpers.*
 import kotlinx.coroutines.*
 import kotlin.collections.ArrayList
 
-enum class NotificationsMode(val requiresIgnoringBattery: Boolean) {
-  OFF(false), PERIODIC(false), SERVICE(true), /*INSTANT(false) - for Firebase notifications */;
+enum class NotificationsMode(private val requiresIgnoringBatterySinceSdk: Int) {
+  OFF(Int.MAX_VALUE), PERIODIC(Build.VERSION_CODES.M), SERVICE(Build.VERSION_CODES.S), /*INSTANT(Int.MAX_VALUE) - for Firebase notifications */;
```
epoberezkin, line 29: I thought this value would somehow affect how this is serialised, but it seems like it's not the case?

PR #1070, opened 2022-09-19
android: Audio messages

avently:
This is just for testing currently


apps/android/app/src/main/java/chat/simplex/app/views/chat/item/CIFileView.kt
```diff
@@ -71,6 +104,17 @@ fun CIFileView(
 
   fun fileAction() {
     if (file != null) {
+      if (file.isVoiceMessage() && file.loaded) {
```
jr-simplex, line 107: I think voice message chat item logic should live in a separate file, similar to CIImageView - CIVoiceMsgView.

Also I don't think that whether it's a voice message should be determined by checking extensions as `isVoiceMessage`does - I believe we don't do this for images.. The branching will be possible based on MsgContent type - and if the format is unplayable (the check for extension can be more downstream) chat item can indicate it.

apps/android/app/src/main/java/chat/simplex/app/views/helpers/Util.kt
```diff
@@ -220,6 +222,11 @@ private fun spannableStringToAnnotatedString(
 // maximum image file size to be auto-accepted
 const val MAX_IMAGE_SIZE: Long = 236700
 const val MAX_IMAGE_SIZE_AUTO_RCV: Long = MAX_IMAGE_SIZE * 2
+const val MAX_VOICE_SIZE_AUTO_RCV: Long = MAX_IMAGE_SIZE_AUTO_RCV
```
jr-simplex, line 225: maybe it should be equal to `MAX_IMAGE_SIZE`?

apps/android/app/src/main/java/chat/simplex/app/views/helpers/Util.kt
```diff
@@ -449,3 +456,9 @@ fun Color.darker(factor: Float = 0.1f): Color =
 fun ByteArray.toBase64String() = Base64.encodeToString(this, Base64.DEFAULT)
 
 fun String.toByteArrayFromBase64() = Base64.decode(this, Base64.DEFAULT)
+
+val LongRange.Companion.saver
+  get() = Saver<MutableState<LongRange>, Pair<Long, Long>>(
+    save = { it.value.first to it.value.last },
+    restore = { mutableStateOf(it.first..it.second) }
+    )
```
jr-simplex, line 464:
```suggestion
    )

```

apps/android/app/src/main/res/values-de/strings.xml
```diff
@@ -225,6 +229,8 @@
 
   <!-- Message Actions - SendMsgView.kt -->
   <string name="icon_descr_send_message">Nachricht senden</string>
+  <string name="icon_descr_record_audio">***Record audio message</string>
+  <string name="tap_and_hold_to_record">***Tap and hold the button, then release to stop the recording</string>
```
jr-simplex, line 233:
```suggestion
  <string name="icon_descr_record_audio">***Record voice message</string>
  <string name="tap_and_hold_to_record">***Tap and hold the button, then release to finish recording</string>
```

apps/android/app/src/main/res/values/strings.xml
```diff
@@ -225,6 +229,8 @@
 
   <!-- Message Actions - SendMsgView.kt -->
   <string name="icon_descr_send_message">Send Message</string>
+  <string name="icon_descr_record_audio">Record audio message</string>
+  <string name="tap_and_hold_to_record">Tap and hold the button, then release to stop the recording</string>
```
jr-simplex, line 233:
```suggestion
  <string name="icon_descr_record_voice_message">Record voice message</string>
```

apps/android/app/src/main/res/values-ru/strings.xml
```diff
@@ -225,6 +229,8 @@
 
     <!-- Message Actions - SendMsgView.kt -->
     <string name="icon_descr_send_message">Отправить сообщение</string>
+    <string name="icon_descr_record_audio">Записать аудио сообщение</string>
+    <string name="tap_and_hold_to_record">Нажмите и удерживайте для записи, а после - отпустите</string>
```
jr-simplex, line 233:
```suggestion
    <string name="icon_descr_record_voice_message">Записать голосовое сообщение</string>
```

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
```diff
@@ -150,10 +86,216 @@ fun SendMsgView(
...
+                stopRecOnNextClick = true
+              }
+            },
+            onCancel = startStopRecording,
+            onRelease = startStopRecording
+          )
+          Spacer(Modifier.weight(1f))
+          Icon(
+            if (recordingTimeRange.last != 0L) Icons.Outlined.ArrowUpward else if (stopRecOnNextClick) Icons.Default.Stop else Icons.Default.Mic,
+            stringResource(R.string.icon_descr_record_audio),
```
jr-simplex, line 180:
```suggestion
            stringResource(R.string.icon_descr_record_voice_message),
```

apps/android/app/src/main/java/chat/simplex/app/views/chat/ComposeView.kt
```diff
@@ -118,11 +125,12 @@ fun chatItemPreview(chatItem: ChatItem): ComposePreview {
     is MsgContent.MCText -> ComposePreview.NoPreview
     is MsgContent.MCLink -> ComposePreview.CLinkPreview(linkPreview = mc.preview)
     is MsgContent.MCImage -> ComposePreview.ImagePreview(images = listOf(mc.image))
+    is MsgContent.MCVoice -> ComposePreview.VoicePreview(voice = chatItem.file?.fileName ?: "", mc.duration / 1000, true)
```
jr-simplex, line 128: I may understand it wrong, but this seems to imply that it's sent in milliseconds - duration should instead be passed in seconds on send

apps/android/app/src/main/java/chat/simplex/app/views/chat/ComposeView.kt
```diff
@@ -497,7 +538,8 @@ fun ComposeView(
       verticalAlignment = Alignment.Bottom,
       horizontalArrangement = Arrangement.spacedBy(2.dp)
     ) {
-      val attachEnabled = !composeState.value.editing
+      var showRecordingUi by rememberSaveable { mutableStateOf(false) }
```
jr-simplex, line 541:
```suggestion
      var recordingVoiceMsg by rememberSaveable { mutableStateOf(false) }
```

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
```diff
@@ -12,127 +15,60 @@ import androidx.compose.foundation.layout.*
...
+import java.io.*
 
 @Composable
 fun SendMsgView(
   composeState: MutableState<ComposeState>,
+  allowVoiceRecord: Boolean,
   sendMessage: () -> Unit,
   onMessageChange: (String) -> Unit,
+  onAudioAdded: (String, Int, Boolean) -> Unit,
+  showRecordingUi: (Boolean) -> Unit,
```
jr-simplex, line 55:
```suggestion
  recordingVoiceMsg: (Boolean) -> Unit,
```

apps/android/app/src/main/java/chat/simplex/app/views/chat/item/CIVoiceView.kt
```diff
@@ -0,0 +1,344 @@
...
+        "                     "
+      else
+        "                 "
+      Text(metaReserve)
+    }
+  }
+}
+
+@Composable
+private fun fileIcon(
```
jr-simplex, line 142: linter complains here "Composable functions that return Unit should start with an uppercase letter"
```suggestion
private fun FileIcon(
```

apps/android/app/src/main/java/chat/simplex/app/views/chat/item/CIVoiceView.kt
```diff
@@ -0,0 +1,344 @@
...
+          Toast.makeText(SimplexApp.context, generalGetString(R.string.file_not_found), Toast.LENGTH_SHORT).show()
+        }
+      }
+      else -> {}
+    }
+  }
+}
+
+@Composable
+private fun fileIndicator(
```
jr-simplex, line 204: same
```suggestion
private fun FileIndicator(
```

apps/android/app/src/main/java/chat/simplex/app/views/chat/item/CIVoiceView.kt
```diff
@@ -0,0 +1,344 @@
...
+      startAngle = -90f,
+      sweepAngle = angle,
+      useCenter = false,
+      style = Stroke(width = strokeWidth, cap = StrokeCap.Round)
+    )
+  }
+}
+
+@Composable
+private fun progressIndicator() {
```
jr-simplex, line 299: same
```suggestion
private fun ProgressIndicator() {
```

apps/android/app/src/main/java/chat/simplex/app/views/chat/item/CIVoiceView.kt
```diff
@@ -0,0 +1,344 @@
...
+import chat.simplex.app.R
+import chat.simplex.app.SimplexApp
+import chat.simplex.app.model.*
+import chat.simplex.app.ui.theme.*
+import chat.simplex.app.views.helpers.*
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.isActive
+
+@Composable
+fun CIVoiceView(
```
jr-simplex, line 37: please add some previews for this - sent/received without/with text (for example one has to modify code to reproduce item with text currently)

apps/android/app/src/main/java/chat/simplex/app/views/chat/item/CIVoiceView.kt
```diff
@@ -0,0 +1,344 @@
...
+
+          CIFileStatus.RcvAccepted -> fileIcon(innerIcon = Icons.Outlined.MoreHoriz)
+          CIFileStatus.RcvTransfer -> progressIndicator()
+          CIFileStatus.RcvComplete -> fileIcon()
+          CIFileStatus.RcvCancelled -> fileIcon(innerIcon = Icons.Outlined.Close)
+        }
+      } else {
+        fileIcon()
+      }
+    }
```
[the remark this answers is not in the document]
jr-simplex: `fileIcon`, `progressIndicator` and `fileSizeValid` functions are redundant as well,

apps/android/app/src/main/java/chat/simplex/app/views/helpers/RecAndPlay.kt
```diff
@@ -0,0 +1,168 @@
...
+    // Notify prev audio listener about stop
+    currentlyPlaying.value?.second?.invoke()
+    currentlyPlaying.value = null
+    player.stop()
+  }
+
+  /**
+   * If player starts playing at 2637 ms in a track 2816 ms long, it will stop immediately after start but will not
+   *  change currentPosition, so it will not be equal to duration. However, it sets isPlaying to false.
+   *  Let's do it ourselves in order to prevent endless waiting loop
```
jr-simplex, line 151: is 2637 ms / 2816 ms an example or specific numbers?.. I'm trying to understand where it's implied in code/what fixes it

apps/android/app/src/main/java/chat/simplex/app/views/chat/item/CIVoiceView.kt
```diff
@@ -0,0 +1,344 @@
...
+        Icon(
+          imageVector = if (audioPlaying) Icons.Filled.Pause else Icons.Filled.PlayArrow,
+          contentDescription = null,
+          Modifier.size(36.dp),
+          tint = MaterialTheme.colors.primary
+        )
+      }
+    } else {
+      Box(drawRingModifier(angle.value, primary, strokeWidth)) {
+        FloatingActionButton(
```
jr-simplex, line 229: I don't think it's a good idea to use `FloatingActionButton` for this - it has many things we don't need/use, and semantically it's not a FAB, just a message icon. Reading FAB's code it seems what we need from it is mostly that it's `Surface` with `MaterialTheme.shapes.small.copy(CornerSize(percent = 50))` Shape, maybe just use `Surface` and add what's necessary?

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
```diff
@@ -150,10 +86,216 @@ fun SendMsgView(
...
+              } else {
+                // tapped and didn't hold a finger
+                stopRecOnNextClick = true
+              }
+            },
+            onCancel = startStopRecording,
+            onRelease = startStopRecording
+          )
+          Spacer(Modifier.weight(1f))
+          Icon(
```
jr-simplex, line 178: It seems it misses `.clip(CircleShape)` and the color should depend as well then for when it's "ready for sent" (it should be white arrow upward in accent circle).

Also these set of buttons also should take into account `cs.sendEnabled()` - both record and send, and the background color if send is disabled is HighOrLowlight instead of primary - see `if (!showVoiceButton)`

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
```diff
@@ -150,10 +86,216 @@ fun SendMsgView(
           )
         }
       }
+      if (showVoiceButton) {
```
jr-simplex, line 89: why is it not in `if (!showVoiceButton)` else branch, and instead on a higher level?

jr-simplex:
There's some bug that makes a keyboard very laggy/almost impossible to close or interact with when voice message is cancelled

apps/android/app/src/main/java/chat/simplex/app/views/helpers/Util.kt
avently: @epoberezkin decides so

apps/android/app/src/main/java/chat/simplex/app/views/chat/ChatView.kt
```diff
@@ -94,7 +94,6 @@ fun ChatView(chatId: String, chatModel: ChatModel, onComposed: () -> Unit) {
     chatModel.chatId.value = null
   } else {
     val chat = activeChat.value!!
-    BackHandler { chatModel.chatId.value = null }
```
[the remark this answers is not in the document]
avently: It's not needed now:
https://github.com/simplex-chat/simplex-chat/blob/b79773f26a1635c14868f2212d0dd67c57e84fcb/apps/android/app/src/main/java/chat/simplex/app/views/chat/ChatView.kt#L123
https://github.com/simplex-chat/simplex-chat/blob/b79773f26a1635c14868f2212d0dd67c57e84fcb/apps/android/app/src/main/java/chat/simplex/app/views/chat/ChatView.kt#L323

apps/android/app/src/main/java/chat/simplex/app/views/chat/item/CIVoiceView.kt
avently: I don't have a preview working at all in IDEA, I don't see any preview in the app. So, no, I will not create any new of them

apps/android/app/src/main/java/chat/simplex/app/views/helpers/RecAndPlay.kt
avently: This is just an example that I saw. Modified a comment to highlight it

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
avently: Previously it was better to do like this but with a new behaviour and UI, it can be placed in if statement, yes.

apps/android/app/src/main/java/chat/simplex/app/views/chat/ComposeVoiceView.kt
```diff
@@ -0,0 +1,124 @@
...
+import androidx.compose.ui.unit.sp
+import chat.simplex.app.R
+import chat.simplex.app.ui.theme.*
+import chat.simplex.app.views.chat.item.MiniAudioPlayer
+import chat.simplex.app.views.chat.item.SentColorLight
+import chat.simplex.app.views.helpers.*
+import kotlinx.coroutines.flow.distinctUntilChanged
+
+@Composable
+fun ComposeVoiceView(filePath: String, durationMs: Int, finished: Boolean, cancelEnabled: Boolean, cancelFile: () -> Unit) {
```
jr-simplex, line 24: or maybe `cancelRecording`?
```suggestion
fun ComposeVoiceView(filePath: String, durationMs: Int, finished: Boolean, cancelEnabled: Boolean, cancelVoiceMessage: () -> Unit) {
```

apps/android/app/src/main/java/chat/simplex/app/views/helpers/Util.kt
jr-simplex: "feel bad" is a strong wording, no, I just saw git diff highlighting

apps/android/app/src/main/java/chat/simplex/app/views/chat/ComposeVoiceView.kt
```diff
@@ -0,0 +1,124 @@
...
+      if (cancelEnabled) {
+        IconButton(
+          onClick = {
+            AudioPlayer.stop(filePath)
+            cancelFile()
+          },
+          modifier = Modifier.padding(0.dp)
+        ) {
+          Icon(
+            Icons.Outlined.Close,
```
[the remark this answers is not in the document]
avently: It was fixed in https://github.com/simplex-chat/simplex-chat/pull/1070/commits/52ac9d84c907910a1eabe675adc9d42a33ba84c5

apps/android/app/src/main/java/chat/simplex/app/views/chat/item/CIVoiceView.kt
avently: What makes you thinking that we need to, basically, recreate a FAB code ourselves? FAB is just surface, yes, so you suggest to use Surface... with all setup that FAB already do

avently: What we will win here if we choose Surface?

PR #1077, opened 2022-09-20
webrtc: change server ports to 443, ios: allow configuring WebRTC ICE servers

epoberezkin:


apps/ios/Shared/Views/UserSettings/SettingsView.swift
```diff
@@ -193,7 +194,7 @@ struct SettingsView: View {
                     } label: {
                         settingsRow("terminal") { Text("Chat console") }
                     }
-                    settingsRow("gear") {
+                    settingsRow("chevron.left.forwardslash.chevron.right") {
```
jr-simplex, line 197: why change?

epoberezkin: never liked this gear

PR #1090, opened 2022-09-22
android: Allow configuring WebRTC ICE servers

avently:


apps/android/app/src/main/java/chat/simplex/app/views/call/WebRTC.kt
```diff
@@ -153,4 +155,47 @@ class ConnectionState(
...
+    }
+  }
+  return if (iceServers.isEmpty()) null else iceServers
+}
+
+fun getIceServers(): List<RTCIceServer>? {
+  val value = SimplexApp.context.chatController.appPrefs.webrtcIceServers.get() ?: return null
+  val servers: List<String> = value.split("\n")
+  return parseRTCIceServers(servers)
+}
```
epoberezkin, line 201:
```suggestion
}

```

PR #1096, opened 2022-09-22
rfc: chat settings

epoberezkin:


docs/rfcs/2022-09-22-chat-settings.md
```diff
@@ -0,0 +1,71 @@
...
+  "additionalProperties": true
+}
+```
+
+Every time user updates the settings and update profile should be sent to affected contacts.
+
+Database schema:
+
+```sql
+ALTER TABLE users ADD COLUMN preferences TEXT DEFAULT '{}' CHECK (preferences NOT NULL);
```
jr-simplex, line 62: considering `preferences` in `users` table is user's global setting, when they're updated is update sent to contacts who the user provided per chat preferences? if user only configured some preference for contact are other globals sent on update? or are global preferences only sent to new contacts (but then the update is counterintuitive)? I see that global preferences are not in scope of this rfc but then I'm not sure what this field is for

jr-simplex: also, are groups where user is owner updated?

jr-simplex (approved):

Issue #1032, opened 2022-09-08
[ANDROID] Arabic text displayed over timestamp.


Icy-Thought:
As the title mentions, chat messages sent in Arabic are displayed in a rather weird way when compared to other chat applications.

Two instances I had noticed during my short SimpleX interaction was:

1. Single line message would wind up placed over the message timestamp: ![20220907_205750](https://user-images.githubusercontent.com/53710398/189005882-54630d4f-7ee8-4a45-9938-d7302e8c981b.jpg)

2. Multi-line message would result in a weird text alignment: ![20220907_205715](https://user-images.githubusercontent.com/53710398/189005908-dcd3c081-decc-4ce3-8846-05d09819fca9.jpg)

epoberezkin:
Thank you for submitting it. I am aware of that problem, and that's not obvious how to fix it... We will prioritise it once database encryption is done.

epoberezkin:
From experimenting with different messages, on iOS right-to-left languages seem to work this way:
1. the character order in right-to-left language is always right-to-left
2. the word order seems determined by the first word - if it's in right-to-left language, then the word order is also right-to-left, if it's left-to-rifht language, then the word order is left-to-right, even if all other words are right-to-left.

Currently, we always align messages on the left side, whether single- or multi-line, and the timestamp - on the right.

For messages that are right-to-left (that is, the first work is in right-to-left language), we should align messages on the right, and the timestamp on the left.

Interestingly, app UI is not getting flipped when the whole phone UI is getting flipped, but it's something else to look at...

avently:
@Icy-Thought, I made some changes that fix issues you mentioned. Let me know if it's not working as expected still

epoberezkin:
Should be fixed in 4.0.1

Icy-Thought:
As of the latest availabe version (v4.0.0), it appears as if the issue has not been resolved yet.

Icy-Thought:
> Should be fixed in 4.0.1

Will report back whenever I have that version installed. Appreciate the fix!

avently:
@Icy-Thought if you want to test the changes right now, you can install this apk (it will be installed side-by-side with your release apk without replacing it, so you can setup a new profile there just for testing).

Gihub has 25 MB file size limit so I uploaded it into different site that allows 50+ MB size.
[...]

PR #1117, opened 2022-09-24
ios: fix layout for messages in right-to-left languages, #1032

epoberezkin:


jr-simplex:
I believe right alignment is also required in compose view quote

PR #1138, opened 2022-09-27
android: Playing call sound in different stream type

avently:


apps/android/app/src/main/java/chat/simplex/app/views/call/SoundPlayer.kt
```diff
@@ -1,27 +1,36 @@
...
 import androidx.core.content.ContextCompat
 import chat.simplex.app.R
+import chat.simplex.app.SimplexApp
 import chat.simplex.app.views.helpers.withScope
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.delay
 
 class SoundPlayer {
-  var player: MediaPlayer? = null
+  private var player: MediaPlayer = MediaPlayer().apply {
```
epoberezkin, line 16: should we at least make it lazy, so it's not initialised until needed?

PR #1158, opened 2022-10-01
mobile: disable notifications on group events, make member role default

epoberezkin:


apps/ios/SimpleXChat/ChatTypes.swift
```diff
@@ -892,6 +892,19 @@ public struct ChatItem: Identifiable, Decodable {
         }
     }
 
+    public var isMutedMemberEvent: Bool {
```
jr-simplex, line 895: add `memberAdded` and `memberDeleted`?

jr-simplex (approved):

PR #1159, opened 2022-10-01
core: protocol/commands to change member role

epoberezkin:


docs/protocol/simplex-chat.md
```diff
@@ -213,6 +213,8 @@ Currently members can have one of three roles - `owner`, `admin` and `member`. T
 
 `x.grp.mem.info` this message is sent as part of member connection handshake - it includes group member profile.
 
+`x.grp.mem.role` message is sent to update group member role - it is sent to all members by the member who updated the role of the member referenced in this message. This message MUST only be sent by members with `admin` or `owner` role. Receiving clients MUST ignore this message if it is received from member with `member` role.
```
jr-simplex, line 216: Since there's a new "author" role
```suggestion
`x.grp.mem.role` message is sent to update group member role - it is sent to all members by the member who updated the role of the member referenced in this message. This message MUST only be sent by members with `admin` or `owner` role. Receiving clients MUST ignore this message if it is received from member with role less than `admin`.
```

src/Simplex/Chat.hs
```diff
@@ -835,7 +835,26 @@ processChatCommand = \case
...
+        Just m -> changeMemberRole user gInfo members m $ SGEMemberRole memberId memRole
+    where
+      changeMemberRole user gInfo@GroupInfo {membership} members m@GroupMember {memberId = mId, memberRole = mRole, memberStatus = mStatus} gEvent = do
+        let userRole = memberRole (membership :: GroupMember)
+            canChangeRole = userRole >= GRAdmin && userRole >= mRole && userRole >= memRole && memberCurrent membership
+        unless canChangeRole $ throwChatError CEGroupUserRole
+        withChatLock . procCmd $ do
+          unless (mRole == memRole) $ do
+            withStore' $ \db -> updateGroupMemberRole db user m memRole
+            unless (mStatus == GSMemInvited) $ do
```
jr-simplex, line 853: How does invited member learn his role has changed? Maybe it should be prohibited - a new invitation has to be sent, maybe even as part of this command? I think repeat invitations may already treat this case, better check though

src/Simplex/Chat.hs
```diff
@@ -835,7 +835,26 @@ processChatCommand = \case
...
+    Group gInfo@GroupInfo {membership} members <- withStore $ \db -> getGroup db user groupId
+    if memberId == groupMemberId' membership
+      then changeMemberRole user gInfo members membership $ SGEUserRole memRole
+      else case find ((== memberId) . groupMemberId') members of
+        Nothing -> throwChatError CEGroupMemberNotFound
+        Just m -> changeMemberRole user gInfo members m $ SGEMemberRole memberId memRole
+    where
+      changeMemberRole user gInfo@GroupInfo {membership} members m@GroupMember {memberId = mId, memberRole = mRole, memberStatus = mStatus} gEvent = do
+        let userRole = memberRole (membership :: GroupMember)
+            canChangeRole = userRole >= GRAdmin && userRole >= mRole && userRole >= memRole && memberCurrent membership
```
jr-simplex, line 848: I'm not sure it's correct we allow for role to govern same level role.. For remove as well - e.g. for owners to remove/change other owners role. As a bonus it would allow less role changes - less possibilities for fragmentation at the current stage

src/Simplex/Chat/Messages.hs
```diff
@@ -505,13 +505,17 @@ rcvGroupEventToText = \case
   RGEMemberAdded _ p -> "added " <> profileToText p
   RGEMemberConnected -> "connected"
   RGEMemberLeft -> "left"
+  RGEMemberRole _ r -> "role " <> safeDecodeUtf8 (strEncode r)
```
jr-simplex, line 508: Include member in text representation, e.g. using `profileToText` - otherwise it's unclear whose role changed

src/Simplex/Chat/Messages.hs
```diff
@@ -505,13 +505,17 @@ rcvGroupEventToText = \case
...
+  RGEUserRole r -> "your role " <> safeDecodeUtf8 (strEncode r)
   RGEMemberDeleted _ p -> "removed " <> profileToText p
   RGEUserDeleted -> "removed you"
   RGEGroupDeleted -> "deleted group"
   RGEGroupUpdated _ -> "group profile updated"
 
 sndGroupEventToText :: SndGroupEvent -> Text
 sndGroupEventToText = \case
+  SGEMemberRole _ r -> "role " <> safeDecodeUtf8 (strEncode r)
```
jr-simplex, line 517: Same

src/Simplex/Chat/Messages.hs
```diff
@@ -544,6 +548,8 @@ data RcvGroupEvent
   = RGEMemberAdded {groupMemberId :: GroupMemberId, profile :: Profile} -- CRJoinedGroupMemberConnecting
   | RGEMemberConnected -- CRUserJoinedGroup, CRJoinedGroupMember, CRConnectedToGroupMember
   | RGEMemberLeft -- CRLeftMember
+  | RGEMemberRole {groupMemberId :: GroupMemberId, role :: GroupMemberRole}
```
jr-simplex, line 551: Above comments would probably require profile in this and SGEMemberRole. I think it'd be necessary in UIs as well

src/Simplex/Chat/View.hs
```diff
@@ -478,6 +480,27 @@ connectedMember m = case memberCategory m of
   GCPostMember -> "new member " <> ttyMember m -- without fullName as as it was shown in joinedGroupMemberConnecting
   _ -> "member " <> ttyMember m -- these case is not used
 
+viewMemberRoleChanged :: GroupInfo -> GroupMember -> GroupMember -> GroupMemberRole -> GroupMemberRole -> [StyledString]
+viewMemberRoleChanged g@GroupInfo {membership} by m r r'
+  | r == r' = [ttyGroup' g <> ": member role did not change"]
+  | groupMemberId' membership == memId = view "your role"
+  | groupMemberId' by == memId = view "the role"
```
jr-simplex, line 487:
```suggestion
  | groupMemberId' by == memId = view "their role"
```

src/Simplex/Chat.hs
epoberezkin: Thought about it too - I think it's ok for admins to be able to downgrade other admins...

src/Simplex/Chat/View.hs
epoberezkin: "the" is ok

src/Simplex/Chat.hs
epoberezkin: updated

PR #1168, opened 2022-10-03
core: optimize bulk chat item deletion

jr-simplex:


src/Simplex/Chat.hs
```diff
@@ -499,22 +497,26 @@ processChatCommand = \case
...
-      ciIdsAndFileInfo <- withStore' $ \db -> getContactChatItemIdsAndFileInfo db user chatId
-      forM_ ciIdsAndFileInfo $ \(itemId, _, fileInfo_) -> deleteDirectChatItem user ct (itemId, fileInfo_)
-      ct' <- case ciIdsAndFileInfo of
-        [] -> pure ct
-        _ -> do
-          let (_, lastItemTs, _) = last ciIdsAndFileInfo
-          withStore' $ \db -> updateContactTs db user ct lastItemTs
-          pure (ct :: Contact) {updatedAt = lastItemTs}
+      filesInfo <- withStore' $ \db -> getContactFileInfo db user ct
+      maxItemTs_ <- withStore' $ \db -> getContactMaxItemTs db user ct
```
epoberezkin, line 501: This might be better to choose ID for contacts - for example the one that all chat items before this ID are older than specified. This way we will be deleting in the order of the chat, but some older chat items may remain.

src/Simplex/Chat.hs
```diff
@@ -499,22 +497,26 @@ processChatCommand = \case
...
       pure $ CRChatCleared (AChatInfo SCTDirect (DirectChat ct'))
     CTGroup -> do
       gInfo <- withStore $ \db -> getGroupInfo db user chatId
-      lastItemTs_ <- clearGroupContent user gInfo
-      gInfo' <- case lastItemTs_ of
-        Just lastItemTs -> do
-          withStore' $ \db -> updateGroupTs db user gInfo lastItemTs
-          pure (gInfo :: GroupInfo) {updatedAt = lastItemTs}
+      filesInfo <- withStore' $ \db -> getGroupFileInfo db user gInfo
+      maxItemTs_ <- withStore' $ \db -> getGroupMaxItemTs db user gInfo
```
epoberezkin, line 513: For groups it probably should be a pair of timestamp and chat item ID.

src/Simplex/Chat.hs
```diff
@@ -1128,41 +1124,26 @@ setExpireCIs b = do
...
-cancelFile user CIFileInfo {fileId, fileStatus = (AFS dir status)} =
-  unless (ciFileEnded status) $
-    case dir of
-      SMDSnd -> do
-        (ftm@FileTransferMeta {cancelled}, fts) <- withStore (\db -> getSndFileTransfer db user fileId)
-        unless cancelled $ cancelSndFile user ftm fts
-      SMDRcv -> do
-        ft@RcvFileTransfer {cancelled} <- withStore (\db -> getRcvFileTransfer db user fileId)
-        unless cancelled $ cancelRcvFileTransfer user ft
+deleteFile :: forall m. ChatMonad m => User -> CIFileInfo -> m ()
```
epoberezkin, line 1127: Does it still delete the connection somewhere?

PR #1172, opened 2022-10-04
core: optimize bulk chat item deletion 2

jr-simplex:


src/Simplex/Chat/Store.hs
```diff
@@ -4107,88 +4094,62 @@ setChatItemTTL db User {userId} chatItemTTL = do
...
+      (userId, groupId, expirationDate, createdAtCutoff)
 
-toItemIdAndFileInfo' :: (ChatItemId, Maybe Int64, Maybe ACIFileStatus, Maybe FilePath) -> (ChatItemId, Maybe CIFileInfo)
-toItemIdAndFileInfo' (chatItemId, fileId_, fileStatus_, filePath) =
-  case (fileId_, fileStatus_) of
-    (Just fileId, Just fileStatus) -> (chatItemId, Just CIFileInfo {fileId, fileStatus, filePath})
-    _ -> (chatItemId, Nothing)
+deleteGroupExpiredCIs :: DB.Connection -> User -> GroupInfo -> UTCTime -> UTCTime -> IO ()
+deleteGroupExpiredCIs db User {userId} GroupInfo {groupId} expirationDate createdAtCutoff = do
+  DB.execute db "DELETE FROM messages WHERE group_id = ? AND created_at <= ?" (groupId, expirationDate)
```
jr-simplex, line 4146: maybe it's better to add this filter here as well..
```suggestion
  DB.execute db "DELETE FROM messages WHERE group_id = ? AND created_at <= ? AND created_at <= ?" (groupId, expirationDate, createdAtCutoff)
```

src/Simplex/Chat.hs
```diff
@@ -449,7 +449,7 @@ processChatCommand = \case
       deleteCIFile :: MsgDirectionI d => User -> Maybe (CIFile d) -> m ()
       deleteCIFile user file =
         forM_ file $ \CIFile {fileId, filePath, fileStatus} -> do
-          let fileInfo = CIFileInfo {fileId, fileStatus = AFS msgDirection fileStatus, filePath}
+          let fileInfo = CIFileInfo {fileId, fileStatus = Just $ AFS msgDirection fileStatus, filePath}
```
epoberezkin, line 452: I think it can be better to supply some default when reading the database than to change the type - it's for old files only, but it complicates the code everywhere.

src/Simplex/Chat.hs
```diff
@@ -1401,18 +1402,45 @@ subscribeUserConnections agentBatchSubscribe user = do
...
-      loop filesInfo expirationDate expire
+    chatsLoop :: [ChatRef] -> UTCTime -> UTCTime -> TVar Bool -> m ()
+    chatsLoop [] _ _ _ = pure ()
+    chatsLoop ((ChatRef cType chatId) : chats) expirationDate createdAtCutoff expire = continue expire $ do
+      case cType of
+        CTDirect ->
+          withStore' (\db -> runExceptT $ getContact db userId chatId) >>= \case
+            Right ct -> do
+              filesInfo <- withStore' $ \db -> getContactExpiredFileInfo db user ct expirationDate
+              maxItemTs_ <- withStore' $ \db -> getContactMaxItemTs db user ct
```
epoberezkin, line 1422: do they need to be in separate transactions?

src/Simplex/Chat.hs
```diff
@@ -1401,18 +1402,45 @@ subscribeUserConnections agentBatchSubscribe user = do
...
+              forM_ filesInfo $ \fileInfo -> deleteFile user fileInfo
+              withStore' $ \db -> deleteContactExpiredCIs db user ct expirationDate
+              ciCount_ <- withStore' $ \db -> getContactCICount db user ct
```
epoberezkin, line 1425: same here

src/Simplex/Chat.hs
```diff
@@ -1401,18 +1402,45 @@ subscribeUserConnections agentBatchSubscribe user = do
...
+              case (maxItemTs_, ciCount_) of
+                (Just ts, Just count) -> when (count == 0) $ withStore' (\db -> updateContactTs db user ct ts)
+                _ -> pure ()
+            _ -> pure ()
+        CTGroup ->
+          withStore' (\db -> runExceptT $ getGroupInfo db user chatId) >>= \case
+            Right gInfo -> do
+              filesInfo <- withStore' $ \db -> getGroupExpiredFileInfo db user gInfo expirationDate createdAtCutoff
+              maxItemTs_ <- withStore' $ \db -> getGroupMaxItemTs db user gInfo
```
epoberezkin, line 1434: 1 transaction?

src/Simplex/Chat.hs
```diff
@@ -1401,18 +1402,45 @@ subscribeUserConnections agentBatchSubscribe user = do
...
+              forM_ filesInfo $ \fileInfo -> deleteFile user fileInfo
+              withStore' $ \db -> deleteGroupExpiredCIs db user gInfo expirationDate createdAtCutoff
+              ciCount_ <- withStore' $ \db -> getGroupCICount db user gInfo
```
epoberezkin, line 1437: same

src/Simplex/Chat/Store.hs
```diff
@@ -4107,88 +4094,62 @@ setChatItemTTL db User {userId} chatItemTTL = do
         "INSERT INTO settings (user_id, chat_item_ttl, created_at, updated_at) VALUES (?,?,?,?)"
         (userId, chatItemTTL, currentTs, currentTs)
 
-getExpiredFileInfo :: DB.Connection -> User -> UTCTime -> IO [CIFileInfo]
-getExpiredFileInfo db User {userId} expirationDate =
+getChatsList :: DB.Connection -> User -> IO [ChatRef]
```
epoberezkin, line 4097: this function can be unnecessary - why can't we just load contacts and groups separately (there are functions for that already I think? although they might be loading some extra data) and then to process in two different loops. Here we combine them into one type first and then pattern match back.

src/Simplex/Chat/Store.hs
epoberezkin: why not `min` outside of query

src/Simplex/Chat.hs
epoberezkin: and maybe they can be just one function then?

src/Simplex/Chat/Messages.hs
```diff
@@ -395,7 +395,7 @@ instance StrEncoding ACIFileStatus where
 -- to conveniently read file data from db
 data CIFileInfo = CIFileInfo
   { fileId :: Int64,
-    fileStatus :: ACIFileStatus,
+    fileStatus :: Maybe ACIFileStatus,
```
epoberezkin, line 398: this - not sure it needs to be Maybe here.

jr-simplex: it has information about direction.. I'd rather not supply default - we don't know where it might be used. As of now we could set some default in db based on existing usages - e.g. status as complete (still there's arbitrary direction then) so basically it doesn't try to cancel - but if some code tries to perform some action based on it I'd rather have it know that this information is unknown. The change in type barely affects code base, right it's only a couple of places, and it doesn't participate in api

epoberezkin (dismissed):

PR #1175, opened 2022-10-05
android: Animated switch between chat and chatList

avently:


jr-simplex:
@avently this crashes for me with `java.lang.NoSuchFieldError: No static field $stable of type I in class Lchat/simplex/app/model/ChatModel; or its superclasses ...` - not sure whether it's after latest merges of master, or specific state - I can help debugging

apps/android/app/src/main/java/chat/simplex/app/views/call/CallView.kt
```diff
@@ -432,10 +432,10 @@ fun WebRTCView(callCommand: MutableState<WCallCommand?>, onResponse: (WVAPIMessa
           }
         }
       ) { wv ->
-        Log.d(TAG, "WebRTCView: webview ready")
+        Log.d(TAG, "LALAL WebRTCView: webview ready")
```
jr-simplex, line 435:
```suggestion
        Log.d(TAG, "WebRTCView: webview ready")
```

avently: Oh, thanks

PR #1179, opened 2022-10-07
ios: update chats after setting chat item TTL

jr-simplex:


jr-simplex:
Currently this only has effect if chat is started.

epoberezkin (approved):

PR #1193, opened 2022-10-10
android: Multiple images can be sent at the same time in different messages

avently:


apps/android/app/src/main/java/chat/simplex/app/views/chat/ComposeView.kt
```diff
@@ -377,33 +380,38 @@ fun ComposeView(
...
+                    msgs.add(MsgContent.MCImage(if (msgs.isEmpty()) cs.message else "", preview.images[index]))
+                  }
+                }
+                is UploadContent.AnimatedImage -> {
+                  val file = saveAnimImage(context, it.uri)
+                  if (file != null) {
+                    files.add(file)
+                    msgs.add(MsgContent.MCImage(if (msgs.isEmpty()) cs.message else "", preview.images[index]))
+                  }
+                }
```
epoberezkin, line 404: can we move this if inside now? saveImage/saveAnimImage is the diff

PR #1200, opened 2022-10-12
android: german translations for v4.1

jr-simplex:


mlanp (commented):
Translations are ok! PR could be merged.

PR #1221, opened 2022-10-17
android: Some small fixes

avently:


apps/android/app/src/main/java/chat/simplex/app/views/chat/item/FramedItemView.kt
```diff
@@ -102,10 +106,16 @@ fun FramedItemView(
     }
   }
 
-  Surface(
-    shape = RoundedCornerShape(18.dp),
-    color = if (sent) SentColorLight else ReceivedColorLight
-  ) {
+  val transparentBackground = ci.content.msgContent is MsgContent.MCImage && ci.content.text.isEmpty()
```
epoberezkin, line 109: here it should check there is no quote too.

epoberezkin (approved):

PR #1228, opened 2022-10-19
core: Forcing chat unread

avently:


src/Simplex/Chat/Migrations/M20221019_chat_unread_status.hs
```diff
@@ -0,0 +1,14 @@
...
+module Simplex.Chat.Migrations.M20221019_chat_unread_status where
+
+import Database.SQLite.Simple (Query)
+import Database.SQLite.Simple.QQ (sql)
+
+m20221019_chat_unread_status :: Query
+m20221019_chat_unread_status =
+  [sql|
+ALTER TABLE contacts ADD COLUMN chat_unread INTEGER CHECK (chat_unread IN (NULL, 0, 1);
+ALTER TABLE groups ADD COLUMN chat_unread INTEGER CHECK (chat_unread IN (NULL, 0, 1);
```
epoberezkin, line 12:
```suggestion
PRAGMA ignore_check_constraints=ON;

ALTER TABLE contacts ADD COLUMN chat_unread INTEGER CHECK (chat_unread NOT NULL);
UPDATE contacts SET chat_unread = 0;

ALTER TABLE groups ADD COLUMN chat_unread INTEGER CHECK (chat_unread NOT NULL);
UPDATE groups SET chat_unread = 0;

PRAGMA ignore_check_constraints=OFF;
```

This way you won't need to deal with Maybe when reading from the table, and any non-0 value will be read as True, so no need to constrain to 0 and 1.

src/Simplex/Chat/Types.hs
```diff
@@ -214,7 +214,8 @@ groupName' GroupInfo {localDisplayName = g} = g
 
 -- TODO when more settings are added we should create another type to allow partial setting updates (with all Maybe properties)
 data ChatSettings = ChatSettings
-  { enableNtfs :: Bool
+  { enableNtfs :: Bool,
+    chatUnread:: Bool
```
epoberezkin, line 218: this is not the right place, it's not a setting. Better to put `unreadChat` into ChatStats (that has unreadCount now)

src/Simplex/Chat/Migrations/M20221019_unread_chat.hs
```diff
@@ -0,0 +1,20 @@
...
+
+m20221019_unread_chat :: Query
+m20221019_unread_chat =
+  [sql|
+PRAGMA ignore_check_constraints=ON;
+
+ALTER TABLE contacts ADD COLUMN unread_chat INTEGER CHECK (unread_chat NOT NULL);
+UPDATE contacts SET unread_chat = 0;
+ALTER TABLE groups ADD COLUMN unread_chat INTEGER CHECK (unread_chat NOT NULL);
+UPDATE groups SET unread_chat = 0;
```
epoberezkin, line 16:
```suggestion
ALTER TABLE contacts ADD COLUMN unread_chat INTEGER DEFAULT 0 CHECK (unread_chat NOT NULL);
UPDATE contacts SET unread_chat = 0;
ALTER TABLE groups ADD COLUMN unread_chat INTEGER DEFAULT 0 CHECK (unread_chat NOT NULL);
UPDATE groups SET unread_chat = 0;
```

PR #1229, opened 2022-10-19
core: allow to delete contacts that are in groups; group contacts management rfc

jr-simplex:


src/Simplex/Chat/Store.hs
```diff
@@ -565,9 +551,15 @@ deleteContactConnectionsAndFiles db userId Contact {contactId} = do
 deleteContact :: DB.Connection -> UserId -> Contact -> IO ()
 deleteContact db userId Contact {contactId, localDisplayName} = do
   DB.execute db "DELETE FROM chat_items WHERE user_id = ? AND contact_id = ?" (userId, contactId)
-  deleteContactProfile_ db userId contactId
+  ctMembers :: [ContactId] <- map fromOnly <$> DB.query db "SELECT contact_id FROM group_members WHERE user_id = ? AND contact_id = ?" (userId, contactId)
```
epoberezkin, line 554: why do you need the list, not count?

epoberezkin: also, there probably should be some similar extra logic when deleting members - if the profile is no longer used, remove it too.

epoberezkin: and local display name

epoberezkin: and when deleting groups

src/Simplex/Chat/Store.hs
```diff
@@ -565,9 +551,15 @@ deleteContactConnectionsAndFiles db userId Contact {contactId} = do
...
+  if null ctMembers
+    then do
+      deleteContactProfile_ db userId contactId
+      DB.execute db "DELETE FROM display_names WHERE user_id = ? AND local_display_name = ?" (userId, localDisplayName)
+    else do
+      currentTs <- getCurrentTime
+      DB.execute db "UPDATE group_members SET contact_id = NULL, updated_at = ? WHERE user_id = ? AND contact_id = ?" (currentTs, userId, contactId)
```
epoberezkin, line 561: also, we need to check what happens with Nothing contacts in members - what does it affect, that it doesn't make member inactive... And "Send direct message" option should be disabled/removed for such members (and I am not sure if we are removing it now for members while they are connecting).

PR #1235, opened 2022-10-20
android: Mark chat unread

avently:


apps/android/app/src/main/java/chat/simplex/app/views/chatlist/ChatListNavLinkView.kt
```diff
@@ -29,27 +29,29 @@ import kotlinx.datetime.Clock
 @Composable
 fun ChatListNavLinkView(chat: Chat, chatModel: ChatModel) {
   val showMenu = remember { mutableStateOf(false) }
-  var showMarkRead by remember { mutableStateOf(false) }
+  val showMarkRead = remember(chat.chatStats.unreadCount, chat.chatStats.unreadChat) {
+    chat.chatStats.unreadCount > 0 || chat.chatStats.unreadChat
+  }
+  val showMarkUnread = remember(chat.chatStats.unreadChat, showMarkRead) { !chat.chatStats.unreadChat && !showMarkRead }
```
epoberezkin, line 35: why do you need to different vars here if they are always the opposite? showMarkRead === !showMarkUnread

epoberezkin: I may be missing something, but I think you always show one or another button. Or is there a case when none can be shown?

apps/android/app/src/main/java/chat/simplex/app/views/chatlist/ChatListNavLinkView.kt
```diff
@@ -142,6 +147,9 @@ fun GroupMenuItems(chat: Chat, groupInfo: GroupInfo, chatModel: ChatModel, showM
       if (showMarkRead) {
         MarkReadChatAction(chat, chatModel, showMenu)
       }
+      if (showMarkUnread) {
```
epoberezkin, line 150:
```suggestion
      else {
```

apps/android/app/src/main/java/chat/simplex/app/views/chatlist/ChatListNavLinkView.kt
avently: That was stupid decision, never mind. Fixed

PR #1261, opened 2022-10-26
core: Chat preferences

avently:


epoberezkin:
- global user preferences -> contact_profiles, record that has user's profile
- preference of the contact -> `contact_profiles`, record that has contact's profile, field `preferences`
- user preferences for the contact with which the user is connected incognito (aka "incognito contact") - `contacts`, field `user_preferences` has the full set of preferences filled in at the time contact is created from global defaults
- user preferences for the normal contacts (with which the user shared the main profile) - `contacts`, field `user_preferences` has only overrides to global defaults that is initialised as NULL (or, possibly, as empty object) - in this case `preferences` in user's `contact_profiles` record would be NULL.

src/Simplex/Chat.hs
```diff
@@ -10,6 +10,8 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeApplications #-}
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
+{-# HLINT ignore "Use fewer imports" #-}
```
epoberezkin, line 14: I don't think these 2 lines are needed, please remove

src/Simplex/Chat.hs
```diff
@@ -640,6 +642,10 @@ processChatCommand = \case
     withCurrentCall contactId $ \userId ct call ->
       updateCallItemStatus userId ct call receivedStatus Nothing $> Just call
   APIUpdateProfile profile -> withUser (`updateProfile` profile)
+  APIUpdateOtherProfile contactId newUserPreferences -> withUser $ \user@User {userId} -> do
```
epoberezkin, line 645:
```suggestion
  APIUpdateOtherProfile contactId newUserPreferences -> withUser $ \user@User {userId} -> do
```

If it's about incognito profile, let's call it APIUpdateIncognitoProfile

src/Simplex/Chat.hs
```diff
@@ -1128,9 +1135,26 @@ processChatCommand = \case
...
-          forM_ contacts $ \ct ->
-            void (sendDirectContactMessage ct $ XInfo p') `catchError` (toView . CRChatError)
+          forM_ contacts $ \ct -> do
+            let Contact{profile = LocalProfile{profile = Profile{userPreferences = contactUserPreferences}}} = ct
+            let mergedPreferences = mergeChatPreferences globalUserPreferences contactUserPreferences
+            let mergedProfile = (p' :: Profile) {contactPreferences = mergedPreferences, userPreferences = Nothing}
+            void (sendDirectContactMessage ct $ XInfo mergedProfile) `catchError` (toView . CRChatError)
           pure $ CRUserProfileUpdated (fromLocalProfile p) p'
+    updateOtherProfile :: User -> Contact -> Maybe ChatPreferences -> Maybe ChatPreferences -> m ChatResponse
+    updateOtherProfile _user@User{userId, profile = LocalProfile{profileId, localAlias, profile = Profile{contactPreferences = globalUserPreferences}}} ct@Contact{contactId, profile = LocalProfile{profile = oldProfile}} oldUserPreferences newUserPreferences
```
epoberezkin, line 1145: the name is unclear - what does it do?

src/Simplex/Chat.hs
```diff
@@ -3109,6 +3153,7 @@ chatCommandP =
       "/_call end @" *> (APIEndCall <$> A.decimal),
       "/_call status @" *> (APICallStatus <$> A.decimal <* A.space <*> strP),
       "/_call get" $> APIGetCallInvitations,
+      "/_profile @" *> (APIUpdateOtherProfile <$> A.decimal <* A.space <*> ("userPreferences=" *> jsonP)),
```
epoberezkin, line 3156: the name is misleading as you cannot really change profile per contact, you are only changing the preference.

The command in rfc is `/_set prefs @ <json>`

src/Simplex/Chat.hs
epoberezkin: it's updateContactPrefs

src/Simplex/Chat.hs
```diff
@@ -1128,9 +1135,26 @@ processChatCommand = \case
...
+            let Contact{profile = LocalProfile{profile = Profile{userPreferences = contactUserPreferences}}} = ct
+            let mergedPreferences = mergeChatPreferences globalUserPreferences contactUserPreferences
+            let mergedProfile = (p' :: Profile) {contactPreferences = mergedPreferences, userPreferences = Nothing}
+            void (sendDirectContactMessage ct $ XInfo mergedProfile) `catchError` (toView . CRChatError)
           pure $ CRUserProfileUpdated (fromLocalProfile p) p'
+    updateOtherProfile :: User -> Contact -> Maybe ChatPreferences -> Maybe ChatPreferences -> m ChatResponse
+    updateOtherProfile _user@User{userId, profile = LocalProfile{profileId, localAlias, profile = Profile{contactPreferences = globalUserPreferences}}} ct@Contact{contactId, profile = LocalProfile{profile = oldProfile}} oldUserPreferences newUserPreferences
+      | oldUserPreferences == newUserPreferences = pure $ CRContactProfileUpdated ct -- nothing changed actually
+      | otherwise = do
+        res <- withStore $ \db -> updateContactUserPreferences db userId contactId newUserPreferences
```
epoberezkin, line 1148: the result here is (), so you don't need to put it anywhere

```suggestion
        withStore' $ \db -> updateContactUserPreferences db userId contactId newUserPreferences
```

src/Simplex/Chat.hs
```diff
@@ -2958,6 +2986,22 @@ deleteAgentConnectionAsync' user connId (AgentConnId acId) = do
   cmdId <- withStore' $ \db -> createCommand db user (Just connId) CFDeleteConn
   withAgent $ \a -> deleteConnectionAsync a (aCorrId cmdId) acId
 
+mergeChatPreferencesToProfile :: Profile -> Maybe ChatPreferences -> Maybe ChatPreferences -> Profile
+mergeChatPreferencesToProfile oldProfile globalUserPreferences contactUserPreferences = do
+    let mergedPreferences = mergeChatPreferences globalUserPreferences contactUserPreferences
+    (oldProfile :: Profile) {contactPreferences = mergedPreferences, userPreferences = Nothing}
+      
+mergeChatPreferences :: Maybe ChatPreferences -> Maybe ChatPreferences -> Maybe ChatPreferences
```
epoberezkin, line 2994:
```suggestion
mergeChatPreferences :: ChatPreferences -> Maybe ChatPreferences -> Maybe ChatPreferences
```

why user preferences are Maybe? they should be ChatPreferences I think

src/Simplex/Chat/Types.hs
```diff
@@ -226,10 +229,68 @@ defaultChatSettings = ChatSettings {enableNtfs = True}
...
+  toJSON = strToJSON
+  toEncoding = strToJEncoding
+
 data Profile = Profile
   { displayName :: ContactName,
     fullName :: Text,
-    image :: Maybe ImageData
+    image :: Maybe ImageData,
+    contactPreferences :: Maybe ChatPreferences,
+    userPreferences :: Maybe ChatPreferences
```
epoberezkin, line 293: this is incorrect, as we send Profile as a part of messages, so it should only include one of the following:
1. the full set of user preferences (if it's to be sent)
2. preference overrides per contact (if it's saved to the database)
3. the preferences received from the contact
 
In all cases it should be in one property `preferences`
```suggestion
    preferences :: Maybe ChatPreferences
```

epoberezkin: the other property should most likely be in Contact

src/Simplex/Chat/View.hs
```diff
@@ -691,6 +692,14 @@ viewUserProfileUpdated Profile {displayName = n, fullName, image} Profile {displ
   where
     notified = " (your contacts are notified)"
 
+viewContactProfileUpdated :: Contact -> [StyledString]
```
epoberezkin, line 695: we did not update profile here, only the preferences.

tests/ChatTests.hs
```diff
@@ -4,6 +4,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE PostfixOperators #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE PatternSynonyms #-}
```
epoberezkin, line 7: I don't think it's used here?

src/Simplex/Chat.hs
```diff
@@ -2958,6 +2987,22 @@ deleteAgentConnectionAsync' user connId (AgentConnId acId) = do
...
+mergeChatPreferences globalUserPreferences contactUserPreferences = case globalUserPreferences of
+  Nothing -> contactUserPreferences
+  Just contact -> case contactUserPreferences of
+    Nothing -> globalUserPreferences
+    Just global -> do
+      let ChatPreferences {voice = globalVoice} = global
+      let ChatPreferences {voice = contactVoice} = contact
+      let voice = fromMaybe globalVoice (Just contactVoice)
+      pure ChatPreferences {voice}
```
epoberezkin, line 3004:
```suggestion
mergeChatPreferences globalPrefs Nothing = globalPrefs
mergeChatPreferences ChatPreferences {voice = globalVoice} (Just ChatPreferences {voice = contactVoice}) = ChatPreferences {voice = contactVoice <|> globalVoice}
```

epoberezkin: the logic was incorrect, it would have never used globalVoice

src/Simplex/Chat.hs
```diff
@@ -1128,9 +1136,26 @@ processChatCommand = \case
...
         withChatLock "updateProfile" . procCmd $ do
-          forM_ contacts $ \ct ->
-            void (sendDirectContactMessage ct $ XInfo p') `catchError` (toView . CRChatError)
+          forM_ contacts $ \ct -> do
+            let Contact {profile = LocalProfile {profile = Profile {userPreferences = contactUserPreferences}}} = ct
+            let mergedPreferences = mergeChatPreferences globalUserPreferences contactUserPreferences
+            let mergedProfile = (p' :: Profile) {contactPreferences = mergedPreferences, userPreferences = Nothing}
+            void (sendDirectContactMessage ct $ XInfo mergedProfile) `catchError` (toView . CRChatError)
           pure $ CRUserProfileUpdated (fromLocalProfile p) p'
+    updateOtherProfile :: User -> Contact -> Maybe ChatPreferences -> Maybe ChatPreferences -> m ChatResponse
```
epoberezkin, line 1145: updateContactPrefs

src/Simplex/Chat.hs
```diff
@@ -1128,9 +1136,26 @@ processChatCommand = \case
           filter (\ct -> isReady ct && not (contactConnIncognito ct))
             <$> withStore' (`getUserContacts` user)
         withChatLock "updateProfile" . procCmd $ do
-          forM_ contacts $ \ct ->
-            void (sendDirectContactMessage ct $ XInfo p') `catchError` (toView . CRChatError)
+          forM_ contacts $ \ct -> do
+            let Contact {profile = LocalProfile {profile = Profile {userPreferences = contactUserPreferences}}} = ct
+            let mergedPreferences = mergeChatPreferences globalUserPreferences contactUserPreferences
+            let mergedProfile = (p' :: Profile) {contactPreferences = mergedPreferences, userPreferences = Nothing}
```
epoberezkin, line 1142: we should only be sending it if profile actually changed, it would be per-contact in this case, because in some cases the update of global default will not lead to the update of the user preferences for the particular contact (if override is the same as the new default).

tests/ChatTests.hs
avently: it's used here, see DefaultChatPrefs in that file

src/Simplex/Chat/View.hs
```diff
@@ -691,6 +692,18 @@ viewUserProfileUpdated Profile {displayName = n, fullName, image} Profile {displ
   where
     notified = " (your contacts are notified)"
 
+viewContactPrefsUpdated :: Contact -> [StyledString]
+viewContactPrefsUpdated Contact {profile = LocalProfile {preferences}, userPreferences = ChatPreferences {voice = userVoice}} =
+  let contactVoice = preferences >>= voice
+   in ["preferences were updated: " <> "contact's voice is " <> viewPreference contactVoice <> ", user's voice is " <> viewPreference userVoice]
```
jr-simplex, line 698: maybe a different phrasing would be better.. "voice" -> "voice messages", also have separate command for viewing current values? then this would be just "preferences were updated, use <cmd @contact_name> to view current preferences"

src/Simplex/Chat.hs
```diff
@@ -646,6 +646,9 @@ processChatCommand = \case
     withCurrentCall contactId $ \userId ct call ->
       updateCallItemStatus userId ct call receivedStatus Nothing $> Just call
   APIUpdateProfile profile -> withUser (`updateProfile` profile)
+  APISetContactPrefs contactId prefs' -> withUser $ \user@User {userId} -> do
```
jr-simplex, line 649: probably `viewContactUpdated` should show contact changed profile

src/Simplex/Chat/View.hs
avently: Not sure about that. Currently it's short and doesn't require any additional commands to execute for viewing a result. However voice messages looks better than voice. Will make it like this

tests/ChatTests.hs
```diff
@@ -107,9 +108,10 @@ chatTests = do
     it "accept contact request incognito" testAcceptContactRequestIncognito
     it "join group incognito" testJoinGroupIncognito
     it "can't invite contact to whom user connected incognito to a group" testCantInviteContactIncognito
-  describe "contact aliases" $ do
+  describe "contact aliases and prefs" $ do
     it "set contact alias" testSetAlias
     it "set connection alias" testSetConnectionAlias
+    it "set contact prefs" testSetContactPrefs
```
jr-simplex, line 114: following on latest changes - we should test changing preferences for incognito contact. Probably have both regular and incognito contact in the same test, and test changing profile and changing global preferences - both should affect regular contact but incognito contact shouldn't receive updates

jr-simplex: btw there doesn't seem to be an api for changing global preferences, or am I missing it?

avently: You mean user's global preferences? Then `/_profile` does this job. And `/_group_profile` for groups

epoberezkin (dismissed):

tests/ChatTests.hs
avently: > following on latest changes - we should test changing preferences for incognito contact. Probably have both regular and incognito contact in the same test, and test changing profile and changing global preferences - both should affect regular contact but incognito contact shouldn't receive updates

https://github.com/simplex-chat/simplex-chat/pull/1280

PR #1262, opened 2022-10-26
core: clean up incognito profiles

jr-simplex:
for reference:

table | field | used for incognito profiles
---|---|---
contacts | contact_profile_id | -
group_members | contact_profile_id | -
group_members | member_profile_id | +
contact_requests | contact_profile_id | -
groups | host_conn_custom_user_profile_id | - (not used at all)
connections | custom_user_profile_id | +

### DIFF
--- a/src/Simplex/Chat/Store.hs
+++ b/src/Simplex/Chat/Store.hs
@@ -580,15 +580,15 @@ deleteUnusedIncognitoProfileById_ db User {userId} profile_id =
     db
     [sql|
       DELETE FROM contact_profiles
-      WHERE 1 NOT IN (
+      WHERE user_id = :user_id AND contact_profile_id = :profile_id AND incognito = 1
+        AND 1 NOT IN (
           SELECT 1 FROM connections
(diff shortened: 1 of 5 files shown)


src/Simplex/Chat/Store.hs
```diff
@@ -572,6 +572,25 @@ deleteContact db userId Contact {contactId, localDisplayName} = do
...
+      DELETE FROM contact_profiles
+      WHERE 1 NOT IN (
+          SELECT 1 FROM connections
+          WHERE user_id = :user_id AND custom_user_profile_id = :profile_id LIMIT 1
+        )
+        AND 1 NOT IN (
+          SELECT group_member_id FROM group_members
+          WHERE user_id = :user_id AND member_profile_id = :profile_id LIMIT 1
+        )
+        AND user_id = :user_id AND contact_profile_id = :profile_id AND incognito = 1
```
epoberezkin, line 591: maybe better to put these conditions first?

PR #1277, opened 2022-10-31
core: switch connection

epoberezkin:


src/Simplex/Chat/View.hs
```diff
@@ -254,39 +254,31 @@ viewChatItem chat ChatItem {chatDir, meta, content, quotedItem, file} doShow = c
...
       where
         to = ttyToGroup g
     CIGroupRcv m -> case content of
       CIRcvMsgContent mc -> withRcvFile from $ rcvMsg from quote mc
-      CIRcvDeleted _ -> showRcvItem from
-      CIRcvCall {} -> showRcvItem from
       CIRcvIntegrityError err -> viewRcvIntegrityError from err meta
       CIRcvGroupInvitation {} -> showRcvItemProhibited from
-      CIRcvGroupEvent {} -> showRcvItem from
+      _ -> showRcvItem from
```
jr-simplex, line 281: I'd keep it as is and add new cases.. if we later add constructors that should have different view functions, this will be easier to not miss.
```suggestion
      _ -> showRcvItem from
```

epoberezkin: if you forget it will just show a standard text event which is what you want in most cases. And then if needed, it can be changed. There is no downside to having catch-all event here.

jr-simplex (approved):

PR #1281, opened 2022-11-01
core: update event name, ios: types/api/ui (wip) to switch connection to another address, fix contact/member info view, fix setting multiple servers

epoberezkin:


apps/ios/SimpleXChat/ChatTypes.swift
```diff
@@ -1498,6 +1507,44 @@ public enum SndGroupEvent: Decodable {
...
+        case let .switchQueue(phase, member):
+            var forMember = ""
+            if let m = member {
+                forMember = String.localizedStringWithFormat(NSLocalizedString(" for %@", comment: "for member name"), m.profile.profileViewName)
+            }
+            if case .completed = phase {
+                return String.localizedStringWithFormat(NSLocalizedString("you changed address%@", comment: "chat item text"), forMember)
+            }
+            return String.localizedStringWithFormat(NSLocalizedString("changing address%@...", comment: "chat item text"), forMember)
+        }
```
jr-simplex, line 1538: again this implies some similar structure in different languages, or makes it harder to translate out of context.. it was ok in terminal where it's only in English. Just make it separate texts depending on whether member is present or not?

jr-simplex: these text combinations in scope of a single message save a bit of code but then it may cause more problems than they're worth.. even for English only this is harder to understand than full messages with a single interpolation

PR #1292, opened 2022-11-03
core: update chat preferences

epoberezkin:


src/Simplex/Chat/Types.hs
```diff
@@ -230,70 +234,267 @@ defaultChatSettings = ChatSettings {enableNtfs = True}
...
+      receipts = Just receipts,
+      voice = Just voice
+    }
+
+defaultChatPrefs :: FullPreferences
+defaultChatPrefs =
+  FullPreferences
+    { fullDelete = Preference {allow = PSNo},
+      receipts = Preference {allow = PSNo},
+      voice = Preference {allow = PSNo}
```
jr-simplex, line 351: I think voice should be enabled by default - seems like most users want it for one reason or another, and not all will have patience to find / set up preferences. "No" disables full feature here, where in other cases it's "sub feature", e.g. full delete doesn't disable deletion altogether, and receipts we don't even have and it's not really necessary for communication + both of these probably have higher impact on privacy/control of the recipient.
```suggestion
      voice = Preference {allow = PSYes}
```

src/Simplex/Chat/Types.hs
```diff
@@ -230,70 +234,267 @@ defaultChatSettings = ChatSettings {enableNtfs = True}
...
+
+instance ToJSON Preference where toEncoding = J.genericToEncoding J.defaultOptions
 
-data PrefSwitch = PSOn | PSOff -- for example it can be extended to include PSMutual, that is only enabled if it's enabled by another party
+instance ToJSON GroupPreference where toEncoding = J.genericToEncoding J.defaultOptions
+
+data PrefAllowed
+  = PSAlways -- allow unconditionally
+  | PSYes -- allow, if peer allows it
+  | PSNo -- do not allo
```
jr-simplex, line 382:
```suggestion
  | PSNo -- do not allow
```

src/Simplex/Chat/Types.hs
```diff
@@ -230,70 +234,270 @@ defaultChatSettings = ChatSettings {enableNtfs = True}
...
-data PrefSwitch = PSOn | PSOff -- for example it can be extended to include PSMutual, that is only enabled if it's enabled by another party
+instance ToJSON GroupPreference where toEncoding = J.genericToEncoding J.defaultOptions
+
+data FeatureAllowed
+  = FAAlways -- allow unconditionally
+  | FAYes -- allow, if peer allows it
+  | FANo -- do not allow
+  deriving (Eq, Show, Generic)
+
+data GroupFeatureEnabled = FEOn | FEOff
```
jr-simplex, line 388:
```suggestion
data GroupFeatureEnabled = GFEOn | GFEOff
```

PR #1313, opened 2022-11-07
android: Connect via group link alert

avently:


apps/android/app/src/main/java/chat/simplex/app/views/newchat/ScanToConnectView.kt
```diff
@@ -49,10 +53,34 @@ fun ScanToConnectView(chatModel: ChatModel, close: () -> Unit) {
   )
 }
 
-fun withUriAction(uri: Uri, run: suspend (String) -> Unit) {
+enum class ConnectionLinkType {
+  CONTACT, INVITATION, GROUP
+}
+
+@Serializable
+sealed class CReqClientData {
```
jr-simplex, line 61: let's call it `CRData` as in swift

PR #1322, opened 2022-11-08
android: Fix of StackoverflowError

avently:


apps/android/app/src/main/java/chat/simplex/app/model/SimpleXAPI.kt
```diff
@@ -264,8 +264,17 @@ open class ChatController(var ctrl: ChatCtrl?, val ntfManager: NtfManager, val a
...
     if (receiverStarted) return
-    thread(name="receiver") {
-      GlobalScope.launch { withContext(Dispatchers.IO) { recvMspLoop() } }
+    receiverStarted = true
+    CoroutineScope(Dispatchers.IO).launch {
+      while (true) {
+        val ctrl = ctrl
+        if (ctrl == null) {
+          receiverStarted = false
+          return@launch
```
epoberezkin, line 273:
```suggestion
          break
```

epoberezkin: I understand there are cases for `goto` but let's please avoid it when there are more conventional alternatives...

apps/android/app/src/main/java/chat/simplex/app/model/SimpleXAPI.kt
```diff
@@ -264,8 +264,17 @@ open class ChatController(var ctrl: ChatCtrl?, val ntfManager: NtfManager, val a
   private fun startReceiver() {
     Log.d(TAG, "ChatController startReceiver")
     if (receiverStarted) return
-    thread(name="receiver") {
-      GlobalScope.launch { withContext(Dispatchers.IO) { recvMspLoop() } }
+    receiverStarted = true
+    CoroutineScope(Dispatchers.IO).launch {
+      while (true) {
+        val ctrl = ctrl
```
epoberezkin, line 270: why is it needed?

epoberezkin: if the idea to capture ctrl from scope to avoid any concurrency issues by the time of the call, I would name the variable `_ctrl` and also add the comment why we are doing it.

apps/android/app/src/main/java/chat/simplex/app/model/SimpleXAPI.kt
avently: Ok

apps/android/app/src/main/java/chat/simplex/app/model/SimpleXAPI.kt
avently: > I would name the variable _ctrl

Of course you would since you use Haskell:) But it's a Kotlin. In Haskell you wouldn't have this recursing problem at all. So, no need to copy Haskell decisions here 1:1

epoberezkin: So let’s just name it differently then, it’s not about Haskell, it’s a general hygiene to avoid shadowing

apps/android/app/src/main/java/chat/simplex/app/model/SimpleXAPI.kt
```diff
@@ -264,8 +264,20 @@ open class ChatController(var ctrl: ChatCtrl?, val ntfManager: NtfManager, val a
...
+      while (true) {
+        /** Global [ctrl] can be null. It's needed for having the same [ChatModel] that already made in [ChatController] without the need
+         * to change it everywhere in code after changing a database.
+         * Since it can be changed in background thread, making this check to prevent NullPointerException */
+        val ctrl = ctrl
+        if (ctrl == null) {
+          receiverStarted = false
+          break
+        }
+        val msg = recvMsg(ctrl)
```
epoberezkin, line 278:
```suggestion
        val _ctrl = ctrl
        if (_ctrl == null) {
          receiverStarted = false
          break
        }
        val msg = recvMsg(_ctrl)
```

PR #1331, opened 2022-11-09
android: Random crashes fix

avently:
Commits in order:
1,5. Prevents `ConcurrentModificationException` that could happen if a user repeatedly tried to add members to a group while previous attempt is in progress and working with a backend
2. Fixing `ForegroundServiceDidNotStartInTimeException` that could happen when a code starting a service and then quickly stops it until it has created. This situation was stopping the service from running `startForeground` function, which is required by Android
3. Restoring app's data from backup tools will still allow to enter passphrase instead of just crashing like it was before. When restoring happens, keys from keystore gets deleted and the app wasn't ready for this situation
4. Catched drawable's `DecodeException` that can happen on some devices for unknown reason. Exception has a message "Failed to create image decoder with message 'unimplemented'Input contained an error." so looks like the user use some non-popular image format that is not implemented in Android's ImageDecoder.

### DIFF
--- a/apps/android/app/src/main/java/chat/simplex/app/SimplexService.kt
+++ b/apps/android/app/src/main/java/chat/simplex/app/SimplexService.kt
@@ -10,7 +10,6 @@ import androidx.core.app.NotificationCompat
 import androidx.core.content.ContextCompat
 import androidx.work.*
 import chat.simplex.app.views.helpers.*
-import chat.simplex.app.views.onboarding.OnboardingStage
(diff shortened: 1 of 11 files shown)


jr-simplex:
split into different PRs:
https://github.com/simplex-chat/simplex-chat/pull/1348
https://github.com/simplex-chat/simplex-chat/pull/1349
https://github.com/simplex-chat/simplex-chat/pull/1350
https://github.com/simplex-chat/simplex-chat/pull/1351/files

PR #1348, opened 2022-11-11
android: prevent exception on concurrent attempts to add members to group

jr-simplex:


apps/android/app/src/main/java/chat/simplex/app/views/chat/group/AddGroupMembersView.kt
```diff
@@ -237,12 +242,14 @@ fun ContactCheckRow(
...
   } else if (checked) {
     icon = Icons.Filled.CheckCircle
-    iconColor = MaterialTheme.colors.primary
+    iconColor = if (enabled) MaterialTheme.colors.primary else HighOrLowlight
   } else {
     icon = Icons.Outlined.Circle
     iconColor = HighOrLowlight
   }
   SectionItemView(click = {
+    if (!enabled) return@SectionItemView
```
jr-simplex, line 251: @avently can we do `break` here?
```suggestion
    if (!enabled) break
```

apps/android/app/src/main/java/chat/simplex/app/views/chat/group/AddGroupMembersView.kt
```diff
@@ -35,14 +34,16 @@ import chat.simplex.app.views.usersettings.SettingsActionItem
...
+  var allowModifyMembers by remember { mutableStateOf(true) }
   BackHandler(onBack = close)
   AddGroupMembersLayout(
     groupInfo = groupInfo,
     contactsToAdd = getContactsToAdd(chatModel),
     selectedContacts = selectedContacts,
     selectedRole = selectedRole,
+    allowModifyMembers = allowModifyMembers,
     inviteMembers = {
+      allowModifyMembers = false
```
jr-simplex, line 46: @avently is it only set to true when view appears? i.e. it will still be disabled even after all members are invited if the view is not closed for some reason, and so it's because the view is automatically dismissed once operation is complete we don't set it to true anywhere else? e.g. can we set it to true after member invitations for loop?

apps/android/app/src/main/java/chat/simplex/app/views/chat/group/AddGroupMembersView.kt
avently: Nope. What are you trying to `break` here? This keyword works in loops (for example, in `while` loop to stop the loop).
Here `return` just skips all code block from executing

apps/android/app/src/main/java/chat/simplex/app/views/chat/group/AddGroupMembersView.kt
avently: Since we closing the view there is no need to set it to true. If we wouldn't close it, then yes, we would have to make it `true`

PR #1349, opened 2022-11-11
android: fix ForegroundServiceDidNotStartInTimeException

jr-simplex:
Fixes ForegroundServiceDidNotStartInTimeException that could happen when a code starting a service and then quickly stops it until it has created. This situation was stopping the service from running startForeground function, which is required by Android.

(cherry picked from commit a5d235e559edec51204c1551d20ee4aaddb8588f)

### DIFF
--- a/apps/android/app/src/main/java/chat/simplex/app/SimplexService.kt
+++ b/apps/android/app/src/main/java/chat/simplex/app/SimplexService.kt
@@ -10,7 +10,6 @@ import androidx.core.app.NotificationCompat
 import androidx.core.content.ContextCompat
 import androidx.work.*
 import chat.simplex.app.views.helpers.*
-import chat.simplex.app.views.onboarding.OnboardingStage
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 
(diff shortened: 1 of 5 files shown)


apps/android/app/src/main/java/chat/simplex/app/SimplexService.kt
```diff
@@ -62,7 +81,7 @@ class SimplexService: Service() {
 
   private fun startService() {
     Log.d(TAG, "SimplexService startService")
-    if (isServiceStarted || isStartingService) return
+    if (wakeLock != null || isStartingService) return
```
jr-simplex, line 84: I'm trying to understand what this achieves - is the `wakeLock != null` check added as a replacement for `isServiceStarted` or was it required before as well but missed? Is this line a sanity check or more of a fix?

avently: The behaviour of `isServiceStarted` variable is changed. Before that it was set `true` inside `startService` and that's why it was ok to check `if (isServiceStarted...` but now I need `isServiceStarted` following different logic: it has to be set `true` once the service is fully started and it has to be locked inside `companion object`. It is needed because I stop the service from that object to prevent unwanted behaviour caused by `stopSelf` function in which case it's called before calling `startForeground`.

So `wakeLock != null` is just a replacement of previously working logic that used `isServiceStarted` check because `wakeLock` will not be null only when it's set inside `startService` (it's set almost at the same place as `isServiceStarted` before).

PR #1362, opened 2022-11-13
cli: message search in CLI app

epoberezkin:


src/Simplex/Chat.hs
```diff
@@ -3256,7 +3256,8 @@ chatCommandP =
       ("\\ " <|> "\\") *> (DeleteMessage <$> chatNameP <* A.space <*> A.takeByteString),
       ("! " <|> "!") *> (EditMessage <$> chatNameP <* A.space <*> (quotedMsg <|> pure "") <*> A.takeByteString),
       "/feed " *> (SendMessageBroadcast <$> A.takeByteString),
-      ("/tail" <|> "/t") *> (LastMessages <$> optional (A.space *> chatNameP) <*> msgCountP),
+      ("/tail" <|> "/t") *> (LastMessages <$> optional (A.space *> chatNameP) <*> msgCountP <*> pure Nothing),
+      ("/search" <|> "/?") *> (LastMessages <$> optional (A.space *> chatNameP) <*> msgCountP <*> (Just <$> (A.space *> stringP))),
```
jr-simplex, line 3260: if search string starts with a number will it get parsed as a message count? maybe search should be separated with `search=`?

src/Simplex/Chat/View.hs
```diff
@@ -802,49 +802,47 @@ viewContactUpdated
...
-              && (timeOfDayToTime (localTimeOfDay localTime) > (6 * 60 * 60 :: DiffTime))
-              then "%m-%d" -- if message is from yesterday or before and 6 hours has passed since midnight
-              else "%H:%M"
-       in styleTime $ formatTime defaultTimeLocale format localTime
-
-viewSentMessage :: StyledString -> [StyledString] -> MsgContent -> CIMeta d -> [StyledString]
-viewSentMessage to quote mc = sentWithTime_ (prependFirst to $ quote <> prependFirst indent (ttyMsgContent mc))
+
+ttyMsgTime :: UTCTime -> ZonedTime -> StyledString
+ttyMsgTime ts t =
```
jr-simplex, line 815: it became harder to understand this without reading function's code and checking what's passed at call sites - change variable names to something more clear? Maybe this:
```suggestion
ttyMsgTime currentTs localItemTs =
```

src/Simplex/Chat.hs
epoberezkin: Yes, but it’s an edge case

src/Simplex/Chat/View.hs
epoberezkin: I think it's ok, it's consistently used in all functions...

PR #1368, opened 2022-11-15
core: voice msg content type

jr-simplex:


src/Simplex/Chat/Protocol.hs
```diff
@@ -322,6 +326,10 @@ msgContentText = \case
   MCText t -> t
   MCLink {text} -> text
   MCImage {text} -> text
+  MCVoice {text, duration} ->
```
jr-simplex, line 329: Not sure about `msgContentText` as we usually only put text here (which is not ideal in case of images as well). Though with this one we do not even plan to allow to include any text at all

src/Simplex/Chat/Protocol.hs
```diff
@@ -322,6 +326,10 @@ msgContentText = \case
...
+    if T.null text then voiceMessage else text <> "; " <> voiceMessage
+    where
+      voiceMessage = T.pack $ "voice message " <> show duration <> "s"
```
epoberezkin, line 332:
```suggestion
  MCVoice {text, duration} ->
    msg <> if T.null text then "" else "; " <> text
    where
      msg = T.pack $ "voice message " <> show duration <> "s"
```

Duration can be show as MM:SS (see line 888 in Messages.hs, that function can be moved here)

PR #1371, opened 2022-11-15
core, ios: advanced server config

epoberezkin:


apps/ios/Shared/Views/UserSettings/SMPServerView.swift
```diff
@@ -0,0 +1,105 @@
...
+@ViewBuilder func showTestStatus(server: ServerCfg) -> some View {
+    switch server.tested {
+    case .some(true):
+        Image(systemName: "checkmark")
+            .foregroundColor(.green)
+    case .some(false):
+        Image(systemName: "multiply")
+            .foregroundColor(.red)
+    case .none:
+        Color.clear
```
jr-simplex, line 81: why not
```suggestion
        EmptyView()
```

apps/ios/Shared/Views/UserSettings/SMPServerView.swift
```diff
@@ -0,0 +1,105 @@
...
+        Section("Use server") {
+            HStack {
+                Button("Test server") {
+                    Task { await testServerConnection(server: $server) }
+                }
+                Spacer()
+                showTestStatus(server: server)
+            }
+            Toggle("Enabled", isOn: $server.enabled)
+            Button("Remove server", role: .destructive) {
```
jr-simplex, line 65: remove button does nothing

apps/ios/SimpleXChat/APITypes.swift
```diff
@@ -289,6 +292,7 @@ public enum ChatResponse: Decodable, Error {
     case apiChats(chats: [ChatData])
     case apiChat(chat: ChatData)
     case userSMPServers(smpServers: [String])
+    case sMPTestResult(smpTestFailure: SMPTestFailure?)
```
jr-simplex, line 295: can api be adjusted so that it's all lowercase? or change to "serverTestResult"?
```suggestion
    case smpTestResult(smpTestFailure: SMPTestFailure?)
```

src/Simplex/Chat/Migrations/M20221115_server_cfg.hs
```diff
@@ -0,0 +1,19 @@
...
+import Database.SQLite.Simple (Query)
+import Database.SQLite.Simple.QQ (sql)
+
+m20221115_server_cfg :: Query
+m20221115_server_cfg =
+  [sql|
+PRAGMA ignore_check_constraints=ON;
+
+ALTER TABLE smp_servers ADD COLUMN preset INTEGER DEFAULT 0 CHECK (preset NOT NULL);
+ALTER TABLE smp_servers ADD COLUMN tested INTEGER;
```
jr-simplex, line 14: probably forgot to remove this..
```suggestion
ALTER TABLE smp_servers ADD COLUMN tested INTEGER;
```

apps/ios/Shared/Views/UserSettings/SMPServerView.swift
epoberezkin: Most buttons do nothing now

apps/ios/SimpleXChat/APITypes.swift
epoberezkin: Once build compiles

src/Simplex/Chat/Migrations/M20221115_server_cfg.hs
epoberezkin: No, I do think this status need to be saved, I don’t see reason not remember / show servers that were tested as suchC , retesting all when test&save is chosen.

apps/ios/Shared/Views/UserSettings/SMPServerView.swift
epoberezkin: EmptyView has no size

PR #1395, opened 2022-11-20
core: api to load all chat items

epoberezkin:


src/Simplex/Chat/Store.hs
```diff
@@ -3707,8 +3707,8 @@ getAllChatItems db user pagination search_ = do
   let search = fromMaybe "" search_
   case pagination of
     CPLast count -> getAllChatItemsLast_ db user count search
-    CPAfter _afterId _count -> throwError $ SEInternalError "not implemented"
-    CPBefore _beforeId _count -> throwError $ SEInternalError "not implemented"
+    CPAfter afterId count -> getAllChatItemsAfter_ db user afterId count search
+    CPBefore beforeId count -> getAllChatItemsBefore_ db user beforeId count search
```
jr-simplex, line 3711: what's going be the use case for this? just local testing/debugging? maybe add a couple of lines to some existing test to test each query at least once?

spaced4ndy (approved):

src/Simplex/Chat/Store.hs
epoberezkin: the old idea of all messages in one chat

PR #1402, opened 2022-11-22
core: chat items for group preferences

epoberezkin:


src/Simplex/Chat.hs
```diff
@@ -2186,12 +2198,19 @@ processAgentMessage (Just user@User {userId}) corrId agentConnId agentMessage =
     newContentMessage ct@Contact {localDisplayName = c, contactUsed, chatSettings} mc msg msgMeta = do
       unless contactUsed $ withStore' $ \db -> updateContactUsed db user ct
       checkIntegrityCreateItem (CDDirectRcv ct) msgMeta
-      let (ExtMsgContent content fileInvitation_) = mcExtMsgContent mc
-      ciFile_ <- processFileInvitation fileInvitation_ $ \db -> createRcvFileTransfer db userId ct
-      ci@ChatItem {formattedText} <- saveRcvChatItem user (CDDirectRcv ct) msg msgMeta (CIRcvMsgContent content) ciFile_
-      toView . CRNewChatItem $ AChatItem SCTDirect SMDRcv (DirectChat ct) ci
-      when (enableNtfs chatSettings) $ showMsgToast (c <> "> ") content formattedText
+      let ExtMsgContent content fileInvitation_ = mcExtMsgContent mc
+      case featureProhibited forContact user ct content of
```
jr-simplex, line 2202: It may be better to check on higher level - `processDirect/GroupMessage` - and pass event instead of message content to the check - next preference already will be prohibiting a chat protocol message, not message content type.

Also I'm not sure we should save the message in case it's prohibited (I thought we wouldn't) - so possibly `saveRcvMSG` should also be optional.

epoberezkin: will move there when needed

epoberezkin: also the next preference will not be prohibiting anything, it will change the behaviour of deletion.

PR #1406, opened 2022-11-23
core: config to send inline files

epoberezkin:
- set limit to send files without acceptance to 6 chunks (same as released limit to receive).
- increase limit to receive files sent without acceptance to 8 chunks (as iOS voice messages seem to be bigger, and currently limited to 30 sec).
- tests

### DIFF
--- a/src/Simplex/Chat/Controller.hs
+++ b/src/Simplex/Chat/Controller.hs
@@ -88,9 +88,9 @@ defaultInlineFilesConfig :: InlineFilesConfig
 defaultInlineFilesConfig =
   InlineFilesConfig
     { offerChunks = 15, -- max when chunks are offered / received with the option - limited to 255 on the encoding level
-      sendChunks = 0, -- max per file when chunks will be sent inline without acceptance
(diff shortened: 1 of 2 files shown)


src/Simplex/Chat/Controller.hs
```diff
@@ -88,9 +88,9 @@ defaultInlineFilesConfig :: InlineFilesConfig
 defaultInlineFilesConfig =
   InlineFilesConfig
     { offerChunks = 15, -- max when chunks are offered / received with the option - limited to 255 on the encoding level
-      sendChunks = 0, -- max per file when chunks will be sent inline without acceptance
+      sendChunks = 6, -- max per file when chunks will be sent inline without acceptance
       totalSendChunks = 30, -- max per conversation when chunks will be sent inline without acceptance
-      receiveChunks = 6 -- max when chunks are accepted
+      receiveChunks = 8 -- max when chunks are accepted
```
jr-simplex, line 93:
```suggestion
      receiveChunks = 9 -- max when chunks are accepted
```

PR #1451, opened 2022-11-28
android: Voice messages enhancements

avently:


apps/android/app/src/main/java/chat/simplex/app/views/chat/item/FramedItemView.kt
```diff
@@ -175,7 +175,7 @@ fun FramedItemView(
           }
         }
       }
-      if (ci.content.msgContent !is MsgContent.MCVoice || ci.content.text.isNotEmpty()) {
+      if (ci.content.msgContent !is MsgContent.MCVoice || ci.content.text.isNotEmpty() || ci.quotedItem != null) {
```
jr-simplex, line 178: I think CIMetaView shouldn't be conditional in FramedItemView. It seems to be so because CIVoiceView is inside FramedItemView, while it actually has no regular frame - we can refactor it similar to emoji. See EmojiItemView in ChatItemView, it uses EmojiText which is reused in FramedItemView. The resulting refactoring would be following: CIVoiceView is only used in ChatItemView (and it's a regular no text voice message w/t a frame, so it shouldn't be defined as part of FramedItemView which has logic for quotes, text, etc. which require frame); and there should be a separate FramedCIVoiceView which reuses player button and playback time text, which has to be used in FramedItemView.

So basically relation is the following:

Framed or not | Emoji view | Voice message view | Parent view
---|---|---|---
no frame | EmojiItemView (contains EmojiText) | CIVoiceView (contains player button and time primitives) | ChatItemView
framed | EmojiText | FramedCIVoiceView (uses player button and time primitives) | FramedItemView

jr-simplex: This would also allow to not account voice messages for background color in `transparentBackground` - images are different, they are framed, but they take up the whole frame so they're a special case. Some other branches may become obsolete as well

PR #1462, opened 2022-11-29
android: Different thread for withApi() call

avently:


epoberezkin:
As is this change is too risky, particularly it's removing withApi wrappers from some model changes. Will be adding withBGApi as a separate function instead and migrating performance critical parts to it one by one...

PR #1465, opened 2022-11-29
core, mobile: add group feature to allow direct messages

epoberezkin:


src/Simplex/Chat.hs
```diff
@@ -3132,11 +3132,11 @@ featureProhibited forWhom Contact {mergedPreferences} = \case
      in if forWhom enabled then Nothing else Just CFVoice
   _ -> Nothing
 
-groupFeatureProhibited :: GroupInfo -> MsgContent -> Maybe ChatFeature
+groupFeatureProhibited :: GroupInfo -> MsgContent -> Maybe GroupFeature
```
jr-simplex, line 3135: e should check "direct messages" feature when sending/processing new direct content message, somewhere around `featureProhibited` (not this function, direct one). I don't think relying only on UI is a good idea, especially considering we sometimes have bugs with state updates, etc., also as a sanity check in backend. And not even only just a sanity check actually - on group preference change sender may not receive this feature update (due to fragmentation w/ owner for example) - at least receiver will be "protected" and creating "rejected" items.

jr-simplex: we should also add a test for it

jr-simplex: the check can be based on the `directContact` function for example

epoberezkin: yes, next pr

PR #1476, opened 2022-12-01
prohibit direct messages to group contacts unless group preferences allow them

epoberezkin:


src/Simplex/Chat/Store.hs
```diff
@@ -1817,6 +1818,20 @@ getUserGroupDetails db User {userId, userContactId} =
...
+getContactGroupPreferences :: DB.Connection -> User -> Contact -> IO [FullGroupPreferences]
+getContactGroupPreferences db User {userId} Contact {contactId} = do
+  map (mergeGroupPreferences . fromOnly)
+    <$> DB.query
+      db
+      [sql|
+        SELECT gp.preferences
+        FROM groups g
+        JOIN group_profiles gp USING (group_profile_id)
+        JOIN group_members m USING (group_id)
```
jr-simplex, line 1830:
```suggestion
        JOIN group_profiles gp USING gp.group_profile_id = g.group_profile_id
        JOIN group_members m USING m.group_id = g.group_id
```

src/Simplex/Chat/Types.hs
```diff
@@ -123,6 +123,9 @@ directContact :: Contact -> Bool
 directContact Contact {contactUsed, activeConn = Connection {connLevel, viaGroupLink}} =
   (connLevel == 0 && not viaGroupLink) || contactUsed
 
+anyDirectContact :: Contact -> Bool
```
jr-simplex, line 126: why `directContact` doesn't suffice? I see the the difference in what they're checking, but don't understand the purpose, maybe name is not descriptive enough

tests/ChatTests.hs
```diff
@@ -3145,7 +3145,7 @@ testAllowFullDeletionGroup =
       threadDelay 1000000
       bob #> "#team hey"
       alice <# "#team bob> hey"
-      alice ##> "/full_delete #team on"
+      alice ##> "/set delete #team on"
```
jr-simplex, line 3148: add a simple test?

src/Simplex/Chat/Store.hs
jr-simplex: it's just easier to support afterwards, even if it's only two joins..

src/Simplex/Chat/Types.hs
epoberezkin: because we should not be prohibiting the direct messages to group admins, as otherwise they cannot even send the group invitation. We might special-case it and say that direct messages with this admin are only allowed until there is a group, but I don't see much damage to allow messaging the user whose link you used to join.

epoberezkin: at least on backend level.

This case was actually clear only after failing group links test, I didn't get it initially.

src/Simplex/Chat.hs
```diff
@@ -571,7 +571,7 @@ processChatCommand = \case
   APISendCallInvitation contactId callType -> withUser $ \user@User {userId} -> do
     -- party initiating call
     ct <- withStore $ \db -> getContact db user contactId
-    assertDirectAllowed user ct
+    assertDirectAllowed user MDSnd ct
```
jr-simplex, line 574: maybe this should live in `sendDirectContactMessage`?

PR #1503, opened 2022-12-05
core: delete unused contacts after deleting group

jr-simplex:


src/Simplex/Chat.hs
```diff
@@ -957,7 +957,11 @@ processChatCommand = \case
       toView . CRNewChatItem $ AChatItem SCTGroup SMDSnd (GroupChat gInfo) ci
       -- TODO delete direct connections that were unused
       deleteGroupLink' user gInfo `catchError` \_ -> pure ()
-      forM_ members $ deleteMemberConnection user
+      forM_ members $ \m -> do
+        deleteMemberConnection user m
+        withStore' $ \db -> do
+          itemId_ <- checkGroupMemberHasItems db user m
+          unless (isJust itemId_) $ deleteGroupMember db user m
```
epoberezkin, line 964: make this a function, it is used again below

epoberezkin: why deleteMemberItem is not called here?

src/Simplex/Chat.hs
```diff
@@ -513,6 +513,7 @@ processChatCommand = \case
     CTGroup -> do
       Group gInfo@GroupInfo {membership} members <- withStore $ \db -> getGroup db user chatId
       let canDelete = memberRole (membership :: GroupMember) == GROwner || not (memberCurrent membership)
+          contactIds = mapMaybe memberContactId members
```
epoberezkin, line 516: move it down, where it's used before forM

src/Simplex/Chat.hs
```diff
@@ -525,7 +526,18 @@ processChatCommand = \case
         withStore' $ \db -> deleteGroupConnectionsAndFiles db user gInfo members
         withStore' $ \db -> deleteGroupItemsAndMembers db user gInfo members
         withStore' $ \db -> deleteGroup db user gInfo
+        forM_ contactIds $ \ctId ->
+          deleteUnusedContact ctId `catchError` \_ -> pure ()
```
epoberezkin, line 530: we have catchAll_ for that, but shouldn't we send these errors to view?

src/Simplex/Chat.hs
```diff
@@ -525,7 +526,18 @@ processChatCommand = \case
...
         pure $ CRGroupDeletedUser gInfo
+      where
+        deleteUnusedContact contactId = do
+          ct <- withStore $ \db -> getContact db user contactId
+          unless (directContact ct) $ do
+            ctGroups <- withStore' $ \db -> getContactGroupProfiles db user ct
```
epoberezkin, line 536: why do we need to load the actual profiles instead of just checking they exist? This could be a function returning a boolean.

src/Simplex/Chat.hs
```diff
@@ -944,10 +956,8 @@ processChatCommand = \case
               ci <- saveSndChatItem user (CDGroupSnd gInfo) msg (CISndGroupEvent $ SGEMemberDeleted memberId (fromLocalProfile memberProfile)) Nothing Nothing
               toView . CRNewChatItem $ AChatItem SCTGroup SMDSnd (GroupChat gInfo) ci
               deleteMemberConnection user m
-              withStore' $ \db ->
-                checkGroupMemberHasItems db user m >>= \case
-                  Just _ -> updateGroupMemberStatus db userId m GSMemRemoved
-                  Nothing -> deleteGroupMember db user m
+              -- undeleted "member connected" chat item will prevent deletion of member record
+              deleteOrUpdateMemberRecord user m
```
epoberezkin, line 960: I thought we decided to never delete, as we need "member deleted" item?

src/Simplex/Chat/Store.hs
```diff
@@ -595,6 +598,14 @@ deleteContact db user@User {userId} Contact {contactId, localDisplayName, active
   DB.execute db "DELETE FROM contacts WHERE user_id = ? AND contact_id = ?" (userId, contactId)
   forM_ customUserProfileId $ \profileId -> deleteUnusedIncognitoProfileById_ db user profileId
 
+deleteContactNoGroups :: DB.Connection -> User -> Contact -> IO ()
```
epoberezkin, line 601: how is it different from any contact deletion? what does "no groups" mean?

src/Simplex/Chat.hs
jr-simplex: it seems to do a different thing.. ; I don't think we should - it's mostly to workaround around the fact that `getContact` is failable and in this context it realistically can fail if a contact was deleted before group

src/Simplex/Chat.hs
jr-simplex: it's what we'll need / close to that for getContactGroups api we for displaying in chat info

src/Simplex/Chat.hs
jr-simplex: "member deleted" item is created for deleting member

jr-simplex: this is the only case where I've added member record deletion in the end - it mirrors member removal by user

src/Simplex/Chat/Store.hs
jr-simplex: doesn't check whether contact is used for any members - it's covered by Chat logic that contact has no groups - see deleteUnusedContact -> when (null ctGroups).

it's the same as old deleteContact function (save for last line), when we didn't allow to delete contacts that were in any groups

PR #1511, opened 2022-12-06
android: Voice messages refactoring

avently:


apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
```diff
@@ -312,6 +184,157 @@ private fun NativeKeyboard(
...
+          android.view.Surface.ROTATION_270 -> ActivityInfo.SCREEN_ORIENTATION_REVERSE_LANDSCAPE
+          else -> ActivityInfo.SCREEN_ORIENTATION_PORTRAIT
+        }
+      }
+      // Unlock orientation
+      onDispose { activity.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED }
+    }
+    StopRecordButton(stopRecordingAndAddAudio)
+  } else {
+    val startStopRecording: () -> Unit = {
```
epoberezkin, line 216: why is it one function and not two?

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
```diff
@@ -312,6 +184,157 @@ private fun NativeKeyboard(
...
+              if (finished) {
+                recState.value = RecordingState.NotStarted
+              }
+            }
+          },
+        )
+      }
+    }
+    val interactionSource = interactionSourceWithTapDetection(
+      onPress = { if (recState.value.isNotStarted) startStopRecording() },
```
epoberezkin, line 235: for example here, you can only start recording, but you call the function that does both.

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
```diff
@@ -312,6 +184,157 @@ private fun NativeKeyboard(
...
+        if (state.stopRecOnNextClick.value) {
+          stopRecordingAndAddAudio()
+          state.stopRecOnNextClick.value = false
+        } else {
+          // tapped and didn't hold a finger
+          state.stopRecOnNextClick.value = true
+        }
+      },
+      onCancel = startStopRecording,
+      onRelease = startStopRecording
```
epoberezkin, line 247: and here you either only stop or do nothing but you call the function that can do both.

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
epoberezkin: and the second one already exists - stopRecordingAndAddAudio, so you need a function that only starts, not the one that does both

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
```diff
@@ -35,185 +36,56 @@ import androidx.core.view.inputmethod.InputConnectionCompat
...
+      val permissionsState = rememberMultiplePermissionsState(listOf(Manifest.permission.RECORD_AUDIO))
+      when {
+        showProgress -> ProgressIndicator()
+        showVoiceButton -> when {
+          needToAllowVoiceToContact ->
+            DisallowedVoiceButton { showNeedToAllowVoiceAlert(allowVoiceToContact) }
+          !allowedVoiceByPrefs ->
+            DisallowedVoiceButton { showDisabledVoiceAlert(isDirectChat) }
+          !permissionsState.allPermissionsGranted ->
+            DisallowedVoiceButton { permissionsState.launchMultiplePermissionRequest() }
```
epoberezkin, line 81: these choices can actually be inside the view callback (or determine which callback is passed, this is not that important, but no need to repeat DisallowedVoiceButton three times here.

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
```diff
@@ -312,6 +184,157 @@ private fun NativeKeyboard(
...
+      val activity = context as Activity
+      if ((recState.value as? RecordingState.Started)?.stopRecOnNextClick?.value == true) {
+        // Lock orientation to current orientation because screen rotation will break the recording
+        activity.requestedOrientation = when (activity.display?.rotation) {
+          android.view.Surface.ROTATION_90 -> ActivityInfo.SCREEN_ORIENTATION_LANDSCAPE
+          android.view.Surface.ROTATION_180 -> ActivityInfo.SCREEN_ORIENTATION_REVERSE_PORTRAIT
+          android.view.Surface.ROTATION_270 -> ActivityInfo.SCREEN_ORIENTATION_REVERSE_LANDSCAPE
+          else -> ActivityInfo.SCREEN_ORIENTATION_PORTRAIT
+        }
+      }
```
epoberezkin, line 210: whenever you write comment, it is usually the place where you lack a function name - create a function with a descriptive name, it'll be easier to read.

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
```diff
@@ -312,6 +184,157 @@ private fun NativeKeyboard(
...
+      // Unlock orientation
+      onDispose { activity.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED }
+    }
```
epoberezkin, line 213: same here

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
```diff
@@ -35,185 +36,56 @@ import androidx.core.view.inputmethod.InputConnectionCompat
...
-          }
-          DisposableEffect(Unit) {
-            onDispose {
-              rec.stop()
-            }
-          }
+  Box(Modifier.padding(vertical = 8.dp)) {
+    val cs = composeState.value
+    val attachEnabled = !composeState.value.editing
+    val showProgress = cs.inProgress && (cs.preview is ComposePreview.ImagePreview || cs.preview is ComposePreview.VoicePreview || cs.preview is ComposePreview.FilePreview)
```
epoberezkin, line 63: which progress is that? this seems incorrect. isn't progress always shown on send, why is it related to what is sent?

apps/android/app/src/main/java/chat/simplex/app/views/chat/ComposeView.kt
```diff
@@ -118,6 +117,18 @@ data class ComposeState(
...
+sealed class RecordingState {
+  object NotStarted: RecordingState()
+  class Started(val filePath: String, val stopRecOnNextClick: MutableState<Boolean> = mutableStateOf(false)): RecordingState()
+
+  val filePathNullable: String?
+    get() = (this as? Started)?.filePath
+  val isStarted: Boolean
+    get() = this is Started
+  val isNotStarted: Boolean
+    get() = this is NotStarted
```
epoberezkin, line 129: these seem unnecessary, they can be replaced with `is` in call sites

apps/android/app/src/main/java/chat/simplex/app/views/chat/ComposeView.kt
```diff
@@ -118,6 +117,18 @@ data class ComposeState(
   }
 }
 
+sealed class RecordingState {
+  object NotStarted: RecordingState()
+  class Started(val filePath: String, val stopRecOnNextClick: MutableState<Boolean> = mutableStateOf(false)): RecordingState()
```
epoberezkin, line 122: stopRecOnNextClick seems to be a local concern of a button, why is it needed on this level?

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
```diff
@@ -312,6 +184,157 @@ private fun NativeKeyboard(
...
+  DisposableEffect(Unit) { onDispose { rec.stop() } }
+  val stopRecordingAndAddAudio: () -> Unit = {
+    recState.value.filePathNullable?.let { onAudioAdded(it, rec.stop(), true) }
+    recState.value = RecordingState.NotStarted
+  }
+  if (recState.value.isStarted && (recState.value as RecordingState.Started).stopRecOnNextClick.value) {
+    val context = LocalContext.current
+    DisposableEffect((recState.value as? RecordingState.Started)?.stopRecOnNextClick?.value) {
+      val activity = context as Activity
+      if ((recState.value as? RecordingState.Started)?.stopRecOnNextClick?.value == true) {
```
epoberezkin, line 202: `stopRecOnNextClick` is only used here, so it can be defined locally.

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
```diff
@@ -312,6 +184,157 @@ private fun NativeKeyboard(
   }
 }
 
+@Composable
+fun RecordVoiceView(
+  recState: MutableState<RecordingState>,
+  onAudioAdded: (filePath: String, durationMs: Int, finished: Boolean) -> Unit,
```
epoberezkin, line 190: the reason you need this callback is exactly the lack of finished state that would have these parameters

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
```diff
@@ -312,6 +184,157 @@ private fun NativeKeyboard(
...
+  } else {
+    val startStopRecording: () -> Unit = {
+      if (recState.value.isStarted) {
+        stopRecordingAndAddAudio()
+      } else {
+        recState.value = RecordingState.Started(
+          filePath = rec.start { progress: Int?, finished: Boolean ->
+            if (recState.value.isStarted) {
+              if (progress != null) {
+                recState.value.filePathNullable?.let { onAudioAdded(it, progress, finished) }
```
epoberezkin, line 224: why finished needs to be passed to onAudioAdded? it's then used as parameter in the compose preview, but it feels like it means something else there.

again, better just remove this function completely and add a third state instead with LanchedEffect on state change (that should already exist somewhere already?)

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
avently: Who knows. Probably you wanted to show it only for "hard" operations, not the text ones, which is understandable.
This is where it's started:
https://github.com/simplex-chat/simplex-chat/commit/884231369f7c7ab3c184ccaffb994759e4ee0725#diff-3db6bd8f1ea1cbfed40e37cc5d395032d65655e6cb7440d58098556dea43a962R69

avently: I can make it showing like this `cs.inProgress && cs.preview !is ComposePreview.NoPreview`

avently: Or just `cs.inProgress && (cs.preview is ComposePreview.ImagePreview || cs.preview is ComposePreview.FilePreview)`

avently: It's quite annoying to see the spinner after every text message sending (image and file are ok)

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
avently: Made like this but I wouldn't do that because it adds two step check which is not good. We won nothing here with such change, only need to give more attention to two checks instead of to one check.
https://github.com/simplex-chat/simplex-chat/pull/1511/commits/1d56e5b217cd51558ec91741f4888373d78c6aea

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
avently: Comment is still needed even with a dedicated function, so a dev can understand why it's really needed to lock the orientation

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
avently: Comment is still needed even with a dedicated function, so a dev can understand why it's really needed to lock the orientation

apps/android/app/src/main/java/chat/simplex/app/views/chat/SendMsgView.kt
```diff
@@ -35,185 +36,57 @@ import androidx.core.view.inputmethod.InputConnectionCompat
...
-            Modifier.clip(CircleShape).background(color)
-          else
-            Modifier
-          IconButton({}, Modifier.size(36.dp), enabled = !cs.inProgress, interactionSource = interactionSource) {
-            Icon(
+  Box(Modifier.padding(vertical = 8.dp)) {
+    val cs = composeState.value
+    val attachEnabled = !composeState.value.editing
+    val showProgress = cs.inProgress && (cs.preview is ComposePreview.ImagePreview  || cs.preview is ComposePreview.FilePreview)
+    val showVoiceButton = cs.message.isEmpty() && showVoiceRecordIcon && attachEnabled &&
```
epoberezkin, line 63:
```suggestion
    val showProgress = cs.inProgress && (cs.preview is ComposePreview.ImagePreview  || cs.preview is ComposePreview.FilePreview)
    val showVoiceButton = cs.message.isEmpty() && showVoiceRecordIcon && !composeState.value.editing &&
```

PR #1512, opened 2022-12-06
android: Cancel notification after message deletion

avently:


apps/android/app/src/main/java/chat/simplex/app/views/chatlist/ChatPreviewView.kt
```diff
@@ -83,7 +83,7 @@ fun ChatPreviewView(chat: Chat, chatModelIncognito: Boolean, currentUserProfileD
     val ci = chat.chatItems.lastOrNull()
     if (ci != null) {
       MarkdownText(
-        if (!ci.meta.itemDeleted) ci.text else generalGetString(R.string.marked_deleted_description),
+        if (!ci.meta.itemDeleted) ci.text else if (ci.isPermDeleted) generalGetString(R.string.deleted_description) else generalGetString(R.string.marked_deleted_description),
```
jr-simplex, line 86: CIContent.RcvDeleted items didn't have flag itemDeleted before, it was basically unused after we changed group messages deletion (even earlier this flag was used to hide deleted but pending group messages). So this check is unnecessary and instead defaultDeleted item (in model) should have itemDeleted set to false, then the preview would be showing a different text.
```suggestion
        if (!ci.meta.itemDeleted) ci.text else generalGetString(R.string.marked_deleted_description),
```

apps/android/app/src/main/java/chat/simplex/app/model/ChatModel.kt
```diff
@@ -1202,6 +1207,25 @@ data class ChatItem (
...
+    val defaultDeleted: ChatItem
+      get() = ChatItem(
+        chatDir = CIDirection.DirectRcv(),
+        meta = CIMeta(
+          itemId = TEMP_DELETED_CHAT_ITEM_ID,
+          itemTs = Clock.System.now(),
+          itemText = generalGetString(R.string.deleted_description),
+          itemStatus = CIStatus.RcvRead(),
+          createdAt = Clock.System.now(),
+          itemDeleted = true,
```
jr-simplex, line 1221:
```suggestion
          itemDeleted = false,
```

apps/android/app/src/main/java/chat/simplex/app/model/ChatModel.kt
```diff
@@ -1202,6 +1207,25 @@ data class ChatItem (
...
+          itemEdited = false,
+          editable = false
+        ),
+        content = CIContent.RcvDeleted(deleteMode = CIDeleteMode.cidmBroadcast),
```
jr-simplex, line 1225: it probably would be cleaner if it was Rcv or SndDeleted depending on direction.. though it's a hack anyway and this would probably require passing additional parameters so overall might not be worth it..

apps/android/app/src/main/java/chat/simplex/app/model/SimpleXAPI.kt
```diff
@@ -1071,6 +1071,15 @@ open class ChatController(var ctrl: ChatCtrl?, val ntfManager: NtfManager, val a
         chatItemSimpleUpdate(r.chatItem)
       is CR.ChatItemDeleted -> {
         AudioPlayer.stop(r.deletedChatItem.chatItem)
+        val isLastChatItem = chatModel.getChat(r.deletedChatItem.chatInfo.id)?.chatItems?.lastOrNull()?.id == r.deletedChatItem.chatItem.id
+        if (isLastChatItem && ntfManager.hasNotificationsForChat(r.deletedChatItem.chatInfo.id)) {
+          ntfManager.cancelNotificationsForChat(r.deletedChatItem.chatInfo.id)
```
jr-simplex, line 1076: does it cancel notifications even for undeleted items?

jr-simplex: and it doesn't seem to remove notification for deleted item if it wasn't the last.. correct me if I'm wrong

avently: > does it cancel notifications even for undeleted items?

What do you call "undeleted items"?

> and it doesn't seem to remove notification for deleted item if it wasn't the last

It's expressed by `isLastChatItem`, yes. 
So, if you received message 1 and then message 2, then the sender deleted 1, notification will still be visible for 2. There is only one message at a time that is visible.

apps/android/app/src/main/java/chat/simplex/app/views/chatlist/ChatPreviewView.kt
avently: Thanks for the info

apps/android/app/src/main/java/chat/simplex/app/model/SimpleXAPI.kt
jr-simplex: > There is only one message at a time that is visible.

yep, that's the part I've missed

jr-simplex (approved):

PR #1530, opened 2022-12-08
core: verify connection

epoberezkin:


src/Simplex/Chat.hs
```diff
@@ -1135,6 +1139,20 @@ processChatCommand = \case
...
+      withStore (\db -> getContactIdByName db user cName) >>= processChatCommand . cmd
+    withMemberName :: GroupName -> ContactName -> (GroupId -> GroupMemberId -> ChatCommand) -> m ChatResponse
+    withMemberName gName mName cmd = withUser $ \user ->
+      getGroupAndMemberId user gName mName >>= processChatCommand . uncurry cmd
+    getConnectionCode :: ConnId -> m Text
+    getConnectionCode connId = verificationCode <$> withAgent (`getConnectionRatchetAdHash` connId)
+    verifyConnectionCode :: User -> Connection -> Text -> m ChatResponse
+    verifyConnectionCode user conn@Connection {connId} code = do
+      code' <- getConnectionCode $ aConnId conn
+      unless (sameVerificationCode code code') $ throwChatError $ CEBadConnectionCode code'
```
jr-simplex, line 1153: maybe it should be a ChatResponse instead? it's not chat's error really - the whole point is to verify this code, the supplied code can be different and the correct response should be that verification failed.

e.g. CRConnVerificationResult Bool

PR #1534, opened 2022-12-09
android: Fix for AddressAlreadyInUse exception

### LINKED ISSUE #1533: Crash at start - address already in use
Hello,

On m'y Pixel 6A with graphen OS it suite common to the app not starting.
OS provides me this message :
`
type: crash
osVersion: google/bluejay/bluejay:13/TQ1A.221205.011/2022120700:user/release-keys
package: chat.simplex.app:75
process: chat.simplex.app
processUptime: 80 + 236 ms

java.io.IOException: Address already in use
	at android.net.LocalSocketImpl.bindLocal(Native Method)
	at android.net.LocalSocketImpl.bind(LocalSocketImpl.java:275)
	at android.net.LocalServerSocket.<init>(LocalServerSocket.java:49)
	at chat.simplex.app.SimplexApp$Companion$1.invoke(SimplexApp.kt:176)
	at chat.simplex.app.SimplexApp$Companion$1.invoke(SimplexApp.kt:174)
	at kotlin.concurrent.ThreadsKt$thread$thread$1.run(Thread.kt:30)

`

Any advise?

avently:
Closes #1533

### DIFF
--- a/apps/android/app/src/main/java/chat/simplex/app/SimplexApp.kt
+++ b/apps/android/app/src/main/java/chat/simplex/app/SimplexApp.kt
@@ -11,8 +11,7 @@ import chat.simplex.app.views.onboarding.OnboardingStage
 import chat.simplex.app.views.usersettings.NotificationsMode
 import kotlinx.coroutines.*
 import kotlinx.serialization.decodeFromString
-import java.io.BufferedReader
-import java.io.InputStreamReader
+import java.io.*
 import java.util.*
 import java.util.concurrent.Semaphore
 import java.util.concurrent.TimeUnit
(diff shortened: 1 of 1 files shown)


apps/android/app/src/main/java/chat/simplex/app/SimplexApp.kt
```diff
@@ -173,7 +172,18 @@ class SimplexApp: Application(), LifecycleEventObserver {
       val s = Semaphore(0)
       thread(name="stdout/stderr pipe") {
         Log.d(TAG, "starting server")
-        val server = LocalServerSocket(socketName)
+        var server: LocalServerSocket? = null
+        for (i in 0..100_000 step 1234) {
```
epoberezkin, line 176:
```suggestion
        for (i in 0..1000) {
```

I don't think step 1234 makes it any more random tbh, as it still scans the same numbers, it just makes it more confusing...

epoberezkin: also trying ~1000 sockets may be slow, not sure, reduced to 100...

PR #1535, opened 2022-12-09
android: Notification mode selection in onboarding stage

avently:


apps/android/app/src/main/java/chat/simplex/app/views/onboarding/SetNotificationsMode.kt
```diff
@@ -0,0 +1,85 @@
...
+            Text(
+              mode.title,
+              fontWeight = FontWeight.Medium,
+              color = if (currentMode == mode.value) MaterialTheme.colors.primary else HighOrLowlight
+            )
+            Text(mode.description, color = MaterialTheme.colors.onBackground)
+          }
+        }
+        Spacer(Modifier.height(DEFAULT_PADDING))
+      }
```
epoberezkin, line 49: this should be just made a function that is called three times, then data structure won't be needed at all.

PR #1541, opened 2022-12-10
core: improve stability of file transfer handshake by using async agent commands

jr-simplex:


src/Simplex/Chat.hs
```diff
@@ -1407,48 +1408,44 @@ acceptFileReceive :: forall m. ChatMonad m => User -> RcvFileTransfer -> Maybe B
 acceptFileReceive user@User {userId} RcvFileTransfer {fileId, fileInvitation = FileInvitation {fileName = fName, fileConnReq, fileInline, fileSize}, fileStatus, grpMemberId} rcvInline_ filePath_ = do
   unless (fileStatus == RFSNew) $ case fileStatus of
     RFSCancelled _ -> throwChatError $ CEFileCancelled fName
-    _ -> pure () -- throwChatError $ CEFileAlreadyReceiving fName
+    _ -> throwChatError $ CEFileAlreadyReceiving fName
```
jr-simplex, line 1411: just reverting a two week ago fix that wasn't enough (we ended up with different fixes)

src/Simplex/Chat.hs
```diff
@@ -1407,48 +1408,44 @@ acceptFileReceive :: forall m. ChatMonad m => User -> RcvFileTransfer -> Maybe B
...
     -- group & direct file protocol
     Nothing -> do
       chatRef <- withStore $ \db -> getChatRefByFileId db user fileId
       case (chatRef, grpMemberId) of
         (ChatRef CTDirect contactId, Nothing) -> do
           ct <- withStore $ \db -> getContact db user contactId
-          (msg, ci) <- acceptFile
-          void $ sendDirectContactMessage ct msg
-          pure ci
+          acceptFile CFCreateConnFileAcptInvDirect $ \msg -> void $ sendDirectContactMessage ct msg
```
jr-simplex, line 1424: I'll think about a different name and refactoring

src/Simplex/Chat/Store.hs
```diff
@@ -1917,6 +1917,28 @@ getGroupMember db user@User {userId} groupId groupMemberId =
       |]
       (groupId, groupMemberId, userId)
 
+getGroupMemberById :: DB.Connection -> User -> GroupMemberId -> ExceptT StoreError IO GroupMember
```
jr-simplex, line 1920: getGroupMember function above checks additionally by groupId - this seems redundant since groupMemberId uniquely identifies group member - in other places we either had or passed (Switch API for example required passing) extra data for it - for files INV it required extra (or more difficult) query I believe.. I decided to basically repeat getGroupMember but we can keep and use only this one

epoberezkin: > for files INV it required extra (or more difficult) query I believe

didn't get it. Why is it a problem to have groupId as well? The point of it was a sanity check.

jr-simplex: RcvFileTransfer is quite ad-hoc and doesn't have it, it does have group member id though - seemed easier to use that than to load group id into file transfers, or querying group_id from files just to get group member (query for which doesn't really need it).

jr-simplex: group member is required for its connection - to send direct message

src/Simplex/Chat.hs
epoberezkin: it cannot be removed until we send inline to UI, as it now happens in groups.

PR #1542, opened 2022-12-10
ios: verify connection security code

epoberezkin:


apps/ios/Shared/Views/Chat/VerifyCodeView.swift
```diff
@@ -0,0 +1,126 @@
...
+
+                    Text(splitCode)
+                        .multilineTextAlignment(.leading)
+                        .font(.body.monospaced())
+                        .lineLimit(20)
+                        .padding(.bottom, 8)
+                }
+                .frame(maxWidth: .infinity, alignment: .center)
+
+                Text("To verify end-to-end encryption with your contact compare (or scan) the code on your devices.")
```
jr-simplex, line 52: parameterize view with text so it says "group member" in place of "your contact" in groups?

jr-simplex: full sentence not just subject

epoberezkin: Don’t think it’s important

epoberezkin: Nobody will verify group members I think…

apps/ios/Shared/Views/Chat/Group/GroupMemberInfoView.swift
```diff
@@ -36,75 +38,92 @@ struct GroupMemberInfoView: View {
...
+        }
+    }
+
+    private func groupMemberInfoView() -> some View {
+        VStack {
+            List {
+                groupMemberInfoHeader(member)
+                    .listRowBackground(Color.clear)
 
+                Section {
```
jr-simplex, line 54: how does it look if there's no send and no verify code button?

jr-simplex: though the second one should always be present..

epoberezkin: As nothing

epoberezkin: Maybe a bit more white space. Verify is only present for active members, not while they are connecting.

epoberezkin: Btw why we are showing send message for inactive?

PR #1554, opened 2022-12-12
ios, android: show send direct message button only for active members

jr-simplex:


apps/ios/Shared/Views/Chat/Group/GroupMemberInfoView.swift
```diff
@@ -52,7 +52,7 @@ struct GroupMemberInfoView: View {
                     .listRowBackground(Color.clear)
 
                 Section {
-                    if let contactId = member.memberContactId {
+                    if let contactId = member.memberContactId, member.memberActive {
```
epoberezkin, line 55: maybe put memberActive condition outside of the section? (it does a bit of whitespace when it's empty)

apps/ios/Shared/Views/Chat/Group/GroupMemberInfoView.swift
```diff
@@ -52,7 +52,7 @@ struct GroupMemberInfoView: View {
...
                         if let chat = chatModel.getContactChat(contactId),
                            chat.chatInfo.contact?.directContact ?? false {
```
epoberezkin, line 57: This condition also needs to change to allow sending to indirect contacts if there are messages with them.

jr-simplex: they're used in this case and this condition is true

PR #1569, opened 2022-12-13
ios: live messages

epoberezkin:


apps/ios/Shared/Views/Chat/ComposeMessage/ComposeView.swift
```diff
@@ -307,6 +322,10 @@ struct ComposeView: View {
             if let fileName = composeState.voiceMessageRecordingFileName {
                 cancelVoiceMessageRecording(fileName)
             }
+            if composeState.liveMessage != nil {
+                sendMessage()
+                resetLinkPreview()
```
jr-simplex, line 327: sendMessage already calls clearState(), is it not enough?

epoberezkin: not synchronously.

PR #1573, opened 2022-12-14
android: Show error instead of crashing after failed to parse chats

avently:


apps/android/app/src/main/java/chat/simplex/app/model/SimpleXAPI.kt
```diff
@@ -404,14 +404,17 @@ open class ChatController(var ctrl: ChatCtrl?, val ntfManager: NtfManager, val a
 
   suspend fun apiGetChats(): List<Chat> {
     val r = sendCmd(CC.ApiGetChats())
-    if (r is CR.ApiChats ) return r.chats
-    throw Exception("failed getting the list of chats: ${r.responseType} ${r.details}")
+    if (r is CR.ApiChats) return r.chats
```
epoberezkin, line 407: the problem here is that it won't return at all if JSON fails to parse, we need a catch all sub-class CR.InvalidResponse(String) that will be returned in case JSON doesn't parse, otherwise it would still crash.

The further improvement would be not to fail the whole list, but only fail individual list items, both hear and in the chat - this would require parse it as a generic JSON list and then parsing each item as an individual type. I don't think we should do it in this PR, but at least catching JSON parse errors is needed, otherwise it doesn't solve anything - it is never the case that incorrect response is returned.

apps/android/app/src/main/java/chat/simplex/app/model/SimpleXAPI.kt
```diff
@@ -404,14 +404,17 @@ open class ChatController(var ctrl: ChatCtrl?, val ntfManager: NtfManager, val a
...
+    Log.e(TAG, "failed getting the list of chats: ${r.responseType} ${r.details}")
+    AlertManager.shared.showAlertMsg(generalGetString(R.string.failed_to_parse_chats_title), generalGetString(R.string.contact_developers))
+    return emptyList()
   }
 
   suspend fun apiGetChat(type: ChatType, id: Long, pagination: ChatPagination = ChatPagination.Last(ChatPagination.INITIAL_COUNT), search: String = ""): Chat? {
     val r = sendCmd(CC.ApiGetChat(type, id, pagination, search))
-    if (r is CR.ApiChat ) return r.chat
+    if (r is CR.ApiChat) return r.chat
```
epoberezkin, line 415: same here.

apps/android/app/src/main/java/chat/simplex/app/model/SimpleXAPI.kt
epoberezkin: It's relatively easy to test by adding some required field to Chat class that is not present in JSON - it won't show alert in this case I think...

avently: > It's relatively easy to test by adding some required field to Chat class that is not present in JSON - it won't show alert in this case I think...

This is literally what I did before making this PR.
I made a commit for your tests:  https://github.com/simplex-chat/simplex-chat/pull/1573/commits/f9c9a20ab63c01d89276c5e4a6054566d553f3c5

Revert it before merging, of course. It works

PR #1609, opened 2022-12-19
core: do not start disappearing timer for live messages until they stop being live, and start it on item update instead, provided they are read

epoberezkin:


src/Simplex/Chat.hs
```diff
@@ -1748,6 +1742,12 @@ cleanupManager user = do
       timedItems <- withStore' $ \db -> getTimedItems db user startTimedThreadCutoff
       forM_ timedItems $ uncurry (startTimedItemThread user)
 
+startProximateTimedItemThread :: ChatMonad m => User -> (ChatRef, ChatItemId) -> UTCTime -> m ()
+startProximateTimedItemThread user itemRef deleteAt = do
+  ts <- liftIO getCurrentTime
+  when (diffInMicros deleteAt ts <= cleanupManagerInterval * 1000000) $
```
jr-simplex, line 1748: add diffInSeconds to Util, the second argument won't have to be converted

``` haskell
diffInSeconds :: UTCTime -> UTCTime -> Int
diffInSeconds a b = (`div` 1000000000000) $ diffInPicos a b
```

src/Simplex/Chat/Store.hs
```diff
@@ -3846,24 +3847,33 @@ updateDirectChatItem db userId contactId itemId newContent live msgId_ = do
...
-      (newContent, newText, edited', live', currentTs, userId, contactId, itemId)
-  pure ci {content = newContent, meta = (meta ci) {itemText = newText, itemEdited = edited', itemLive = live'}, formattedText = parseMaybeMarkdownList newText}
+      (newContent, newText, edited', delAt', live', currentTs, userId, contactId, itemId)
+  pure ci {content = newContent, meta = meta {itemText = newText, itemEdited = edited', itemTimed = timed', itemLive = live'}, formattedText = parseMaybeMarkdownList newText}
   where
     correctDir :: CChatItem c -> Either StoreError (ChatItem c d)
     correctDir (CChatItem _ ci) = first SEInternalError $ checkDirection ci
 
+-- the condition to enable the timed deletion when the item that was live is updated
+ciLiveDeleteAt :: CIMeta d -> Bool -> UTCTime -> Maybe UTCTime
```
jr-simplex, line 3871: maybe move this logic to Chat, and add an explicit check item went from live to not in startUpdatedTimedItemThread as a sanity check

jr-simplex: I guess getDirectChatItem would have to be moved from Store update functions to Chat as well then

src/Simplex/Chat/Store.hs
```diff
@@ -3846,24 +3847,33 @@ updateDirectChatItem db userId contactId itemId newContent live msgId_ = do
...
+ciLiveDeleteAt CIMeta {itemTimed, itemStatus = CISRcvNew} _live _ = itemTimed >>= deleteAt
```
jr-simplex, line 3872: what's the purpose of this branch?

src/Simplex/Chat.hs
epoberezkin: Why? Even diffInMicros is unnecessary, you could just use the fact that it can be converted to seconds

src/Simplex/Chat/Store.hs
epoberezkin: Not to start timer of the unread item

src/Simplex/Chat/Store.hs
epoberezkin: That’s the reason I didn’t want to change it all - it’s ok as is

epoberezkin: In any case if it’s to be done it should be a separate PR

epoberezkin: startUpdatedTimedItemThread - it has the check that deleteAt was added, why does it matter for what reason?

jr-simplex: btw items are already being read in Chat before update both on send and receive

PR #1614, opened 2022-12-20
ios: disappearing messages

epoberezkin:


apps/ios/Shared/Views/Chat/ChatItem/FramedCIVoiceView.swift
```diff
@@ -72,5 +72,6 @@ struct FramedCIVoiceView_Previews: PreviewProvider {
                 .environmentObject(ChatModel())
         }
         .previewLayout(.fixed(width: 360, height: 360))
+        .environmentObject(Chat.sampleData)
```
jr-simplex, line 75: remove .environmentObject(ChatModel()) here too?

jr-simplex (approved):

PR #1620, opened 2022-12-21
core: chat items with offered feature

epoberezkin:


src/Simplex/Chat/Messages.hs
```diff
@@ -809,6 +813,8 @@ ciContentToText = \case
   CISndConnEvent event -> sndConnEventToText event
   CIRcvChatFeature feature enabled param -> chatFeatureToText feature <> ": " <> prefToText enabled param
   CISndChatFeature feature enabled param -> chatFeatureToText feature <> ": " <> prefToText enabled param
+  CIRcvFeatureOffer feature param -> "allow " <> featureParamText feature param <> "?"
+  CISndFeatureOffer feature param -> "offered " <> featureParamText feature param
```
jr-simplex, line 817:
```suggestion
  CIRcvFeatureOffer feature param -> "offered " <> featureParamText feature param
  CISndFeatureOffer feature param -> "you offered " <> featureParamText feature param
```

jr-simplex: why not use prefParamText instead of featureParamText? "off" on Nothing param is unclear, it will be something like this:
"alice> offered disappearing messages (off)" - what off means here?
or w/t change from "allow?" - even more strange:
"alice> allow disappearing messages (off)?"

src/Simplex/Chat.hs
```diff
@@ -3425,13 +3427,46 @@ userProfileToSend user@User {profile = p} incognitoProfile ct =
...
+    createItem f
+      | state /= state' = create $ uncurry (ciFeature f') state'
+      | offer = create $ ciOffer f' param'
+      | otherwise = pure ()
+      where
+        create content = createInternalChatItem user (chatDir ct') content Nothing
+        f' = chatFeature f
+        state = featureState cup
+        state' = featureState cup'
+        offer = (allow' == FAYes || allow' == FAAlways) && (allow == FANo || param /= param')
```
jr-simplex, line 3464: why does this affect only disappearing messages in tests? shouldn't it create offers for other features as well?

jr-simplex: I think I understand, it's because other features have items created on state /= state' condition, so they won't have interactive offer.. not sure it's desired behaviour

epoberezkin: one-way state changes can be interactive too

jr-simplex (approved):

PR #1656, opened 2022-12-28
docs: user profiles rfc

jr-simplex:


docs/rfcs/2022-12-17-user-profiles.md
```diff
@@ -0,0 +1,130 @@
...
+
+- terminal APIs
+
+### Subscriptions
+
+- get connections for all users, subscribe to all - to receive events for users other than active
+
+- map subscriptions to users?
+
+- option to subscribe to connections for only one user? `StartChat {allUsers :: Bool}` API
```
epoberezkin, line 37: seems unnecessary, as the reason to turn off chat is to avoid database notifications.

docs/rfcs/2022-12-17-user-profiles.md
```diff
@@ -0,0 +1,130 @@
...
+
+- APISetActiveUser UserId
+
+- CRUsersList {users :: [User]}
+
+- terminal APIs
+
+### Subscriptions
+
+- get connections for all users, subscribe to all - to receive events for users other than active
```
epoberezkin, line 33: There are two options here:
1. map subscriptions to users at the time of subscription (and creating connection), so agent doesn't store information about which subscription belongs to which user, but has some sort of session ID per user.
2. associate user ID with connection in the agent database.

Currently, I like option 1 more as it would allow to prioritise subscriptions to the currently active profile.

docs/rfcs/2022-12-17-user-profiles.md
```diff
@@ -0,0 +1,130 @@
...
+
+- map subscriptions to users?
+
+- option to subscribe to connections for only one user? `StartChat {allUsers :: Bool}` API
+
+- if more than one user profiles - start with last active user (persist?), or always load list of users first and wait for choice? same for terminal?
```
epoberezkin, line 39: last active makes sense, I don't like the UI when the first the user see is the list of profiles. It should be clear which profile is currently active - avatar should be visible in the list of chats, we would probably need some equivalent if avatar is not set.

docs/rfcs/2022-12-17-user-profiles.md
```diff
@@ -0,0 +1,130 @@
...
+- events need to have information for which user they happened
+
+  ```haskell
+  data UserChatResponse = UserChatResponse
+    { user :: User,
+      chatResponse :: ChatResponse
+    }
+  ```
+
+- replace ChatResponse with UserChatResponse in APIResponse, toView, etc.
```
epoberezkin, line 66: or maybe just add user field when it is relevant...

docs/rfcs/2022-12-17-user-profiles.md
```diff
@@ -0,0 +1,130 @@
...
+
+- `AgentConfig` should depend on network configuration per user?
+
+- double check other state
+
+### Storage
+
+- schema already has support for multiple users
+
+- persist last active user?
```
epoberezkin, line 104: also supported

docs/rfcs/2022-12-17-user-profiles.md
```diff
@@ -0,0 +1,130 @@
...
+
+- persist user settings? (see below)
+
+### Frontend
+
+- view with list of user profiles available from settings, change active user from there
+
+- do we need additional information in that view? for example, number of unread messages
+
+- network configuration is stored in app preferences - if unchanged will be the same across users
```
epoberezkin, line 114: I think network settings being the same across users might be ok... There is the question of SOCKS though

docs/rfcs/2022-12-17-user-profiles.md
```diff
@@ -0,0 +1,130 @@
...
+
+  - persist in database per user and load?
+
+  - different configuration across platforms
+
+  - save network settings in separate preference for each user as json? dynamic preferences or predefined number of users?
+
+- same for other settings - auto-accept images, send link previews, etc.
+
+### Terminal view
```
epoberezkin, line 124: I don't think it's necessary to support multi-user in terminal. We need something for testing, but it doesn't have to be too user friendly, and maybe only in test view.

PR #1658, opened 2022-12-28
mobile: preserve group description in profile

epoberezkin:


apps/ios/SimpleX.xcodeproj/xcshareddata/xcschemes/SimpleX (iOS).xcscheme
```diff
@@ -44,7 +44,7 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      enableUBSanitizer = "YES"
+      enableAddressSanitizer = "YES"
```
jr-simplex, line 47: this probably should be disabled

epoberezkin: does it run for you without it? I don't think it affects release anyway.

PR #1714, opened 2023-01-09
ios: Live messages without sending an empty text

avently:


apps/ios/Shared/Views/Chat/ComposeMessage/ComposeView.swift
```diff
@@ -232,43 +245,47 @@ struct ComposeView: View {
...
+            case (true, .voicePreview): EmptyView() // ? we may allow playback when editing is allowed
+            default: previewView()
             }
             HStack (alignment: .bottom) {
                 Button {
                     showChooseSource = true
                 } label: {
                     Image(systemName: "paperclip")
-                        .resizable()
+                            .resizable()
```
epoberezkin, line 257:
```suggestion
                        .resizable()
```

apps/ios/Shared/Views/Chat/ComposeMessage/ComposeView.swift
```diff
@@ -232,43 +245,47 @@ struct ComposeView: View {
...
                 }
-                .disabled(composeState.attachmentDisabled)
-                .frame(width: 25, height: 25)
-                .padding(.bottom, 12)
-                .padding(.leading, 12)
+                        .disabled(composeState.attachmentDisabled)
+                        .frame(width: 25, height: 25)
+                        .padding(.bottom, 12)
+                        .padding(.leading, 12)
```
epoberezkin, line 262:
```suggestion
                .disabled(composeState.attachmentDisabled)
                .frame(width: 25, height: 25)
                .padding(.bottom, 12)
                .padding(.leading, 12)
```

apps/ios/Shared/Views/Chat/ComposeMessage/SendMessageView.swift
```diff
@@ -220,6 +226,20 @@ struct SendMessageView: View {
...
+            cancel()
+        } label: {
+            Image(systemName: "multiply")
+                    .resizable()
+                    .scaledToFit()
+                    .foregroundColor(.accentColor)
+                    .frame(width: 15, height: 15)
+        }
+                .frame(width: 29, height: 29)
+                .padding([.bottom, .horizontal], 4)
```
epoberezkin, line 240:
```suggestion
        .frame(width: 29, height: 29)
        .padding([.bottom, .horizontal], 4)
```

apps/ios/Shared/Views/Chat/ComposeMessage/ComposeView.swift
```diff
@@ -232,43 +245,47 @@ struct ComposeView: View {
...
+                        },
+                        finishVoiceMessageRecording: finishVoiceMessageRecording,
+                        allowVoiceMessagesToContact: allowVoiceMessagesToContact,
+                        onImagesAdded: { images in if !images.isEmpty { chosenImages = images }},
+                        keyboardVisible: $keyboardVisible
                 )
-                .padding(.trailing, 12)
-                .background(.background)
+                        .padding(.trailing, 12)
+                        .background(.background)
```
epoberezkin, line 288:
```suggestion
                .padding(.trailing, 12)
                .background(.background)
```

apps/ios/Shared/Model/ChatModel.swift
```diff
@@ -230,7 +230,13 @@ final class ChatModel: ObservableObject {
             }
             return false
         } else {
-            withAnimation { reversedChatItems.insert(cItem, at: 0) }
+            withAnimation {
+                if reversedChatItems.first?.id == ChatItem.TEMP_LIVE_CHAT_ITEM_ID {
+                    reversedChatItems.insert(cItem, at: min(1, reversedChatItems.count))
+                } else {
+                    reversedChatItems.insert(cItem, at: 0)
+                }
```
epoberezkin, line 238: refactor using ternary for at: parameter

apps/ios/Shared/Model/ChatModel.swift
```diff
@@ -274,6 +280,26 @@ final class ChatModel: ObservableObject {
...
 
+    func addLiveChatItemDummy(_ quotedCItem: ChatItem?, _ chatInfo: ChatInfo) -> ChatItem {
+        var quoted: CIQuote? = nil
+        if let quotedItem = quotedCItem, let msgContent = quotedItem.content.msgContent {
+            quoted = CIQuote.getSampleWithMsgContent(itemId: quotedItem.id, sentAt: quotedItem.meta.updatedAt, msgContent: msgContent, chatDir: quotedItem.chatDir)
+        }
+        var direct = false
+        if case .direct = chatInfo {
+            direct = true
+        }
```
epoberezkin, line 291: bool should not be used here

apps/ios/Shared/Views/Chat/ChatView.swift
```diff
@@ -462,7 +462,7 @@ struct ChatView: View {
         private func menu() -> [UIAction] {
             var menu: [UIAction] = []
             if let mc = ci.content.msgContent, !ci.meta.itemDeleted || revealed {
-                if !ci.meta.itemDeleted {
+                if !ci.meta.itemDeleted && ci.id != ChatItem.TEMP_LIVE_CHAT_ITEM_ID {
```
epoberezkin, line 465: using special ID for an item to indicate that it is special is a hack, a better approach is to have a flag that is not included in JSON.

apps/ios/Shared/Views/Chat/ComposeMessage/ComposeView.swift
```diff
@@ -19,10 +19,22 @@ enum ComposePreview {
     case filePreview(fileName: String)
 }
 
-enum ComposeContextItem {
+enum ComposeContextItem: Equatable {
```
epoberezkin, line 22: making it Equatable is a bad idea, there should be another way

apps/ios/Shared/Views/Chat/ComposeMessage/ComposeView.swift
```diff
@@ -232,43 +245,47 @@ struct ComposeView: View {
...
+                        sendMessage: {
+                            sendMessage()
+                            resetLinkPreview()
+                        },
+                        sendLiveMessage: sendLiveMessage,
+                        updateLiveMessage: updateLiveMessage,
+                        cancelLiveMessage: {
+                            composeState = composeState.copy(liveMessage: nil)
+                            chatModel.removeLiveChatItemDummy()
+                        },
```
epoberezkin, line 274: there is no need to pass this function as a parameter given that composeState is passed to SendMessageView and that the model is available there too

apps/ios/Shared/Views/Chat/ComposeMessage/ComposeView.swift
```diff
@@ -35,6 +47,7 @@ struct LiveMessage {
     var chatItem: ChatItem
     var typedMsg: String
     var sentMsg: String
+    var sent: Bool
```
epoberezkin, line 50:
```suggestion
    var sentMsg: String?
```

this is redundant, as you could make sentMsg optional

apps/ios/Shared/Views/Chat/ComposeMessage/ComposeView.swift
```diff
@@ -280,6 +297,16 @@ struct ComposeView: View {
...
+        .onChange(of: composeState.contextItem) { _ in
+            if composeState.liveMessage?.sent == false {
+                chatModel.removeLiveChatItemDummy()
+                var quoted: ChatItem? = nil
+                if case let .quotedItem(item) = composeState.contextItem {
+                    quoted = item
+                }
+                let _ = chatModel.addLiveChatItemDummy(quoted, chat.chatInfo)
+            }
+        }
```
epoberezkin, line 309: this should not be a change listener, as this can only happen during a function call that you control - it should happen in that function

apps/ios/Shared/Views/Chat/ComposeMessage/ComposeView.swift
```diff
@@ -396,10 +424,19 @@ struct ComposeView: View {
     private func sendLiveMessage() async {
         let typedMsg = composeState.message
         let sentMsg = truncateToWords(typedMsg)
-        if composeState.liveMessage == nil,
+        if !sentMsg.isEmpty && (composeState.liveMessage == nil || composeState.liveMessage?.sent == false),
```
epoberezkin, line 427: this condition cannot be correct, as if:
1. message is not empty
2. AND it is already sent
it will be sent again if this function is called

epoberezkin: ah, no.

epoberezkin: possibly it's better to swap branches making creating dummy the first branch, to make conditions non-negative and less confusing.

apps/ios/Shared/Views/Chat/ComposeMessage/SendMessageView.swift
```diff
@@ -103,6 +104,10 @@ struct SendMessageView: View {
                         }
                     } else if vmrs == .recording && !holdingVMR {
                         finishVoiceMessageRecordingButton()
+                    } else if composeState.liveMessage?.sent == false && composeState.message.isEmpty {
+                        cancelLiveMessageButton {
+                            cancelLiveMessage?()
```
epoberezkin, line 109: this function can be defined locally

apps/ios/Shared/Views/Chat/ComposeMessage/ComposeView.swift
epoberezkin: you can always include it via view definition or via ChatModel.shared - it's a global singleton

epoberezkin: In this case it's also EnvironmentObject - so cleaner to get it there.

apps/ios/Shared/Views/Chat/ComposeMessage/ComposeView.swift
avently: There is no way to swap branches since the second branch allows to be executed with `livemessage == nil` like the first branch. If I will swap them, I have to write almost identical code to it, so nothing actually changes in terms of less confusing

epoberezkin: there is of course, but ok...

PR #1716, opened 2023-01-10
core: test async file transfer (sender & receiver restarts); close files in stopChatController; handle openFile error in getFileHandle

jr-simplex:


src/Simplex/Chat.hs
```diff
@@ -3252,8 +3260,7 @@ getFileHandle fileId filePath files ioMode = do
   maybe (newHandle fs) pure h_
   where
     newHandle fs = do
-      -- TODO handle errors
-      h <- liftIO (openFile filePath ioMode)
+      h <- liftIO (openFile filePath ioMode) `E.catch` (throwChatError . CEFileInternal . (show :: E.SomeException -> String))
```
epoberezkin, line 3263:
```suggestion
      h <- liftIO (openFile filePath ioMode) `catchAll` throwChatError . CEFileInternal . show
```

PR #1725, opened 2023-01-11
ios: fix opening via notification returning to chat list when screen lock is enabled

avently:


apps/ios/Shared/Views/ChatList/ChatListView.swift
```diff
@@ -80,7 +80,7 @@ struct ChatListView: View {
                 destination: chatView(),
                 isActive: Binding(
                     get: { chatModel.chatId != nil },
-                    set: { _, _ in chatModel.chatId = nil }
+                    set: { _, _ in }
```
epoberezkin, line 83:
```suggestion
                    set: { active in
                        print(active)
                        if !active { chatModel.chatId = nil }
                    }
```

try to see what this does and when this is called. The way it was seems wrong indeed.

PR #1729, opened 2023-01-12
core: catch errors when sending messages in loops where they haven't been previously caught

jr-simplex:


src/Simplex/Chat.hs
```diff
@@ -1327,13 +1329,16 @@ processChatCommand = \case
...
-              when (directOrUsed ct') $ createSndFeatureItems user' ct ct'
+            processContact user' ct `catchError` (toView . CRChatError)
           pure $ CRUserProfileUpdated (fromLocalProfile p) p'
+      where
+        processContact user' ct = do
+          let mergedProfile = userProfileToSend user Nothing $ Just ct
+              ct' = updateMergedPreferences user' ct
+              mergedProfile' = userProfileToSend user' Nothing $ Just ct'
+          when (mergedProfile' /= mergedProfile) $ do
+            void (sendDirectContactMessage ct' $ XInfo mergedProfile')
```
epoberezkin, line 1340:
```suggestion
            void . sendDirectContactMessage ct' $ XInfo mergedProfile'
```

PR #1755, opened 2023-01-15
terminal: options for log level and internal queue sizes

epoberezkin:


src/Simplex/Chat/View.hs
```diff
@@ -218,8 +218,8 @@ responseToView user_ testView liveItems ts = \case
     ]
   CRAgentStats stats -> map (plain . intercalate ",") stats
   CRConnectionDisabled entity -> viewConnectionEntityDisabled entity
-  CRMessageError prefix err -> [plain prefix <> ": " <> plain err]
-  CRChatError e -> viewChatError e
+  CRMessageError prefix err -> [plain prefix <> ": " <> plain err | prefix == "error" || logLevel <= CLLWarning]
```
jr-simplex, line 221: shouldn't it be `>=`?
```suggestion
  CRMessageError prefix err -> [plain prefix <> ": " <> plain err | prefix == "error" || logLevel >= CLLWarning]
```

jr-simplex (approved):

PR #1793, opened 2023-01-18
ios: Multiusers feature continue

avently:


apps/ios/Shared/Model/NtfManager.swift
```diff
@@ -46,8 +54,8 @@ class NtfManager: NSObject, UNUserNotificationCenterDelegate, ObservableObject {
                 chatModel.ntfContactRequest = chatId
             }
         } else if let (chatId, ntfAction) = ntfCallAction(content, action) {
-            if let invitation = chatModel.callInvitations.removeValue(forKey: chatId) {
-                CallController.shared.callAction(invitation: invitation, action: ntfAction)
+            if let userId = userId, let invitation = chatModel.callInvitations.removeValue(forKey: chatId) {
+                CallController.shared.callAction(invitation: (userId, invitation), action: ntfAction)
```
epoberezkin, line 58: I am not yet sure why userId is needed here, but if it is it's a hack to pass it as a part of invitation parameter without making it a part of invitation. So either it should be inside invitation, or a separate parameter...

apps/ios/SimpleXChat/APITypes.swift
```diff
@@ -530,99 +530,99 @@ public enum ChatResponse: Decodable, Error {
             case .chatRunning: return noDetails
             case .chatStopped: return noDetails
             case .chatSuspended: return noDetails
-            case let .apiChats(chats): return String(describing: chats)
-            case let .apiChat(chat): return String(describing: chat)
-            case let .userSMPServers(smpServers, presetServers): return "smpServers: \(String(describing: smpServers))\npresetServers: \(String(describing: presetServers))"
-            case let .smpTestResult(smpTestFailure): return String(describing: smpTestFailure)
-            case let .chatItemTTL(chatItemTTL): return String(describing: chatItemTTL)
+            case let .apiChats(_, chats): return String(describing: chats)
```
epoberezkin, line 533: the point of these strings is to debug... it's probably better to include userId as the first line in console response views, so we can see what it is...

I would probably define something like `showUser(user, <current string>)` that prepends it, it'll be less noisy than if you just change interpolations everywhere...

apps/ios/Shared/Model/ChatModel.swift
```diff
@@ -16,7 +16,7 @@ final class ChatModel: ObservableObject {
     @Published var onboardingStage: OnboardingStage?
     @Published var v3DBMigration: V3DBMigrationState = v3DBMigrationDefault.get()
     @Published var currentUser: User?
-    @Published var users: [UserInfo] = []
+    @Published private(set) var users: [UserInfo] = []
```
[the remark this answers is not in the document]
avently: I want to limit setter to only this class (ChatModel). So other classes can only interact with updateUsers() function which sorts users after every update and puts active user at the top. This way we have consistent state always.

apps/ios/Shared/Model/NtfManager.swift
avently: I'm not entirelly sure yet how to make this thing good, will do in a separate PR

apps/ios/SimpleXChat/APITypes.swift
avently: I named it `withUser`

avently: https://github.com/simplex-chat/simplex-chat/pull/1793/commits/1e80b354db519fa470f40c7a4d8bf072d13efca4

apps/ios/SimpleXChat/APITypes.swift
```diff
@@ -530,104 +530,108 @@ public enum ChatResponse: Decodable, Error {
...
+            case let .chatCmdError(u, chatError): return withUser(u, String(describing: chatError))
+            case let .chatError(u, chatError): return withUser(u, String(describing: chatError))
             }
         }
     }
 
     private var noDetails: String { get { "\(responseType): no details" } }
+
+    private func withUser(_ u: User?, _ s: String) -> String {
+        "userId: \(u?.userId ?? -1)\n\(s)"
```
epoberezkin, line 633:
```suggestion
        if let id = u?.userId {
            return "userId: \(id)\n\(s)"
        }
        return s
```

apps/ios/Shared/Model/ChatModel.swift
```diff
@@ -395,6 +397,24 @@ final class ChatModel: ObservableObject {
...
+    func decreaseUnreadCounter(user: User, by: Int = 1) {
+        changeUnreadCounter(user: user, by: -by)
+    }
+
+    private func changeUnreadCounter(user: User, by: Int) {
+        for i in 0..<users.count {
+            let u = users[i]
+            if u.user.id == user.id {
+                users[i].unreadCount += Int64(by)
+            }
```
jr-simplex, line 417:
```suggestion
        if let i = users.firstIndex(where: {$0.user.id == user.id}) {
            users[i].unreadCount += Int64(by)
```

apps/ios/Shared/Views/UserSettings/UserProfilesView.swift
```diff
@@ -37,8 +38,18 @@ struct UserProfilesView: View {
...
-    private func userProfileView(_ userBinding: Binding<UserInfo>, index: Int) -> some View {
-        let user = userBinding.wrappedValue.user
+    private func removeUser(index: Int) {
+        do {
+            try apiDeleteUser(m.users[index].user.userId)
+            var users = m.users
+            users.remove(at: index)
+            m.updateUsers(users)
+        } catch {
+            fatalError("Failed to delete user: \(responseError(error))")
```
jr-simplex, line 48: alert instead of fatalError?

epoberezkin: yes, in another PR as deletion needs to be changed to swipe actions anyway

PR #1801, opened 2023-01-19
ios: user profiles view, per-user settings

epoberezkin:


apps/ios/Shared/Views/Database/DatabaseView.swift
```diff
@@ -67,6 +67,23 @@ struct DatabaseView: View {
...
+                    if case .seconds = chatItemTTL {
+                        Text(chatItemTTL.deleteAfterText).tag(chatItemTTL)
+                    }
+                }
+                .frame(height: 36)
+                .disabled(m.chatDbChanged || progressIndicator)
+            } header: {
+                Text("Messages")
+            } footer: {
+                Text("This setting applies to messages in your current chat profile **\(m.currentUser?.displayName ?? "")**.")
```
jr-simplex, line 84: if I remember correctly some other views use chatViewName, probably should align

apps/ios/Shared/Views/UserSettings/UserProfilesView.swift
```diff
@@ -9,57 +9,90 @@ import SimpleXChat
...
+        }
+    }
 
     var body: some View {
         List {
-            Section("Your profiles") {
-                ForEach(Array(m.users.enumerated()), id: \.0) { i, userInfo in
-                    userProfileView(userInfo, index: i)
-                        .deleteDisabled(userInfo.user.activeUser)
+            Section {
```
jr-simplex, line 28: There's a following bug: if you change a user (switch or create a new one), then Edit users list without closing the view, the initial active user cannot be deleted.

Regardless, maybe we should close settings view after changing user?

apps/ios/Shared/Views/Database/DatabaseView.swift
epoberezkin: I don't think it's needed in this comment, it can be long...

PR #1803, opened 2023-01-20
ios: maintain connections network statuses map separately from chats (allows to keep track of network statuses for all users)

jr-simplex:


apps/ios/Shared/Model/ChatModel.swift
```diff
@@ -522,6 +516,10 @@ final class ChatModel: ObservableObject {
             logger.error("Unable to set active user: \(error.localizedDescription)")
         }
     }
+
+    func connNetworkStatus(_ conn: Connection) -> NetworkStatus {
+        return networkStatuses[conn.connId] ?? .unknown
```
epoberezkin, line 521:
```suggestion
        networkStatuses[conn.connId] ?? .unknown
```

PR #1824, opened 2023-01-23
ios: enable swipe to go back from chat to list

avently:


apps/ios/Shared/Views/ChatList/ChatListView.swift
```diff
@@ -66,7 +66,7 @@ struct ChatListView: View {
         .onDisappear() { withAnimation { userPickerVisible = false } }
         .offset(x: -8)
         .listStyle(.plain)
-        .navigationTitle("Your chats")
+        .navigationTitle("Chats")
```
epoberezkin, line 69:
```suggestion
        .navigationTitle("Your chats")
```

PR #1830, opened 2023-01-24
core: use batch delete api when deleting unused group contacts

jr-simplex:


src/Simplex/Chat.hs
```diff
@@ -630,18 +630,26 @@ processChatCommand = \case
...
+                then pure []
+                else do
+                  ctGroupId <- withStore' $ \db -> checkContactHasGroups db user ct
+                  if isNothing ctGroupId
+                    then pure []
+                    else do
+                      conns <- withStore $ \db -> getContactConnections db userId ct
+                      withStore' (\db -> deleteContactWithoutGroups db user ct)
+                        `catchError` (toView . CRChatError (Just user))
+                      pure $ map aConnId conns
```
epoberezkin, line 652:
```suggestion
          withStore (\db -> getContact db user contactId) >>= delete `catchError` (\e -> toView (CRChatError (Just user) e) >> pure [])
          where
            delete ct
              | directOrUsed ct = pure []
              | otherwise = do
                  withStore' (\db -> checkContactHasGroups db user ct) >>= \case
                    Nothing -> pure []
                    _ -> do
                      conns <- withStore $ \db -> getContactConnections db userId ct
                      withStore' (\db -> deleteContactWithoutGroups db user ct)
                        `catchError` (toView . CRChatError (Just user))
                      pure $ map aConnId conns
```

epoberezkin (approved):

PR #1836, opened 2023-01-24
ios: fix layout of voice message

epoberezkin:


jr-simplex (changes requested):
this seems to break layout, chat item both on sent and received side twitches from side to side when being played

PR #1837, opened 2023-01-25
mobile: use GMT timezone in filenames to prevent leaking user location

epoberezkin:


apps/android/app/src/main/java/chat/simplex/app/views/helpers/Util.kt
```diff
@@ -390,15 +388,23 @@ fun saveFileFromUri(context: Context, uri: Uri): String? {
   }
 }
 
+fun generateNewFileName(context: Context, prefix: String, ext: String): String {
+  val sdf = SimpleDateFormat("yyyyMMdd_HHmmss", Locale.US)
+  sdf.timeZone = TimeZone.getTimeZone("GMT")
+  val timestamp = sdf.format(Date())
+  return uniqueCombine(context, "${prefix}_$timestamp.$ext")
```
jr-simplex, line 395:
```suggestion
  return uniqueCombine(context, "$prefix_$timestamp.$ext")
```

jr-simplex (approved):

apps/android/app/src/main/java/chat/simplex/app/views/helpers/Util.kt
epoberezkin: I also thought it would work... but there is underscore...

PR #1841, opened 2023-01-25
core: speed up get chats queries [exploration]

jr-simplex:


epoberezkin:
removing one index solved the problem

PR #1909, opened 2023-02-07
android: prevent showing system alert after start (Android 13+)

avently:


apps/android/app/src/main/java/chat/simplex/app/model/NtfManager.kt
```diff
@@ -268,6 +264,21 @@ class NtfManager(val context: Context, private val appPreferences: AppPreference
     }
   }
 
+  /**
+   * [force] param bypasses check for enabled notifications. When app is started for the first time, Android 13+ will ask a user
+   * if he allows to enable notifications from the app. Since we don't want to show this alert immediately after start, we're creating
+   * notification channels a bit later to not trigger this system alert
+   * */
+  fun ensureNtfChannelsExist(force: Boolean = false) {
+    if (!force && !manager.areNotificationsEnabled()) return
```
epoberezkin, line 273: it does seem like the boolean value should be flipped as it is very difficult to understand the logic and the intention here. Also it's unclear what is the meaning of the name force

epoberezkin: most likely this parameter needs to be removed and this condition be at call site, it's unclear why there is a need to pass boolean

PR #1933, opened 2023-02-14
ios: native WebRTC

avently:


apps/ios/Shared/Views/Call/ActiveCallView.swift
```diff
@@ -117,7 +135,9 @@ struct ActiveCallView: View {
                 call.callState = .connected
                 call.connectionInfo = connectionInfo
             case .ended:
-                closeCallView(webView)
+                DispatchQueue.main.async {
```
epoberezkin, line 138: The difference between this and `await MainActor.run` is that the latter would wait for the completion of the code inside - in most cases it is what you want, there are only rare cases when you want concurrent execution of code inside and after.

apps/ios/Shared/Views/Call/ActiveCallView.swift
```diff
@@ -106,6 +118,12 @@ struct ActiveCallView: View {
                         }
                     }
                 }
+                if state.connectionState == "closed" && m.activeCall != nil {
+                    DispatchQueue.main.async {
```
epoberezkin, line 122: await MainActor.run?

apps/ios/Shared/Views/Call/CallViewRenderers.swift
```diff
@@ -0,0 +1,77 @@
...
+    func updateUIView(_ view: UIView, context: Context) {
+        logger.debug("CallView.updateUIView remote")
+    }
+}
+
+struct CallViewLocal: UIViewRepresentable {
+    var client: WebRTCClient
+    var activeCall: Binding<WebRTCClient.Call?>
+    var localRendererAspectRatio: Binding<CGFloat?>
+    let localRenderer = RTCEAGLVideoView(frame: .zero)
```
epoberezkin, line 39: Purple warning:
-[UIView initWithFrame:] must be used from main thread only

apps/ios/Shared/Views/Call/ActiveCallView.swift
epoberezkin: also, on line 112 (call.callState = .connected):
/Users/evgeny/code/simplex-chat/simplex-chat/apps/ios/Shared/Views/Call/ActiveCallView.swift:112: warning run: Publishing changes from background threads is not allowed; make sure to publish values from the main thread (via operators like receive(on:)) on model updates.

apps/ios/Shared/Views/Call/ActiveCallView.swift
```diff
@@ -130,14 +150,16 @@ struct ActiveCallView: View {
                     call.localCamera = camera
                     Task {
                         // This disables microphone if it was disabled before flipping the camera
-                        await webView.setMicrophoneCaptureState(call.audioEnabled ? .active : .muted)
+                        client.setAudioEnabled(call.audioEnabled)
```
epoberezkin, line 153: on lines 148 (call.callState = .negotiated) and 150 (call.localCamera = camera):
/Users/evgeny/code/simplex-chat/simplex-chat/apps/ios/Shared/Views/Call/ActiveCallView.swift:148: warning run: Publishing changes from background threads is not allowed; make sure to publish values from the main thread (via operators like receive(on:)) on model updates.

apps/ios/Shared/MediaEncryption.playground/Contents.swift
```diff
@@ -0,0 +1,78 @@
...
+let ivTagBytes = 28
+var base: UnsafeMutableRawPointer = malloc(totalBytes)
+
+let assume = base.assumingMemoryBound(to: UInt8.self)
+assume[0] = 0 // key frame
+for i in 1..<totalBytes {
+    assume[i] = UInt8(i)
+}
+let unencrypted = NSData(bytesNoCopy: base, length: totalBytes)
+let aesKey = "PI-bV-FTgRqZM_lsDH9T21a0yRVMsvLFmvilJ9Ssk3g="
```
ltguillaume, line 16: Just wondering if it's okay to expose this.

epoberezkin: This is just a playground to test the code, it’s not used in the app.

PR #1935, opened 2023-02-16
core: encrypt/decrypt WebRTC frames

epoberezkin:


src/Simplex/Chat/Mobile/WebRTC.hs
```diff
@@ -0,0 +1,27 @@
+module Simplex.Chat.Mobile.WebRTC where
```
spaced4ndy, line 1: probably forgot to add to cabal file

PR #1963, opened 2023-03-01
Added translated docs to czech

zen0bit:


docs/ANDROID.md
```diff
@@ -1,3 +1,5 @@
+| Updated 26.02.2023 | Languages: EN, [CZ](/docs/lang/cs/ANDROID.md) |
```
epoberezkin, line 1: Please don't change the update dates if it's the only change - it'll make other translations look older than they actually are.

docs/lang/cs/ANDROID.md
```diff
@@ -0,0 +1,41 @@
+| Aktualizováno 26.02.2023 | Jazyky: CZ, [EN](/docs/ANDROID.md) |
```
epoberezkin, line 1: I suggest to make this date match the date of the original document that you translated.

epoberezkin:
Thanks a lot! I've made couple comments about the dates, but I can update myself of course.

docs/lang/cs/ANDROID.md
zen0bit: Wasn't in original file so...

07.02 from last commit on file is good?

zen0bit:
No should be as requested...

epoberezkin:
Thank you!

epoberezkin:
Merged, but looks like image URLs need to be amended to point to the original locations... If you are able to do it, please use relative links, so it works correctly when we get around to porting all these pages to the website.

PR #1971, opened 2023-03-03
core: types to support xftp

epoberezkin:


src/Simplex/Chat/Protocol.hs
```diff
@@ -316,7 +318,8 @@ instance ToJSON LinkPreview where
 data MsgContent
   = MCText Text
   | MCLink {text :: Text, preview :: LinkPreview}
-  | MCImage {text :: Text, image :: ImageData}
+  | MCImage {text :: Text, image :: ImageData, fileDescr :: Maybe Text}
+  | MCVideo {text :: Text, poster :: ImageData}
```
spaced4ndy, line 322: why it doesn't need fileDescr same as image?

epoberezkin: This fileDescr is specifically for low-resolution image in case a high-resolution image is passed as an attachment. I thought about supporting multiple attachments, and it seems like it is too complex in this case - we would have to support multiple files per message... Although the schema seems to allow multiple files per message, unless I'm mistaken, so maybe we should do it this way. But this is a case when we have a variant of the file - a low resolution option to complement a higher resolution option. Possibly we can allow the same for vides. This is different from multiple attachments.

epoberezkin: although I guess we can have lower resolution video as well, but I don't think it needs to happen in v1

epoberezkin: and maybe never

src/Simplex/Chat/Migrations/M20230304_file_description.hs
```diff
@@ -0,0 +1,13 @@
...
+
+module Simplex.Chat.Migrations.M20230304_file_description where
+
+import Database.SQLite.Simple (Query)
+import Database.SQLite.Simple.QQ (sql)
+
+m20230304_file_description :: Query
+m20230304_file_description =
+  [sql|
+ALTER TABLE rcv_files ADD COLUMN file_descr_rcv_file_id INTEGER NULL REFERENCES rcv_files (rcv_file_id) ON DELETE CASCADE;
```
spaced4ndy, line 11: not sure about ON DELETE CASCADE
```suggestion
ALTER TABLE rcv_files ADD COLUMN file_descr_rcv_file_id INTEGER NULL REFERENCES rcv_files (rcv_file_id);
```

src/Simplex/Chat/Migrations/M20230304_file_description.hs
```diff
@@ -12,11 +12,16 @@ m20230304_file_description :: Query
 m20230304_file_description =
   [sql|
 CREATE TABLE recipient_file_descriptions (
-  rcp_file_descr_id INTEGER PRIMARY KEY AUTOINCREMENT,
-  rcp_file_descr_text TEXT NOT NULL
+  file_descr_id INTEGER PRIMARY KEY AUTOINCREMENT,
```
spaced4ndy, line 15: why make it AUTOINCREMENT?

src/Simplex/Chat/Migrations/M20230304_file_description.hs
```diff
@@ -12,11 +12,16 @@ m20230304_file_description :: Query
...
+  file_descr_size INTEGER NOT NULL,
+  file_descr_status TEXT NOT NULL,
+  file_descr_text TEXT NOT NULL
 );
 
-ALTER TABLE rcv_files ADD COLUMN rcp_file_descr_id TEXT NULL;
+ALTER TABLE rcv_files ADD COLUMN file_descr_id TEXT NULL;
 
-ALTER TABLE snd_files ADD COLUMN rcp_file_descr_id TEXT NULL;
+ALTER TABLE snd_files ADD COLUMN file_descr_id TEXT NULL;
```
spaced4ndy, line 23: why are these TEXT? also should they be foreign keys?

src/Simplex/Chat/Types.hs
```diff
@@ -1470,8 +1471,10 @@ type FileTransferId = Int64
 data FileInvitation = FileInvitation
   { fileName :: String,
     fileSize :: Integer,
+    fileDigest :: Maybe FileDigest,
     fileConnReq :: Maybe ConnReqInvitation,
-    fileInline :: Maybe InlineFileMode
+    fileInline :: Maybe InlineFileMode,
+    fileDescrSize :: Maybe Integer -- if this property is set, it means that the sent file will be file description, not the actual file
```
spaced4ndy, line 1477: how will it be used? if to check size, there's non optional fileSize already; if just as a flag why not use Bool?

spaced4ndy:
when ready maybe it's better to merge to a dedicated `xftp` branch for now as in simplexmq, not master

src/Simplex/Chat/Types.hs
epoberezkin: fileSize includes the size of the file itself, but we need to receive fileDescription first that will arrive in chunks - this is its size, so we know how many chunks and where to truncate.

src/Simplex/Chat/Migrations/M20230304_file_description.hs
```diff
@@ -0,0 +1,29 @@
...
+  file_descr_size INTEGER NOT NULL,
+  file_descr_status TEXT NOT NULL,
+  file_descr_text TEXT NOT NULL
+);
+
+ALTER TABLE rcv_files ADD COLUMN file_descr_id INTEGER NULL
+  REFERENCES recipient_file_descriptions(file_descr_id) ON DELETE RESTRICT;
+
+ALTER TABLE snd_files ADD COLUMN file_descr_id INTEGER NULL
+  REFERENCES recipient_file_descriptions(file_descr_id) ON DELETE RESTRICT;
```
spaced4ndy, line 25: ON DELETE SET NULL?

+ add indexes

Issue #1610, opened 2022-12-20
Is it possible or in future (short-term) plans to increase the duration of voice notes?


tuqueque:
Hello.

First of all, this project has improved quite a lot! Congrats to the very competent dev team and I hope this project keeps improving as quickly as it has been so far! I've become very disappointed with the Session project. Their dev team is more focused in adding idiotic features, leaving already present ones full of bugs or with a sense of unfinished state.

Anyway, back to SimpleX (which, BTW, how should I pronounce it? "Simplex"?... "Simple Ex"?...)

I'm just getting familiar with this app, but me and all my contacts use voice notes A LOT, and it ends up being a deal breaker having such a limited time for voice notes. We need at least a 3-to-5-minute time limit for voice notes... Is it possible to increase the time limit? Personally, I don't care about the quality of the audio as long as the voice message is understandable.

As far as I know, the OPUS audio codec is currently a fantastic option, open source and present in pretty much all operanting systems. Maybe using OPUS instead of AAC and lowering the quality of the audio files could help increasing the time of voice notes without affecting the file size too much.

I really hope this can be done somehow, since this is untimately a deal breaker for me and my contacts.
Thank you in advance.

grm34:
Absolutely agree, 43s is really short.

Full support on this project 👌

epoberezkin:
We will be increasing it once XFTP support is rolled out (see the blog post about XFTP). The reason for the current limit is that voice messages must be 1-step async - that is, fully sent out without acceptance and can be fully received without the sender ever being online again (you know the problem with images and files, I assume). Old (current) files sub-protocol is multi-step async, so not good enough for larger voice files. XFTP on another hand is 1-step async for any size files, so increasing both the duration and quality of the voice messages won’t be a problem (and it’ll send much faster too).

epoberezkin:
About the ETA - April or early May

PR #1984, opened 2023-03-07
android: increases max duration of voice messages

grm34:
The current maximum duration of 43 seconds is really annoying (#1610 ). Increases from 6 to 36 chunks * 15780 bytes per chunk (43s to 4m18s).


epoberezkin:
this is not as simple I am afraid... Firstly, this limit is also imposed in the core library, not only in the UI. Secondly, the reason to constrain the size to 6 chunks is because these chunks are sent without recipient acceptance in the same connection where the messages are sent. Also they can be sent to group, where they need to be sent separately to each recipient. So if the size is increased that would clog the connection and normal messages won't be sendable until this larger file is fully sent. Sending them as normal files would require the sender to be online for them to be received (which is not the case for voice messages with the current approach), which is also bad UX, as voice messages, unlike files, are not attachments that can be received later, they are part of the conversation... This is a hard trade-off really.

The good news XFTP will solve it all and will allow sending large voice messages of higher quality: https://simplex.chat/blog/20230301-simplex-file-transfer-protocol.html

PR #1987, opened 2023-03-09
#italian-language-interface correction

zen0bit:


epoberezkin:
I think this is correct now, Italian was added in v4.5

PR #2022, opened 2023-03-16
v4.5.4: add support for observer role

epoberezkin:


apps/android/app/src/main/java/chat/simplex/app/model/ChatModel.kt
```diff
@@ -486,6 +486,24 @@ data class Chat (
...
+      else -> false
+    }
+
+  val userIsObserver: Boolean get() = when(chatInfo) {
+    is ChatInfo.Group -> {
+      val m = chatInfo.groupInfo.membership
+      m.memberActive && m.memberRole == GroupMemberRole.Observer
+    }
+    else -> false
+  }
```
spaced4ndy, line 505: could userCanSend in case of group be defined as !userIsObserver? these functions seem to almost mirror one another

apps/ios/Shared/Model/ChatModel.swift
```diff
@@ -545,6 +545,25 @@ final class Chat: ObservableObject, Identifiable {
...
+    }
+
+    var userIsObserver: Bool {
+        switch chatInfo {
+        case let .group(groupInfo):
+            let m = groupInfo.membership
+            return m.memberActive && m.memberRole == .observer
+        default: return false
+        }
+    }
```
spaced4ndy, line 565: same

spaced4ndy (dismissed):

apps/android/app/src/main/java/chat/simplex/app/model/ChatModel.kt
epoberezkin: not exactly

PR #2025, opened 2023-03-17
core, iOS: hidden and muted user profiles

epoberezkin:


src/Simplex/Chat/Migrations/M20230317_hidden_profiles.hs
```diff
@@ -0,0 +1,20 @@
...
+  [sql|
+ALTER TABLE users ADD COLUMN view_pwd_hash BLOB NULL;
+
+ALTER TABLE users ADD COLUMN view_pwd_salt BLOB NULL;
+
+ALTER TABLE users ADD COLUMN wipe_pwd_hash BLOB NULL;
+
+ALTER TABLE users ADD COLUMN wipe_pwd_salt BLOB NULL;
+
+ALTER TABLE users ADD COLUMN hide_ntfs INTEGER NULL DEFAULT 0;
```
spaced4ndy, line 19: why not make it not nullable with check and default and populate existing rows?

epoberezkin: ok

apps/ios/Shared/Model/NtfManager.swift
```diff
@@ -31,15 +31,16 @@ class NtfManager: NSObject, UNUserNotificationCenterDelegate, ObservableObject {
...
+                                withCompletionHandler handler: @escaping () -> Void) {
         logger.debug("NtfManager.userNotificationCenter: didReceive")
         let content = response.notification.request.content
         let chatModel = ChatModel.shared
         let action = response.actionIdentifier
         logger.debug("NtfManager.userNotificationCenter: didReceive: action \(action), categoryIdentifier \(content.categoryIdentifier)")
         if let userId = content.userInfo["userId"] as? Int64,
            userId != chatModel.currentUser?.userId {
-           changeActiveUser(userId)
+            // TODO check that user is not hidden, suppress notification if it is
```
spaced4ndy, line 42: this is done below, isn't it?
```suggestion
```

src/Simplex/Chat/Migrations/M20230317_hidden_profiles.hs
```diff
@@ -0,0 +1,19 @@
...
+import Database.SQLite.Simple.QQ (sql)
+
+m20230317_hidden_profiles :: Query
+m20230317_hidden_profiles =
+  [sql|
+PRAGMA ignore_check_constraints=ON;
+
+ALTER TABLE users ADD COLUMN view_pwd_hash BLOB NULL;
+ALTER TABLE users ADD COLUMN view_pwd_salt BLOB NULL;
+ALTER TABLE users ADD COLUMN show_ntfs INTEGER NOT NULL DEFAULT 1;
```
spaced4ndy, line 15: does it complete successfully on existing database? I don't remember exactly why we started doing CHECK and not just NOT NULL in such cases, maybe it was failing on some platforms/sqlite versions?

src/Simplex/Chat/View.hs
```diff
@@ -1210,9 +1228,15 @@ viewChatError logLevel = \case
     CENoConnectionUser agentConnId -> ["error: message user not found, conn id: " <> sShow agentConnId | logLevel <= CLLError]
     CEActiveUserExists -> ["error: active user already exists"]
     CEUserExists name -> ["user with the name " <> ttyContact name <> " already exists"]
+    CEUserUnknown -> ["user does not exist or incorrect password"]
     CEDifferentActiveUser commandUserId activeUserId -> ["error: different active user, command user id: " <> sShow commandUserId <> ", active user id: " <> sShow activeUserId]
     CECantDeleteActiveUser _ -> ["cannot delete active user"]
     CECantDeleteLastUser _ -> ["cannot delete last user"]
+    CECantHideLastUser _ -> ["cannot hide last visible user"]
```
spaced4ndy, line 1235: just got confused by the text, as if it's last active. ok to keep as is
```suggestion
    CECantHideOnlyUnhiddenUser _ -> ["cannot hide the only unhidden user"]
```

apps/ios/Shared/Views/UserSettings/UserProfilesView.swift
```diff
@@ -9,19 +9,32 @@ import SimpleXChat
...
+    @AppStorage(DEFAULT_SHOW_HIDDEN_PROFILES_NOTICE) private var showHiddenProfilesNotice = false
+    @AppStorage(DEFAULT_SHOW_MUTE_PROFILE_ALERT) private var showMuteProfileAlert = true
     @State private var showDeleteConfirmation = false
-    @State private var userToDelete: Int?
+    @State private var userToDelete: UserInfo?
     @State private var alert: UserProfilesAlert?
-    @State var authorized = !UserDefaults.standard.bool(forKey: DEFAULT_PERFORM_LA)
+    @State private var authorized = !UserDefaults.standard.bool(forKey: DEFAULT_PERFORM_LA)
+    @State private var searchTextOrPassword = ""
+    @State private var showProfilePassword = false
```
spaced4ndy, line 20: seems it's unused
```suggestion
```

apps/ios/Shared/Views/UserSettings/UserProfilesView.swift
```diff
@@ -41,45 +54,103 @@ struct UserProfilesView: View {
...
+                    message: m.users.count > 1
+                            ? Text("There should be at least one visible user profile.")
+                            : Text("There should be at least use user profile.")
+                )
+            case .hiddenProfilesNotice:
+                return Alert(
+                    title: Text("Make profile private!"),
+                    message: Text("You can hide or mute a user profile - swipe it to the right.\nSimpleX Lock must be enabled."),
+                    primaryButton: .default(Text("Don't show again")) {
+                        showHiddenProfilesNotice = false
```
spaced4ndy, line 141: maybe one time is enough? it's a bit irritating if I just click OK, which is bold. could be just a single button

apps/ios/Shared/Views/UserSettings/UserProfilesView.swift
```diff
@@ -41,45 +54,103 @@ struct UserProfilesView: View {
...
+                    },
+                    secondaryButton: .default(Text("Ok"))
+                )
+            case .muteProfileAlert:
+                return Alert(
+                    title: Text("Muted when inactive!"),
+                    message: Text("You would still receive calls and notifications from muted profiles when they are active."),
```
spaced4ndy, line 148: ok to to keep as is
```suggestion
                    message: Text("You will still receive calls and notifications from muted profiles when they are active."),
```

apps/ios/Shared/Views/UserSettings/UserProfilesView.swift
```diff
@@ -41,45 +54,103 @@ struct UserProfilesView: View {
...
-                Text("Your chat profiles are stored locally, only on your device.")
+                Text("Tap to activate profile.")
+                    .font(.body)
+                    .padding(.top, 8)
+
             }
         }
         .toolbar { EditButton() }
         .navigationTitle("Your chat profiles")
+        .searchable(text: $searchTextOrPassword, placement: .navigationBarDrawer(displayMode: .always))
```
spaced4ndy, line 98: a couple of strange search behaviours:
1. active hidden profile is always present in search results
2. unhiding profile from search results removes it from them (contrary to hiding)

epoberezkin: unhiding removes them because they no longer have password. I decided not to bother with that gimmick that keeps them, it's not difficult to find them. Will check 1.

src/Simplex/Chat/View.hs
epoberezkin: updated

apps/ios/Shared/Views/UserSettings/UserProfilesView.swift
epoberezkin: yes, but I thought it's important and people click OK without thinking. this way it'll at least will show again if they don't read it first time.

src/Simplex/Chat/Migrations/M20230317_hidden_profiles.hs
epoberezkin: will test on android, removed disabling checks/update

PR #2054, opened 2023-03-22
xftp: set app tmp directory

spaced4ndy:


src/Simplex/Chat.hs
```diff
@@ -4202,6 +4219,7 @@ chatCommandP =
       "/_app activate" $> APIActivateChat,
       "/_app suspend " *> (APISuspendChat <$> A.decimal),
       "/_resubscribe all" $> ResubscribeAllConnections,
+      "/_file_paths " *> (APISetFilePathsConfig <$> jsonP),
```
epoberezkin, line 4222:
```suggestion
      "/_set file paths " *> (APISetFilePathsConfig <$> jsonP),
      "/_file_paths " *> (APISetFilePathsConfig <$> filePathP),
```

src/Simplex/Chat.hs
```diff
@@ -355,6 +361,17 @@ processChatCommand = \case
...
+    let FilePathsConfig {appFilesFolder, appTempDirectory} = cfg
+    createDirectoryIfMissing True appFilesFolder
+    createDirectoryIfMissing True appTempDirectory
+    ff <- asks filesFolder
+    td <- asks tempDirectory
+    atomically $ do
+      writeTVar ff $ Just appFilesFolder
+      writeTVar td $ Just appTempDirectory
+    ok_
   SetFilesFolder ff -> do
```
epoberezkin, line 375: this can be implemented via APISetFilePaths

src/Simplex/Chat.hs
epoberezkin, line 4222: this is confusingly similar to the other one... Maybe better make /_set file paths

Or maybe it's better to just have two commands, not sure what's the point of combining it into one, and still keep the other one.

epoberezkin: In which case it would be just SetTempFolder (/_temp_folder)

tests/ChatTests/Files.hs
```diff
@@ -48,8 +48,10 @@ chatFileTests = do
       it "v2" testAsyncFileTransfer
       it "v1" testAsyncFileTransferV1
     xit "send and receive file to group, fully asynchronous" testAsyncGroupFileTransfer
-  describe "file transfer over XFTP" $ do
+  fdescribe "file transfer over XFTP" $ do
```
epoberezkin, line 51:
```suggestion
  describe "file transfer over XFTP" $ do
```

epoberezkin (approved):

PR #2063, opened 2023-03-22
docs: FR Update

ishi-sama:


epoberezkin:
Thank you!

Let's please use the dates on the documents to reflect the date of the English content, not when the file or some meta-data (like links to other documents) was changed.

So it would be a revision date, not change date.

epoberezkin:
Thank you!

Re "private by definition" instead of "private by design" - the latter is more conventional term.

PR #2069, opened 2023-03-24
android: hidden and muted user profiles

avently:


apps/android/app/src/main/java/chat/simplex/app/MainActivity.kt
```diff
@@ -395,14 +395,18 @@ fun MainPage(
 
 fun processNotificationIntent(intent: Intent?, chatModel: ChatModel) {
   val userId = getUserIdFromIntent(intent)
+  if (userId != null && chatModel.getUser(userId)?.showNotifications == false) {
+    // Inactive user with disabled notifications
+    return
+  }
```
epoberezkin, line 401: this probably should be processed

apps/android/app/src/main/java/chat/simplex/app/views/chatlist/UserPicker.kt
```diff
@@ -162,15 +169,17 @@ fun UserProfilePickerItem(u: User, unreadCount: Int = 0, onLongClick: () -> Unit
       )
     }
     if (u.activeUser) {
-      Icon(Icons.Filled.Done, null, Modifier.size(20.dp), tint = MaterialTheme.colors.onBackground)
+        Icon(Icons.Filled.Done, null, Modifier.size(20.dp), tint = MaterialTheme.colors.onBackground)
+    } else if (u.hidden) {
+        Icon(Icons.Filled.Lock, null, Modifier.size(20.dp), tint = HighOrLowlight)
```
epoberezkin, line 174:
```suggestion
```

this shouldn't show

avently: this code is reused with UserPicker and UserProfilesView. But since UserPicker filters users and skips hidden, this condition will never be called in picker. It will be called in UserProfilesView, so it's ok to have it here

epoberezkin: ok

PR #2091, opened 2023-03-28
ios: update developer options

epoberezkin:


apps/ios/Shared/Views/UserSettings/DeveloperView.swift
```diff
@@ -12,30 +12,52 @@ import SimpleXChat
...
+                    settingsRow("internaldrive") {
+                        Toggle("Confirm database upgrades", isOn: $confirmDatabaseUpgrades)
+                    }
                     settingsRow("chevron.left.forwardslash.chevron.right") {
                         Toggle("Show developer options", isOn: $developerTools)
                     }
-                    settingsRow("internaldrive") {
-                        Toggle("Confirm database upgrades", isOn: $confirmDatabaseUpgrades)
+                } footer: {
+                    (developerTools ? Text("Show: ") : Text("Hide: ")) + Text("Database IDs and Trasport isolation option.")
```
spaced4ndy, line 42:
```suggestion
                    (developerTools ? Text("Show: ") : Text("Hide: ")) + Text("Database IDs and Transport isolation option.")
```

spaced4ndy: maybe it's better for localization to concat these strings?
"Hide database IDs and Transport isolation option."
"Show database IDs and Transport isolation option."

PR #2102, opened 2023-03-29
android: video support

avently:


apps/android/app/src/main/java/chat/simplex/app/views/chat/ChatView.kt
```diff
@@ -963,16 +974,28 @@ private fun providerForGallery(
...
+          if (imageBitmap != null && filePath != null) {
+            val uri = FileProvider.getUriForFile(SimplexApp.context, "${BuildConfig.APPLICATION_ID}.provider", File(filePath))
+            ProviderMedia.Image(uri, imageBitmap)
+          } else null
+        }
+        is MsgContent.MCVideo -> {
+          val filePath = getLoadedFilePath(SimplexApp.context, item.file)
+          if (filePath != null) {
+            val uri = FileProvider.getUriForFile(SimplexApp.context, "${BuildConfig.APPLICATION_ID}.provider", File(filePath))
+            ProviderMedia.Video(uri, (item.content.msgContent as MsgContent.MCVideo).image)
```
epoberezkin, line 994: where does it get image from? with image it calls getLoadedImage, and here it doesn't - how does it work?

avently: https://github.com/simplex-chat/simplex-chat/pull/2102/files#diff-4a4ce64754443563623a9d8dd631e88d2f4c8ab682c3fff8db7c045cec08afa4R243

PR #2123, opened 2023-04-01
ios: UI to cancel receiving file

epoberezkin:


apps/ios/Shared/Views/Chat/ChatView.swift
```diff
@@ -586,14 +586,13 @@ struct ChatView: View {
 
         private func cancelFileUIAction(_ fileId: Int64) -> UIAction {
             UIAction(
-                title: NSLocalizedString("Cancel", comment: "chat item action"),
-                image: UIImage(systemName: "xmark"),
-                attributes: [.destructive]
+                title: NSLocalizedString("Stop file", comment: "chat item action"),
```
spaced4ndy, line 589: I haven't called it Stop because you can cancel before transfer is started, there's nothing to stop yet.

spaced4ndy: At least receive is not showing this button in RcvInvitation status, only Delete. Maybe exclude SndStored as well? Then at least it makes more sense as it's only for "in progress" statuses.

epoberezkin: > I haven't called it Stop because you can cancel before transfer is started, there's nothing to stop yet.

"Stop" also means "prevent something from happening", not necessarily "terminate" something that is already happening": https://dictionary.cambridge.org/dictionary/english/stop. So stop is clear enough in this context, and also shorter.

> Then at least it makes more sense as it's only for "in progress" statuses.

It's ok to use "stop" for something due to start.

epoberezkin: "Cancel" was very unclear. "Cancel file" is an option, but it is less clear what it means, I think

spaced4ndy (approved):

apps/ios/Shared/Views/Chat/ChatView.swift
```diff
@@ -587,14 +587,13 @@ struct ChatView: View {
...
+                image: UIImage(systemName: "xmark")
             ) { _ in
                 AlertManager.shared.showAlert(Alert(
-                    title: Text("Cancel file transfer?"),
-                    message: Text("File transfer will be cancelled. If it's in progress it will be stoppped."),
-                    primaryButton: .destructive(Text("Confirm")) {
+                    title: Text("Stop receiving file?"),
+                    message: Text("File reception will be stopped, and the received part removed."),
```
spaced4ndy, line 595:
```suggestion
                    title: Text("Stop file transfer?"),
                    message: Text("File transfer will be stopped."),
```

spaced4ndy (approved):

PR #2133, opened 2023-04-03
mobile: delete XFTP files after uploading

spaced4ndy:


apps/android/app/src/main/java/chat/simplex/app/model/SimpleXAPI.kt
```diff
@@ -1440,8 +1440,18 @@ open class ChatController(var ctrl: ChatCtrl?, val ntfManager: NtfManager, val a
...
       is CR.SndFileProgressXFTP ->
         chatItemSimpleUpdate(r.user, r.chatItem)
-      is CR.SndFileCompleteXFTP ->
+      is CR.SndFileCompleteXFTP -> {
         chatItemSimpleUpdate(r.user, r.chatItem)
+        val cItem = r.chatItem.chatItem
+        val mc = cItem.content.msgContent
+        val fileName = cItem.file?.fileName
+        if (
+          mc is MsgContent.MCFile
```
epoberezkin, line 1449: I though we wanted to only constrain it for images, e.g. large videos can also be removed. Or you can't remove videos as you won't be able to play them?

Probably it needs some separate action/setting...

spaced4ndy: that's why I didn't exclude videos, yes, but it can be considered

Issue #2132, opened 2023-04-02
[bug]: android: sending direct message in the group when the first message is an image


epoberezkin:
In a group chat if I try to "Send direct message" to a somebody I already have in my contacts list it only shows the last message. If I do the same for somebody I don't have in my contacts list and (without sending a message first) choose a photo to send, it goes back to the group chat and the app thinks I want to send a photo there.

PR #2139, opened 2023-04-03
android: open direct chat simplified

avently:


epoberezkin:
partially fixes #2132

PR #2140, opened 2023-04-03
core: update servers API to include XFTP servers, ios: generalize UI to manage servers

epoberezkin:


src/Simplex/Chat.hs
```diff
@@ -4413,17 +4426,17 @@ chatCommandP =
...
+      "/smp test " *> (TestProtoServer . AProtoServerWithAuth SPSMP <$> strP),
+      "/xftp test " *> (TestProtoServer . AProtoServerWithAuth SPXFTP <$> strP),
+      "/_servers " *> (APISetUserProtoServers <$> A.decimal <* A.space <*> srvCfgP),
+      "/smp " *> (SetUserProtoServers . APSC SPSMP . ProtoServersConfig . map toServerCfg <$> protocolServersP),
+      "/smp default" $> SetUserProtoServers (APSC SPSMP $ ProtoServersConfig []),
+      "/xftp " *> (SetUserProtoServers . APSC SPXFTP . ProtoServersConfig . map toServerCfg <$> protocolServersP),
+      "/xftp default" $> SetUserProtoServers (APSC SPXFTP $ ProtoServersConfig []),
+      "/_servers " *> (APIGetUserProtoServers <$> A.decimal <* A.space <*> strP),
+      "/smp" $> GetUserProtoServers (AProtocolType SPSMP),
+      "/xftp" $> GetUserProtoServers (AProtocolType SPXFTP),
```
spaced4ndy, line 4439: /xftp is already used for APISetXFTPConfig

apps/ios/Shared/Views/UserSettings/ProtocolServersView.swift
```diff
@@ -119,11 +124,22 @@ struct SMPServersView: View {
                 )
             }
         }
+        .onAppear {
```
spaced4ndy, line 127: didn't we have issues with loading data in onAppear? I think they happened sporadically / were hard to reproduce.. and maybe there was a better fix than to replace onAppear with a parameter so it's ok..

src/Simplex/Chat.hs
epoberezkin: with parameters, so should be ok

spaced4ndy (approved):

PR #2153, opened 2023-04-07
core: allow messages with contact requests

epoberezkin:


ahadley1124:
What will happen to the messages sent the contact if the recipient denies the contact. Will these messages be stored in the database or will they be deleted from both parties?

tomquas:
hey, any chance this going to happen soon?

PR #2170, opened 2023-04-11
core: request and save extra recipient file descriptions

spaced4ndy:


src/Simplex/Chat.hs
```diff
@@ -604,6 +605,8 @@ processChatCommand = \case
                   withStore' $ \db -> createSndFTDescrXFTP db user (Just m) conn ft fileDescr
               saveMemberFD _ = pure ()
         pure (fInv, ciFile, ft)
+      roundToPowerOfTwo :: Int -> Int
+      roundToPowerOfTwo n = fromIntegral $ (2 :: Integer) ^ (ceiling (logBase 2 (fromIntegral n) :: Double) :: Integer)
```
epoberezkin, line 609:
```suggestion
      roundedFDCount :: Int -> Int
      roundedFDCount n = fromIntegral $ (2 :: Integer) ^ (ceiling (logBase 2 (fromIntegral n) :: Double) :: Integer)
```

PR #2192, opened 2023-04-17
core: change default for Disappearing Messages to "allow", mobile: support disabling without prohibiting

epoberezkin:


tests/ChatTests/Direct.hs
```diff
@@ -1530,14 +1530,11 @@ testUsersTimedMessages tmp = do
...
+      alice <## ("Disappearing messages: enabled (you allow: yes (" <> ttl <> " sec), contact allows: yes)")
       bob <## (aliceName <> " updated preferences for you:")
-      bob <## ("Disappearing messages: off (you allow: no, contact allows: yes (" <> ttl <> " sec))")
-      bob ##> ("/set disappear @" <> aliceName <> " yes")
-      bob <## ("you updated preferences for " <> aliceName <> ":")
       bob <## ("Disappearing messages: enabled (you allow: yes (" <> ttl <> " sec), contact allows: yes (" <> ttl <> " sec))")
-      alice <## "bob updated preferences for you:"
-      alice <## ("Disappearing messages: enabled (you allow: yes (" <> ttl <> " sec), contact allows: yes (" <> ttl <> " sec))")
+      bob ##> ("/set disappear @" <> aliceName <> " yes")
+      bob <## ("your preferences for " <> aliceName <> " did not change")
```
spaced4ndy, line 1537: try to not set disappearing messages as bob and test that they would be disappearing as alice enabled
```suggestion
```

PR #2193, opened 2023-04-17
core, mobile: file error statuses, cancel sent file

spaced4ndy:


src/Simplex/Chat.hs
```diff
@@ -1395,9 +1395,14 @@ processChatCommand = \case
...
     withChatLock "cancelFile" . procCmd $
       withStore (\db -> getFileTransfer db user fileId) >>= \case
-        FTSnd ftm@FileTransferMeta {cancelled} fts
+        FTSnd ftm@FileTransferMeta {xftpSndFile, cancelled} fts
           | cancelled -> throwChatError $ CEFileCancel fileId "file already cancelled"
-          | not (null fts) && all (\SndFileTransfer {fileStatus = s} -> s == FSComplete || s == FSCancelled) fts ->
+          | not (null fts)
+              && all
+                ( \SndFileTransfer {fileStatus = s} ->
+                    s == FSCancelled || (s == FSComplete && isNothing xftpSndFile)
```
epoberezkin, line 1403: better give it a name as a function...

apps/ios/SimpleXChat/ChatTypes.swift
```diff
@@ -2243,32 +2243,62 @@ public struct CIFile: Decodable {
...
             }
         }
     }
 }
 
+private var sndCancelTexts = (
+  NSLocalizedString("Stop file", comment: "cancel file action"),
+  NSLocalizedString("Stop sending file?", comment: "cancel file question"),
+  NSLocalizedString("Sending file will be stopped.", comment: "cancel file text"),
+  NSLocalizedString("Stop", comment: "cancel file confirmation")
```
epoberezkin, line 2285: these can be struct, and also LocalizedStringKey - there is no need to use NSLocalizedString here I think? At least for the last three, maybe the first needs to be NSLocalizedString, not sure.

PR #2233, opened 2023-04-24
website: add atom feeds for blogs

MSarmadQadeer:


website/src/blogs-feed.njk
```diff
@@ -0,0 +1,41 @@
+---json
```
epoberezkin, line 1: what's wrong with yaml?

MSarmadQadeer: It's a templating file. It will generate the Yaml. That JSON is just to hold some values that are used below.

epoberezkin: I understand. It seems like the default for front matter in njk is yaml, not JSON, template will generate xml

PR #2252, opened 2023-04-25
Create simplex-chat.md

lastimeoceanisfall:
translated to polish


epoberezkin:
we won't be translating that, as we would have to maintain it - it can't be just out of date, unlike other docs

PR #2270, opened 2023-04-25
Create Options.hs

lastimeoceanisfall:
translated to polish


epoberezkin:
we won't be translating this code, sorry.

PR #2273, opened 2023-04-25
Create install.sh

lastimeoceanisfall:
translated to polish


epoberezkin:
doesn't need translation

PR #2304, opened 2023-04-25
Create README.md

lastimeoceanisfall:
translated to polish


epoberezkin:
Thanks for the translations!

This is a bit of a disaster to have 50 different PRs though. please don't add individual files in the future. It probably was via web UI... In which case we should have created a separate branch and added them there directly without PRs.

epoberezkin:
I've added a branch docs-polish - will be merging the docs there.

PR #2327, opened 2023-04-25
shellcheck all the things

doomguy:
- Tried to make shellcheck happy
- Hardened the files with "bash strict mode" header
- Some minor code & performance improvements
- UNTESTED - There might be dragons..


shumvgolove:
Hey, thanks for contribution! Please don't hardcore `build-android.sh` to `bash`, I've specifically made this script POSIX complaint. And we also can get rid of `build-android-bundle.sh`, because this is no longer needed.

PR #2333, opened 2023-04-26
core: get all chat items API

epoberezkin:


tests/ChatTests/Direct.hs
```diff
@@ -1544,7 +1544,9 @@ testUserPrivacy =
       showActiveUser alice "alisa"
       -- connect using second user
       connectUsers alice bob
+      threadDelay 1000000
       alice #> "@bob hello"
+      threadDelay 1000000
```
spaced4ndy, line 1549: is second delay necessary?

tests/ChatTests/Direct.hs
```diff
@@ -1576,12 +1578,19 @@ testUserPrivacy =
       alice <## "alisa (active, hidden, muted)"
       -- hidden message is saved
       alice ##> "/tail"
+      alice <##? chatHistory
+      alice ##> "/_get items count=10"
+      alice <##? chatHistory
+      alice ##> "/_get items before=9 count=10"
```
spaced4ndy, line 1584: I think it'd be better to test in a separate test where there're both direct and group chats and not in this one, but it's ok

spaced4ndy (approved):

tests/ChatTests/Direct.hs
epoberezkin: yes, because the time of received messages is rounded down and the order is not deterministic when sorted by time.

PR #2349, opened 2023-04-29
core: prevent failure loading chat on invalid item content JSON

epoberezkin:


src/Simplex/Chat/Messages.hs
```diff
@@ -764,7 +768,9 @@ instance FromJSON MsgDecryptError where
 
 ciRequiresAttention :: forall d. MsgDirectionI d => CIContent d -> Bool
 ciRequiresAttention content = case msgDirection @d of
-  SMDSnd -> True
+  SMDSnd -> case content of
+    CIInvalidJSON _ -> False
+    _ -> True
```
spaced4ndy, line 773: Why make a differentiation here? It only affects chat timestamp in backend for snd items, and no chat items are going to be created in invalid json type, so it wouldn't even change that (and if they are created it's probably still correct to update chat ts)

```suggestion
  SMDSnd -> True
```

src/Simplex/Chat/Messages.hs
```diff
@@ -790,6 +796,7 @@ ciRequiresAttention content = case msgDirection @d of
     CIRcvChatFeatureRejected _ -> True
     CIRcvGroupFeatureRejected _ -> True
     CIRcvModerated -> True
+    _ -> False
```
spaced4ndy, line 799: I think it's better to keep this case complete to remember to make a decision for new chat item types, instead of guessing at unexpected effects later

```suggestion
    CIInvalidJSON _ -> False
```

spaced4ndy (approved):

PR #2363, opened 2023-05-02
android: contact address UX

avently:


apps/android/app/src/main/res/values/strings.xml
```diff
@@ -103,6 +103,7 @@
   <string name="error_smp_test_server_auth">Server requires authorization to create queues, check password</string>
   <string name="error_xftp_test_server_auth">Server requires authorization to upload, check password</string>
   <string name="error_smp_test_certificate">Possibly, certificate fingerprint in server address is incorrect</string>
+  <string name="error_setting_address">Error settings address</string>
```
spaced4ndy, line 106:
```suggestion
  <string name="error_setting_address">Error setting address</string>
```

apps/android/app/src/main/java/chat/simplex/app/views/newchat/ContactConnectionInfoView.kt
```diff
@@ -43,21 +45,24 @@ fun ContactConnectionInfoView(
       }
     }
   }
+  val context = LocalContext.current
   ContactConnectionInfoLayout(
     connReq = connReqInvitation,
     contactConnection,
+    connIncognito = chatModel.incognito.value,
```
spaced4ndy, line 52: on pending connection view (opened from chat list) incorrect shared profile text is shown - it depends on incognito mode instead of connection state

apps/android/app/src/main/java/chat/simplex/app/views/usersettings/UserAddressView.kt
```diff
@@ -1,118 +1,398 @@
...
+    Box(
+      Modifier.fillMaxSize(),
+      contentAlignment = Alignment.Center
+    ) {
+      CircularProgressIndicator(
+        Modifier
+          .padding(horizontal = 2.dp)
+          .size(30.dp),
+        color = MaterialTheme.colors.secondary,
+        strokeWidth = 3.dp
```
spaced4ndy, line 147: better to add semitransparent white circle underneath when userAddress != null - currently it's hard to see on top of qr code

avently: added circle with background color with alpha 0.9f

PR #2378, opened 2023-05-04
ios: progress indicator on migrations w/t confirmation

spaced4ndy:


apps/ios/Shared/ContentView.swift
```diff
@@ -104,6 +106,14 @@ struct ContentView: View {
         Button(action: runAuthenticate) { Label("Unlock", systemImage: "lock") }
     }
 
+    private func initializationView() -> some View {
+        VStack {
+            ProgressView().scaleEffect(2)
+            Text("Initialization…")
```
epoberezkin, line 112:
```suggestion
            Text("Preparing database…")
```

apps/ios/Shared/ContentView.swift
```diff
@@ -69,6 +73,8 @@ struct ContentView: View {
     @ViewBuilder private func contentView() -> some View {
         if prefPerformLA && userAuthorized != true {
             lockButton()
+        } else if chatModel.chatDbStatus == nil {
```
epoberezkin, line 76:
```suggestion
        } else if chatModel.chatDbStatus == nil && showInitializationView {
```

apps/ios/Shared/ContentView.swift
```diff
@@ -104,6 +110,18 @@ struct ContentView: View {
...
+        if showInitializationView {
+            VStack {
+                ProgressView().scaleEffect(2)
+                Text("Preparing database…")
+                    .padding()
+            }
+        } else {
+            EmptyView()
+        }
+    }
```
epoberezkin, line 123:
```suggestion
    private func initializationView() -> some View {
        VStack {
            ProgressView().scaleEffect(2)
            Text("Preparing database…")
                .padding()
        }
    }
```

PR #2409, opened 2023-05-08
desktop: RFC

avently:


docs/rfcs/2023-04-28-multiplatform.md
```diff
@@ -0,0 +1,27 @@
...
+- make the project to build successfully without using hardware features in desktop
+- implement video support like [here](https://github.com/JetBrains/compose-multiplatform/blob/master/experimental/components/VideoPlayer/demo/src/jvmMain/kotlin/org/jetbrains/compose/videoplayer/demo/Main.kt). Under the hood it uses [vlcj](https://github.com/caprica/vlcj) library which is using VLC native library
+- implement audio support using the same `vlcj` library
+- implement camera support using [webcam-capture](https://github.com/sarxos/webcam-capture/tree/master/webcam-capture-drivers/driver-vlcj) that is using `vlcj` too
+- since we have a working camera on this stage, QR Code scanner will be implemented too by the lib that is already used on Android
+- rewrite the whole WebRTC stack for using Google's WebRTC lib with a help of [webrtc-java](https://github.com/devopvoid/webrtc-java) helper but with [our own build](https://github.com/simplex-chat/WebRTC) of WebRTC lib since it has our changes to encryption/decryption possibilities. This has to be used on Android as well
+- ensure the app can be run on all platforms with JDK
+- prepare distribution of binaries per platform.
```
epoberezkin, line 26: This needs to be done much earlier, as soon as something is working, as it may expose additional problems, and it won't be usable without it - after "make the project to build successfully without using hardware features in desktop".

jittygitty: @avently @epoberezkin I'm really rooting for you guys, so please do not make the same mistake https://getsession.org/ did, please make the Desktop apps easily "portable" so you can move all your profile and data to another device. (Since I think it would be very difficult to "sync" data across different device installs in this local only setup with servers acting as just "dumb relays" etc.) Keep all in the app folder etc. Also, not sure webrtc-java is the best route...

Here's just one of the reasons I had recommended considering Flutter:
https://flutter-webrtc.org/docs/docs/
https://pub.dev/packages/flutter_webrtc
You might consider leveraging livekit: https://github.com/livekit/livekit
Example video chat: https://element.io/blog/element-call-beta-3/
(feel free to move comment elsewhere if wrong place for it, thanks!)

PR #2442, opened 2023-05-15
ios: reactions UI

epoberezkin:


apps/ios/Shared/Views/Chat/ChatView.swift
```diff
@@ -473,8 +474,8 @@ struct ChatView: View {
             VStack(alignment: .trailing, spacing: 4) {
                 ChatItemView(chatInfo: chat.chatInfo, chatItem: ci, showMember: showMember, maxWidth: maxWidth, scrollProxy: scrollProxy, revealed: $revealed, allowMenu: $allowMenu, audioPlayer: $audioPlayer, playbackState: $playbackState, playbackTime: $playbackTime)
                     .uiKitContextMenu(menu: uiMenu, allowMenu: $allowMenu)
-                if ci.reactions.count > 0 {
-                    chatItemReactions(ci.reactions)
+                if ci.content.msgContent != nil && ci.meta.itemDeleted == nil && ci.reactions.count > 0 {
```
spaced4ndy, line 477: for received message - if the message is short and reactions row is wide, message bubble will be pushed to center

apps/ios/Shared/Views/Chat/ChatView.swift
```diff
@@ -555,6 +566,10 @@ struct ChatView: View {
                    let cancelAction = file.cancelAction  {
                     menu.append(cancelFileUIAction(file.fileId, cancelAction))
                 }
+                if chat.chatInfo.featureEnabled(.reactions) && ci.allowAddReaction && developerTools,
+                   let rm = reactionUIMenu() {
+                    menu.append(rm)
+                }
```
spaced4ndy, line 572: make it first in the list? it's closer to the initial touch and to the "reply"

apps/ios/Shared/Views/Chat/ChatView.swift
```diff
@@ -586,7 +601,42 @@ struct ChatView: View {
                 }
             }
         }
-        
+
+        private func reactionUIMenu() -> UIMenu? {
+            let rs = MsgReaction.values.compactMap { r in
+                ci.reactions.contains(where: { $0.reaction == r })
+                ? nil
+                : UIAction(title: r.text) { _ in setReaction(r, add: true) }
```
spaced4ndy, line 609: it's counterintuitive that reaction is not in the list when it's on chat item and the user didn't react with it. ok to remove from the list if the user used it

spaced4ndy: maybe don't remove them from this list at all, and instead disable reactions made by user?

apps/ios/SimpleXChat/ChatTypes.swift
```diff
@@ -560,6 +598,12 @@ public enum ChatFeature: String, Decodable, Feature {
             case .yes: return "Allow irreversible message deletion only if your contact allows it to you."
             case .no: return "Contacts can mark messages for deletion; you will be able to view them."
             }
+        case .reactions:
+            switch allowed {
+            case .always: return "Allow your contacts reacting to messages."
```
spaced4ndy, line 603:
```suggestion
            case .always: return "Allow message reactions to your contacts."
```

apps/ios/SimpleXChat/ChatTypes.swift
```diff
@@ -593,6 +637,14 @@ public enum ChatFeature: String, Decodable, Feature {
...
                     ? "Only your contact can irreversibly delete messages (you can mark them for deletion)."
                     : "Irreversible message deletion is prohibited in this chat."
+        case .reactions:
+            return enabled.forUser && enabled.forContact
+                    ? "Both you and your contact can react to messages."
+                    : enabled.forUser
+                    ? "Only you can react to messages."
+                    : enabled.forContact
+                    ? "Only your contact can react to messages."
+                    : "Message reactions are prohibited in this chat."
```
spaced4ndy, line 647:
```suggestion
                    ? "Both you and your contact can make message reactions."
                    : enabled.forUser
                    ? "Only you can make message reactions."
                    : enabled.forContact
                    ? "Only your contact can make message reactions."
                    : "Message reactions are prohibited in this chat."
```

apps/ios/SimpleXChat/ChatTypes.swift
```diff
@@ -705,6 +764,11 @@ public enum GroupFeature: String, Decodable, Feature {
                 case .on: return "Group members can irreversibly delete sent messages."
                 case .off: return "Irreversible message deletion is prohibited in this group."
                 }
+            case .reactions:
+                switch enabled {
+                case .on: return "Group members can react to messages."
```
spaced4ndy, line 769:
```suggestion
                case .on: return "Group members can make message reactions."
```

apps/ios/Shared/Views/Chat/ChatView.swift
epoberezkin: if it's narrower - should it then be left-aligned on received?

spaced4ndy: I think yes

apps/ios/Shared/Views/Chat/ChatView.swift
spaced4ndy: yes, didn't think think about it, but still closer to similar actions

PR #2452, opened 2023-05-17
ios: remove "off" from group preferences ttl picker; core: allow optional ttl for group preference

spaced4ndy:


src/Simplex/Chat/Messages.hs
```diff
@@ -376,7 +376,7 @@ contactTimedTTL Contact {mergedPreferences = ContactUserPreferences {timedMessag
 
 groupTimedTTL :: GroupInfo -> Maybe (Maybe Int)
 groupTimedTTL GroupInfo {fullGroupPreferences = FullGroupPreferences {timedMessages = TimedMessagesGroupPreference {enable, ttl}}}
-  | enable == FEOn = Just $ Just ttl
```
epoberezkin, line 379: not sure about this - we want to only prevent it in UI, why preventing it here?

PR #2454, opened 2023-05-18
android: ui for message details

epoberezkin:


apps/android/app/src/main/java/chat/simplex/app/model/SimpleXAPI.kt
```diff
@@ -595,6 +595,18 @@ open class ChatController(var ctrl: ChatCtrl?, val ntfManager: NtfManager, val a
     }
   }
 
+  suspend fun apiGetChatItemInfo(type: ChatType, id: Long, itemId: Long): ChatItemInfo? {
+    return when (val r = sendCmd(CC.ApiGetChatItemInfo(type, id, itemId))) {
+      is CR.ApiChatItemInfo -> r.chatItemInfo
+      else -> {
+        if (!(networkErrorAlert(r))) {
```
spaced4ndy, line 602: how can it be a network error in this api

epoberezkin: :)

apps/android/app/src/main/java/chat/simplex/app/views/chat/ChatItemInfoView.kt
```diff
@@ -80,23 +80,27 @@ fun ChatItemInfoView(ci: ChatItem, ciInfo: ChatItemInfo, devTools: Boolean) {
       if (!sent) {
         InfoRow(stringResource(R.string.info_row_received_at), localTimestamp(ci.meta.createdAt))
       }
-      when (ci.meta.itemDeleted) {
+      when (val itemDeleted = ci.meta.itemDeleted) {
```
epoberezkin, line 89: maybe get timestamp outside with some CIDeleted.itemDeletedTs not to have two ifs and check it's not null?

apps/android/app/src/main/java/chat/simplex/app/views/chat/ChatItemInfoView.kt
```diff
@@ -129,22 +123,26 @@ fun itemInfoShareText(ci: ChatItem, chatItemInfo: ChatItemInfo, devTools: Boolea
   if (!ci.chatDir.sent) {
     shareText.add(String.format(generalGetString(R.string.share_text_received_at), localTimestamp(meta.createdAt)))
   }
-  when (ci.meta.itemDeleted) {
+  when (val itemDeleted = ci.meta.itemDeleted) {
```
epoberezkin, line 132: same comment here

PR #2468, opened 2023-05-20
mobile: hide share address until v5.2

epoberezkin:


spaced4ndy (commented):
Also profile_address cli commands and 2 places in help - address help and on address creation

epoberezkin:
probably enough to hide in help/output

PR #2487, opened 2023-05-23
mobile: remove file on apiSendMessage error

spaced4ndy:


apps/android/app/src/main/java/chat/simplex/app/views/chat/ComposeView.kt
```diff
@@ -376,8 +376,13 @@ fun ComposeView(
...
       ttl = ttl
     )
-    if (aChatItem != null) chatModel.addChatItem(cInfo, aChatItem.chatItem)
-    return aChatItem?.chatItem
+    if (aChatItem != null) {
+      chatModel.addChatItem(cInfo, aChatItem.chatItem)
+      return aChatItem.chatItem
+    } else if (file != null) {
+      removeFile(context, file)
+    }
```
epoberezkin, line 384:
```suggestion
    }
    if (file != null) removeFile(context, file)
```

epoberezkin: usually better not to use else after return

PR #2520, opened 2023-05-26
Multiplatform support

avently:


apps/android/common/src/commonMain/resources/distribute/SimpleX.desktop
```diff
@@ -0,0 +1,7 @@
+[Desktop Entry]
+Type=Application
+Name=SimpleX
+Comment=Messenger for the future
```
epoberezkin, line 4: where does it show? needs to be changed.

apps/android/common/src/commonMain/resources/MR/el/strings.xml
```diff
@@ -42,7 +42,6 @@
     <string name="accept_connection_request__question">Αποδοχή αιτήματος σύνδεσης;</string>
     <string name="callstatus_accepted">αποδεκτή κλήση</string>
     <string name="network_enable_socks_info">Πρόσβαση στους διακομιστές μέσω SOCKS proxy στην πόρτα 9050; Ο διακομιστής μεσολάβησης (proxy server) πρέπει να είναι ενεργός πριν ενεργοποιηθεί αυτή η ρύθμιση.</string>
-    <string name="accept_requests">Αποδοχή αιτημάτων</string>
```
epoberezkin, line 45: string removed

apps/android/common/src/commonMain/resources/MR/base/strings.xml
```diff
@@ -678,12 +678,14 @@
   <!-- markdown demo - MarkdownHelpView.kt -->
   <string name="how_to_use_markdown">How to use markdown</string>
   <string name="you_can_use_markdown_to_format_messages__prompt">You can use markdown to format messages:</string>
-  <string name="bold">bold</string>
-  <string name="italic">italic</string>
+  <string name="bold_font">bold</string>
```
epoberezkin, line 681: not all language files are renamed, same about the next one

apps/android/common/src/commonMain/kotlin/chat/simplex/common/views/usersettings/UserProfileView.kt
```diff
@@ -199,29 +198,28 @@ fun EditImageButton(click: () -> Unit) {
 fun DeleteImageButton(click: () -> Unit) {
   IconButton(onClick = click) {
     Icon(
-      painterResource(R.drawable.ic_close),
-      contentDescription = stringResource(R.string.delete_image),
+      painterResource(MR.images.ic_close),
```
epoberezkin, line 201: was it necessary to rename R to MR?

apps/android/common/src/commonMain/resources/MR/el/strings.xml
avently: There is not such string key in base locale. so it should be removed. Unused

apps/android/common/src/commonMain/resources/MR/base/strings.xml
avently: true

apps/android/common/src/commonMain/kotlin/chat/simplex/common/views/usersettings/UserProfileView.kt
avently: renamed back just for review. And yes, it's necessary for the lib we use since it's not an Android plugin anymore

apps/android/common/src/commonMain/resources/distribute/SimpleX.desktop
aicynide: Linux desktop environments like kde,xfce,gnome

avently:
Closing since the code was divided to pieces and merged into master

PR #2552, opened 2023-06-06
core, mobile: fix time format

spaced4ndy:


spaced4ndy:
It remains to repeat in android

PR #2560, opened 2023-06-08
core: optimize group deletion (delays deletion of unused contacts)

spaced4ndy:


src/Simplex/Chat.hs
```diff
@@ -122,6 +122,7 @@ defaultChatConfig =
       testView = False,
       initialCleanupManagerDelay = 30 * 1000000, -- 30 seconds
       cleanupManagerInterval = 30 * 60, -- 30 minutes
+      cleanupManagerStepDelay = 3 * 1000000, -- 3 seconds
```
epoberezkin, line 125: why 3 seconds?

src/Simplex/Chat.hs
```diff
@@ -837,21 +838,22 @@ processChatCommand = \case
...
       pure $ CRContactConnectionDeleted user conn
     CTGroup -> do
-      Group gInfo@GroupInfo {membership} members <- withStore $ \db -> getGroup db user chatId
+      Group gInfo@GroupInfo {membership} members <-
+        withStoreCtx (Just "APIDeleteChat CTGroup, getGroup") $ \db -> getGroup db user chatId
       let isOwner = memberRole (membership :: GroupMember) == GROwner
           canDelete = isOwner || not (memberCurrent membership)
       unless canDelete $ throwChatError $ CEGroupUserRole gInfo GROwner
-      filesInfo <- withStore' $ \db -> getGroupFileInfo db user gInfo
+      filesInfo <- withStoreCtx' (Just "APIDeleteChat CTGroup, getGroupFileInfo") $ \db -> getGroupFileInfo db user gInfo
```
epoberezkin, line 846: why do we need to use withStoreCtx everywhere?

src/Simplex/Chat/Migrations/M20230608_contacts_marked_for_deletion.hs
```diff
@@ -0,0 +1,22 @@
...
+
+CREATE INDEX msg_delivery_events_msg_delivery_id ON msg_delivery_events(msg_delivery_id);
+|]
+
+down_m20230608_contacts_marked_for_deletion :: Query
+down_m20230608_contacts_marked_for_deletion =
+  [sql|
+DROP INDEX msg_delivery_events_msg_delivery_id;
+
+ALTER TABLE contacts DROP COLUMN marked_for_deletion;
```
epoberezkin, line 21: I think "deleted" is a conventional name here

src/Simplex/Chat/Types.hs
```diff
@@ -176,7 +176,8 @@ data Contact = Contact
     mergedPreferences :: ContactUserPreferences,
     createdAt :: UTCTime,
     updatedAt :: UTCTime,
-    chatTs :: Maybe UTCTime
+    chatTs :: Maybe UTCTime,
+    markedForDeletion :: Bool
```
epoberezkin, line 180: I don't see this field ever used - why it is added?

epoberezkin (commented):
"deleted" can be removed from the type, it will reduce the noice.

Also, so queries in Contacts table don't check "deleted" column.

src/Simplex/Chat.hs
spaced4ndy: arbitrary, to give other queries period when cleanup manager queries don't occupy database

src/Simplex/Chat.hs
spaced4ndy: it's not necessary

PR #2592, opened 2023-06-18
ios: UI to filter favorite and unread chats

epoberezkin:


apps/ios/Shared/Views/ChatList/ChatListView.swift
```diff
@@ -106,6 +112,30 @@ struct ChatListView: View {
...
+    private func toggleShowFavouritesButton() -> some View {
+        Button {
+            showFavorites = !showFavorites
+        } label: {
+            filterButtonImage(showFavorites, image: "star", size: 24)
+        }
+    }
+
+    private func filterButtonImage(_ on: Bool, image: String, size: CGFloat) -> some View {
+        Image(systemName: image + (on ? ".fill" : ""))
```
spaced4ndy, line 132: Better to pass as separate parameters
```suggestion
    private func filterButtonImage(_ on: Bool, image: String, imageFill: String, size: CGFloat) -> some View {
        Image(systemName: image + (on ? imageFill : image))
```

apps/ios/Shared/Views/ChatList/ChatListView.swift
```diff
@@ -181,10 +211,16 @@ struct ChatListView: View {
...
-        return s == ""
+        return s == "" && !showFavorites && !showUnread
             ? chatModel.chats
             : chatModel.chats.filter { chat in
-                let contains = chat.chatInfo.chatViewName.localizedLowercase.contains(s)
+                let contains = s != ""
+                                ? chat.chatInfo.chatViewName.localizedLowercase.contains(s)
+                                : (
+                                    (showFavorites && (chat.chatInfo.chatSettings?.favorite ?? false)) ||
+                                    (showUnread && (chat.chatStats.unreadCount > 0 || chat.chatStats.unreadChat)) ||
```
spaced4ndy, line 221: It seems if both are toggled, unread but not favorite, and favorite but not unread chats will be unioned. For these acting like filters, it'd be more natural if they produced intersection of favorite unread chats. Union doesn't have much use since filters can be searched separately

apps/ios/Shared/Views/ChatList/ChatListNavLink.swift
```diff
@@ -171,6 +173,24 @@ struct ChatListNavLink: View {
 
     }
 
+    @ViewBuilder private func toggleFavoriteButton() -> some View {
+        if chat.chatInfo.chatSettings?.favorite == true {
+            Button {
+                toggleChatFavorite(chat, favorite: false)
+            } label: {
+                Label("Unfav", systemImage: "star.slash")
```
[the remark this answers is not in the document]
epoberezkin: doesn't fit

PR #2598, opened 2023-06-19
ios, android: allow to set custom ttl per message if conversation timer is not set

spaced4ndy:


epoberezkin (approved):
Do I understand correctly that it both allows to enable disappearing when disabled (as long as it's allowed), and also to disable it when it's enabled. I thought it would require some core changes though?

spaced4ndy:
> Do I understand correctly that it both allows to enable disappearing when disabled (as long as it's allowed), and also to disable it when it's enabled. I thought it would require some core changes though?

There's no UI for the second

PR #2610, opened 2023-06-23
core: group permision to allow files and media

epoberezkin:


spaced4ndy (commented):
Add a test, otherwise seems ok

PR #2619, opened 2023-06-27
ios: improve chat filter, "No filtered chats" note

epoberezkin:


apps/ios/Shared/Views/ChatList/ChatListView.swift
```diff
@@ -200,11 +206,14 @@ struct ChatListView: View {
...
                 case let .direct(contact):
-                    return contains
-                    || contact.profile.displayName.localizedLowercase.contains(s)
-                    || contact.fullName.localizedLowercase.contains(s)
-                case .contactConnection: return false
-                default: return contains
+                    return contains ||
+                            (s != "" &&
+                             (contact.profile.displayName.localizedLowercase.contains(s) ||
+                              contact.fullName.localizedLowercase.contains(s)))
```
spaced4ndy, line 212: why is this filter only for direct contacts? could be included into `contains`

spaced4ndy (approved):

PR #2620, opened 2023-06-27
android: group preference to prohibit files and media

avently:


apps/android/app/src/main/java/chat/simplex/app/views/chat/ComposeView.kt
```diff
@@ -713,11 +714,22 @@ fun ComposeView(
       modifier = Modifier.padding(end = 8.dp),
       verticalAlignment = Alignment.Bottom,
     ) {
-      IconButton(showChooseAttachment, enabled = !composeState.value.attachmentDisabled && rememberUpdatedState(chat.userCanSend).value) {
+      val isGroupAndProhibitedFiles = chat.chatInfo is ChatInfo.Group && !chat.chatInfo.groupInfo.fullGroupPreferences.files.on
+      val attachmentClicked = if (isGroupAndProhibitedFiles) {
+        {
```
epoberezkin, line 719: why do you need these extra parens?

PR #2623, opened 2023-06-27
android: filter favorite and unread chats

avently:


apps/android/app/src/main/java/chat/simplex/app/views/chatlist/ChatListView.kt
```diff
@@ -235,6 +231,14 @@ private fun ChatListToolbar(chatModel: ChatModel, drawerState: DrawerState, user
             modifier = Modifier.padding(10.dp).size(26.dp)
           )
         }
+        Text(
+          stringResource(R.string.settings_section_title_chats).lowercase().capitalize(Locale.current),
```
epoberezkin, line 235: this is wrong to do it like this, this should certainly be a separate string resource that needs to be added to all languages using the existing translations of "Your chats" without "Your", as the word chosen for settings can be different in some cases.

PR #2639, opened 2023-07-01
multiplatform: make ChatModel and ChatController as object

avently:


apps/multiplatform/android/src/main/java/chat/simplex/app/views/helpers/Share.kt
```diff
@@ -96,7 +96,7 @@ fun imageMimeType(fileName: String): String {
 
 /** Before calling, make sure the user allows to write to external storage [Manifest.permission.WRITE_EXTERNAL_STORAGE] */
 fun saveImage(cxt: Context, ciFile: CIFile?) {
-  val filePath = getLoadedFilePath(cxt, ciFile)
+  val filePath = getLoadedFilePath(ciFile)
```
epoberezkin, line 99: why safeImage still needs cxt?

epoberezkin: also other functions in this file - they accept Context but don't pass it. Looks like there may be other places where Context is passed?

avently: done

PR #2641, opened 2023-07-01
core: use GHC 9.6.2

angerman:
Can be built with:

```
cabal build all -j --allow-newer=base --allow-newer=ghc-prim --allow-newer=template-haskell --allow-newer=bytestring --allow-newer=memory --allow-newer=cryptonite
```

Using ghc 9.6

It mostly runs afoul of https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0366-no-ambiguous-field-access.rst


angerman:
This is just a first stab at it, it won't build with 8.10 _and_ 9.6; for that the `OverloadedRecordDot` can't be used. I was primarily interested to see how much work it would be to even get it to build. For `simplex-chat`, outside of the rather idiotic changes made in proposal 366 (idiotic wrt to breakage), the dependencies were luckily all ready.

dpwiz:
Getting into 9.6 would be nice. Not just for ORFs, but for all those debugging/profiling goodies.

epoberezkin:
This now compiles, passes all tests and doesn't use more than 2gb (instead of 11gb in 8.10.7) - whoever wants to compile CLI for ARM (e.g. on Pi) this is the way.

PR #2652, opened 2023-07-04
ios, core: pull to reconnect relays

epoberezkin:


apps/ios/Shared/Views/ChatList/ChatListView.swift
```diff
@@ -55,6 +55,22 @@ struct ChatListView: View {
         .onChange(of: chatModel.appOpenUrl) { _ in connectViaUrl() }
         .onAppear() { connectViaUrl() }
         .onDisappear() { withAnimation { userPickerVisible = false } }
+        .refreshable {
+            AlertManager.shared.showAlert(Alert(
+                title: Text("Reconnect?"),
+                message: Text("Reconnect all relays"),
```
spaced4ndy, line 61: Relays may be unclear for many users
```suggestion
                message: Text("Reconnect all contacts"),
```

epoberezkin: contacts and groups?

PR #2656, opened 2023-07-04
multiplatform: images and fonts

avently:


apps/multiplatform/android/src/main/java/chat/simplex/app/ui/theme/Type.kt
```diff
@@ -4,16 +4,16 @@ import androidx.compose.material.Typography
...
-  Font(R.font.inter_semi_bold, weight = FontWeight.SemiBold),
-  Font(R.font.inter_medium, weight = FontWeight.Medium),
-  Font(R.font.inter_light, weight = FontWeight.Light),
+val Inter: FontFamily = FontFamily(
+  Font(MR.fonts.Inter.regular.fontResourceId),
+  Font(MR.fonts.Inter.italic.fontResourceId, style = FontStyle.Italic),
+  Font(MR.fonts.Inter.bold.fontResourceId, FontWeight.Bold),
+  Font(MR.fonts.Inter.semibold.fontResourceId, FontWeight.SemiBold),
+  Font(MR.fonts.Inter.medium.fontResourceId, FontWeight.Medium),
+  Font(MR.fonts.Inter.light.fontResourceId, FontWeight.Light)
```
epoberezkin, line 16: are we even using these fonts? I think we have some others installed

apps/multiplatform/common/src/commonMain/resources/distribute/SimpleX.desktop
```diff
@@ -0,0 +1,7 @@
+[Desktop Entry]
+Type=Application
+Name=SimpleX
+Comment=Messenger for the future
```
epoberezkin, line 4: we need different text here

apps/multiplatform/android/src/main/java/chat/simplex/app/ui/theme/Type.kt
avently: https://github.com/simplex-chat/simplex-chat/pull/2656/files#diff-f910305dc69b00400b4b17861797a27e1a98d377ff8ffcecffb291961c870fceL22

https://github.com/simplex-chat/simplex-chat/pull/2656/files#diff-f910305dc69b00400b4b17861797a27e1a98d377ff8ffcecffb291961c870fceL27

https://github.com/simplex-chat/simplex-chat/pull/2656/files#diff-79a51e6d9c4a2a89f013c84187ef8672202cddee9e2f7e7776553ec873902ffcR226

So Bold and Normal are in use at least

> I think we have some others installed

Where?

apps/multiplatform/common/src/commonMain/resources/distribute/SimpleX.desktop
avently: Like what?

apps/multiplatform/android/src/main/java/chat/simplex/app/ui/theme/Type.kt
spaced4ndy: Inter is the font we use

PR #2683, opened 2023-07-12
multiplatform: default files and directories were changed

avently:


apps/multiplatform/common/src/desktopMain/kotlin/chat/simplex/common/platform/Platform.desktop.kt
```diff
@@ -4,14 +4,16 @@ import java.io.File
 import java.util.*
 
 private val home = System.getProperty("user.home")
+private val unixConfigPath = System.getenv("XDG_CONFIG_HOME")?.plus("/SimpleX") ?: "$home/.config/SimpleX"
+private val unixDataPath = System.getenv("XDG_DATA_HOME")?.plus("/SimpleX") ?: "$home/.local/share/SimpleX"
```
epoberezkin, line 8:
```suggestion
private val unixConfigPath = (System.getenv("XDG_CONFIG_HOME") ?: "$home/.config") + "/simplex"
private val unixDataPath = (System.getenv("XDG_DATA_HOME") ?: "$home/.local/share") + "/simplex"
```

PR #2691, opened 2023-07-13
desktop: send images and files

avently:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ComposeView.kt
```diff
@@ -180,7 +180,7 @@ fun ComposeView(
         isImage(uri) -> {
           // Image
           val drawable = getDrawableFromUri(uri)
-          bitmap = if (drawable != null) getBitmapFromUri(uri) else null
+          bitmap = getBitmapFromUri(uri)
```
epoberezkin, line 183: Why condition is removed? This code below is very hard to follow...

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/helpers/ChooseAttachmentView.kt
```diff
@@ -37,21 +38,32 @@ fun ChooseAttachmentView(
...
-        hide()
-      }
-      ActionButton(Modifier.fillMaxWidth(0.50f), null, stringResource(MR.strings.gallery_video_button), icon = painterResource(MR.images.ic_smart_display)) {
-        attachmentOption.value = AttachmentOption.GalleryVideo
-        hide()
-      }
-      ActionButton(Modifier.fillMaxWidth(1f), null, stringResource(MR.strings.choose_file), icon = painterResource(MR.images.ic_note_add)) {
-        attachmentOption.value = AttachmentOption.File
-        hide()
+      if (appPlatform.isAndroid) {
```
epoberezkin, line 41: given the size it's the good candidate to extract into expect/actual functions

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ComposeView.kt
avently: `getDrawableFromUri` needed on Android because it can say is it AnimatedDrawable or not. There is no such thing on Desktop and this function returns null always. So if on desktop it returns null and the check `if (drawable != null)` is here, bitmap will be null too on desktop which is not what I want. 

`if (drawable != null)` was needed just to make less work on Android: if `getDrawableFromUri` returns null there, no file actually exists so no need to try to get bitmap from it. But as I said, this optimization will not work with desktop,

epoberezkin: can we make this optimisation conditional on the platform then?

avently: Let CPU do it's 1 ms work

PR #2701, opened 2023-07-14
ios: delivery receipts

spaced4ndy:


apps/ios/Shared/Model/SimpleXAPI.swift
```diff
@@ -1121,6 +1133,7 @@ func startChat(refreshInvitations: Bool = true) throws {
             m.onboardingStage = [.step1_SimpleXInfo, .step2_CreateProfile].contains(savedOnboardingStage) && m.users.count == 1
                                 ? .step3_CreateSimpleXAddress
                                 : savedOnboardingStage
+            // TODO don't show on first start
```
spaced4ndy, line 1136: tested on a new app - it doesn't show
```suggestion
```

apps/ios/Shared/Views/UserSettings/PrivacySettings.swift
```diff
@@ -86,46 +87,45 @@ struct PrivacySettings: View {
...
+                        contactReceipts.toggle()
                     }
                 }
             }
         }
         .onAppear {
-            if let currentUser = m.currentUser {
-                sendReceiptsContacts = currentUser.sendRcptsContacts
+            if let u = m.currentUser, contactReceipts != u.sendRcptsContacts {
+                contactReceiptsReset = true
```
spaced4ndy, line 120: I've tried setting it as well and it sometimes lead to a different race where first toggle change wasn't working.. but maybe something else changed and it's ok now (it didn't bug again when I tested with new code)

Issue #2703, opened 2023-07-15
[Q]: Incorrect installation instructions for GHC


justinmoon:
### Question

[This section](https://github.com/simplex-chat/simplex-chat/blob/stable/docs/CLI.md#in-any-os) contains instructions for installing GHC and other Haskell build tools. When I run them, I get:

```
$ ghc --version
The Glorious Glasgow Haskell Compilation System, version 9.2.8
```

But we need version `8.10.7`.

PR #2704, opened 2023-07-16
nix: install ghc, cabal, zlib and gmp in dev shell

justinmoon:
I've never used Haskell before, and had some trouble installing it using [the instructions](https://github.com/simplex-chat/simplex-chat/issues/2703). But adding all the Haskell and build dependencies to the shell in `flake.nix` worked for me. I was able to run `nix develop` and `cabal install` and build a usable `simplex-chat` binary.

Perhaps this will be useful to others?

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -339,7 +339,7 @@
           '';
         }; in
 	pkgs.mkShell {
-          buildInputs = [ updateCmd ];
+          buildInputs = [ updateCmd pkgs.haskell.compiler.ghc810 pkgs.cabal-install pkgs.zlib pkgs.gmp ];
           shellHook = ''
             echo "welcome to the shell!"
           '';


dunxen:
> Perhaps this will be useful to others?

Yes, this would be great! I had to do the same as in your diff. I think it would round up the flake nicely to just be able to `nix develop` :)

flake.nix
```diff
@@ -339,7 +339,7 @@
           '';
         }; in
 	pkgs.mkShell {
-          buildInputs = [ updateCmd ];
+          buildInputs = [ updateCmd pkgs.haskell.compiler.ghc810 pkgs.cabal-install pkgs.zlib pkgs.gmp ];
```
dunxen, line 342:
```suggestion
          buildInputs = [ updateCmd pkgs.haskell.compiler.ghc810 pkgs.cabal-install pkgs.zlib pkgs.gmp pkgs.openssl ];
```

I believe I also needed openssl as I didn't have that globally. I'm not sure, but `pkg-config` might also be needed to find the openssl headers.

PR #2719, opened 2023-07-18
core: group snd status

spaced4ndy:


src/Simplex/Chat/Messages.hs
```diff
@@ -702,6 +704,44 @@ ciCreateStatus content = case msgDirection @d of
...
+  | GSPOneThird
+  | GSPTwoThirds
+  | GSPComplete
+  deriving (Show, Generic)
+
+instance ToJSON GroupSndCIStatusProgress where
+  toJSON = J.genericToJSON . sumTypeJSON $ dropPrefix "GSP"
+  toEncoding = J.genericToEncoding . sumTypeJSON $ dropPrefix "GSP"
+
+type SndCIStatus = CIStatus 'MDSnd
```
epoberezkin, line 733: no need for synonym

PR #2741, opened 2023-07-20
multiplatform: merged master

avently:


spaced4ndy:
apps/multiplatform/android/src/main/java/chat/simplex/app/views/helpers/Util.kt likely can be deleted, says empty file

PR #2793, opened 2023-07-28
desktop: responsive design for large and small screens

avently:


avently:
The codebase has changed heavily since then, code here is not actual anymore.

PR #2814, opened 2023-07-31
ci: Github Action to build desktop app binaries

avently:


.github/workflows/build.yml
```diff
@@ -111,28 +116,74 @@ jobs:
...
-        id: unix_build
+      - name: Unix build CLI
+        id: unix_cli_build
         if: matrix.os != 'windows-latest'
         shell: bash
         run: |
           cabal build --enable-tests
           echo "::set-output name=bin_path::$(cabal list-bin simplex-chat)"
 
+      - name: Unix test
```
spaced4ndy, line 127: better to keep this step after upload to release steps - it was moved as a workaround for sporadic test failures in CI

avently: ok

avently:
This is an example of an output from this action in releases: https://github.com/avently/simplex-chat/releases/tag/v1.0.10

PR #2844, opened 2023-08-03
directory: delist/relist groups when service or owner roles change

epoberezkin:


tests/Bots/DirectoryTests.hs
```diff
@@ -215,10 +219,121 @@ testDelistedServiceRemoved tmp =
...
+        bob <## "The group is listed in the directory again."
+        superUser <# "SimpleX-Directory> The group ID 1 (privacy) is listed (SimpleX-Directory role is changed to admin)."
+        groupFound cath "privacy"
+        -- de-listed if owner role changed
+        cath ##> "/mr privacy bob admin"
+        cath <## "#privacy: you changed the role of bob from owner to admin"
+        bob <## "#privacy: cath changed your role from owner to admin"
+        bob <# "SimpleX-Directory> Your role in the group ID 1 (privacy) is changed to admin."
+        bob <## ""
+        bob <## "The group is no longer listed in the directory."
```
spaced4ndy, line 259: service could also send a message to the group so that member who changes role of registrator learns this results in group being de-listed. Same for profile change, member removal, etc.

tests/Bots/DirectoryTests.hs
```diff
@@ -215,10 +219,121 @@ testDelistedServiceRemoved tmp =
...
+testNotDelistedMemberRoleChanged :: HasCallStack => FilePath -> IO ()
+testNotDelistedMemberRoleChanged tmp =
+  withDirectoryService tmp $ \superUser dsLink ->
+    withNewTestChat tmp "bob" bobProfile $ \bob -> do
+      withNewTestChat tmp "cath" cathProfile $ \cath -> do
+        bob `connectVia` dsLink
+        registerGroup superUser bob "privacy" "Privacy"
+        addCathAsOwner bob cath
+        groupFound cath "privacy"
+        -- de-listed if service role changed
```
spaced4ndy, line 281:
```suggestion
```

tests/Bots/DirectoryTests.hs
epoberezkin: interesting, but probably unnecessary, as only owners can do it, and usually there is only one owner, and if not - they can say it in the group... Also, in the future approval should be automated, so no drama...

PR #2851, opened 2023-08-04
ios: rework incognito mode - choose when making connection

spaced4ndy:


apps/ios/Shared/ContentView.swift
```diff
@@ -28,6 +28,17 @@ struct ContentView: View {
     @State private var showWhatsNew = false
     @State private var showChooseLAMode = false
     @State private var showSetPasscode = false
+    @State private var mainViewActionSheet: MainViewActionSheet? = nil
```
epoberezkin, line 31: Why does action sheet needs to be passed via state, when it can be just boolean?

epoberezkin: And even if it needs to become enum, why pass the actual view definition via the state, and not just plain enum?

apps/ios/Shared/Views/UserSettings/IncognitoHelp.swift
```diff
@@ -18,10 +18,9 @@ struct IncognitoHelp: View {
             ScrollView {
                 VStack(alignment: .leading) {
                     Group {
-                    Text("Incognito mode protects the privacy of your main profile name and image — for each new contact a new random profile is created.")
+                    Text("Incognito mode protects the privacy of your profile — you can choose to create a new random profile for each new contact.")
```
epoberezkin, line 21:
```suggestion
                    Text("Incognito mode protects your privacy by using a new random profile for each contact.")
```

apps/ios/Shared/Views/UserSettings/IncognitoHelp.swift
```diff
@@ -18,10 +18,9 @@ struct IncognitoHelp: View {
             ScrollView {
                 VStack(alignment: .leading) {
                     Group {
-                    Text("Incognito mode protects the privacy of your main profile name and image — for each new contact a new random profile is created.")
+                    Text("Incognito mode protects your privacy by using a new random profile for each contact.")
                     Text("It allows having many anonymous connections without any shared data between them in a single chat profile.")
                     Text("When you share an incognito profile with somebody, this profile will be used for the groups they invite you to.")
-                    Text("To find the profile used for an incognito connection, tap the contact or group name on top of the chat.")
```
epoberezkin, line 24: why this is removed?

apps/ios/Shared/Views/ChatList/ChatListView.swift
```diff
@@ -60,8 +60,6 @@ struct ChatListView: View {
                 chatList
             }
         }
-        .onChange(of: chatModel.appOpenUrl) { _ in connectViaUrl() }
-        .onAppear() { connectViaUrl() }
```
epoberezkin, line 64: this onAppear handler wasn't replace with the one above

epoberezkin: found it

PR #2873, opened 2023-08-08
android: workaround of system restricted background

avently:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/call/CallManager.kt
```diff
@@ -24,10 +25,12 @@ class CallManager(val chatModel: ChatModel) {
     }
   }
 
-  fun acceptIncomingCall(invitation: RcvCallInvitation) {
+  fun acceptIncomingCall(invitation: RcvCallInvitation, lockscreen: Boolean = false) = withBGApi {
     if (appPlatform.isDesktop) {
-      return showInDevelopingAlert()
+      return@withBGApi showInDevelopingAlert()
```
epoberezkin, line 30: what does it mean @withBGApi here?

PR #2886, opened 2023-08-10
 core: communicate connection chat version range

spaced4ndy:


src/Simplex/Chat/Protocol.hs
```diff
@@ -100,9 +107,25 @@ data AppMessage (e :: MsgEncoding) where
...
+
+instance ToJSON ChatVersionRange where
+  toJSON (ChatVersionRange vr) = strToJSON vr
+  toEncoding (ChatVersionRange vr) = strToJEncoding vr
+
+toChatVRange :: VersionRange -> ChatVersionRange
+toChatVRange = ChatVersionRange
+
+fromChatVRange :: ChatVersionRange -> VersionRange
+fromChatVRange (ChatVersionRange vr) = vr
```
epoberezkin, line 123: why these are needed when you have constructor and selector?

src/Simplex/Chat/Protocol.hs
```diff
@@ -100,9 +107,25 @@ data AppMessage (e :: MsgEncoding) where
   AMJson :: AppMessageJson -> AppMessage 'Json
   AMBinary :: AppMessageBinary -> AppMessage 'Binary
 
+newtype ChatVersionRange = ChatVersionRange {versionRange :: VersionRange} deriving (Eq, Show)
```
epoberezkin, line 110: not sure we need this new type - where JSON instance for VersionRange is used?

src/Simplex/Chat/Store/Direct.hs
```diff
@@ -489,10 +496,26 @@ createOrUpdateContactRequest db user@User {userId} userContactLinkId invId Profi
...
-        then Right <$> DB.execute db "UPDATE contact_requests SET agent_invitation_id = ?, updated_at = ? WHERE user_id = ? AND contact_request_id = ?" (invId, currentTs, userId, cReqId)
+        then
+          Right
+            <$> DB.execute
+              db
+              [sql|
+            UPDATE contact_requests
+            SET agent_invitation_id = ?, chat_vrange_min_version = ?, chat_vrange_max_version = ?, updated_at = ?
+            WHERE user_id = ? AND contact_request_id = ?
+          |]
```
epoberezkin, line 507:
```suggestion
              [sql|
                UPDATE contact_requests
                SET agent_invitation_id = ?, chat_vrange_min_version = ?, chat_vrange_max_version = ?, updated_at = ?
                WHERE user_id = ? AND contact_request_id = ?
              |]
```

src/Simplex/Chat.hs
```diff
@@ -2210,7 +2215,8 @@ acceptFileReceive user@User {userId} RcvFileTransfer {fileId, xftpRcvFile, fileI
   case (xftpRcvFile, fileConnReq) of
     -- direct file protocol
     (Nothing, Just connReq) -> do
-      connIds <- joinAgentConnectionAsync user True connReq . directMessage $ XFileAcpt fName
+      dm <- directMessage $ XFileAcpt fName
+      connIds <- joinAgentConnectionAsync user True connReq dm
```
epoberezkin, line 2219:
```suggestion
      connIds <- joinAgentConnectionAsync user True connReq <=< directMessage $ XFileAcpt fName
```

epoberezkin: a lot of noice because of not having the version in scope – why not just make it available in the scope of directMessage?

Or s/./<=</ - just changing one composition with another

epoberezkin: or

```
connIds <- joinAgentConnectionAsync user True connReq =<< directMessage (XFileAcpt fName)

Comment (epoberezkin) [reply src/Simplex/Chat.hs:2219]: by introducing an intermediate variable in all these places it only becomes harder to read, not easier.

File: src/Simplex/Chat/Protocol.hs [hunk 5]
```diff
@@ -161,7 +184,11 @@ instance ToJSON MsgRef where
   toJSON = J.genericToJSON J.defaultOptions {J.omitNothingFields = True}
   toEncoding = J.genericToEncoding J.defaultOptions {J.omitNothingFields = True}
 
-data ChatMessage e = ChatMessage {msgId :: Maybe SharedMsgId, chatMsgEvent :: ChatMsgEvent e}
+data ChatMessage e = ChatMessage
+  { chatVRange :: Maybe VersionRange,
```
epoberezkin, line 188: this doesn't need to be maybe, the absence of version in JSON should be interpreted as specific version, that would simplify the logic in other places.

src/Simplex/Chat/Protocol.hs
epoberezkin: in the db it's also not nullable

src/Simplex/Chat/Protocol.hs
epoberezkin: in the db it's also not nullable

src/Simplex/Chat/Store/Direct.hs
```diff
@@ -411,13 +413,16 @@ getUserContacts db user@User {userId} = do
   contactIds <- map fromOnly <$> DB.query db "SELECT contact_id FROM contacts WHERE user_id = ? AND deleted = 0" (Only userId)
   rights <$> mapM (runExceptT . getContact db user) contactIds
 
-createOrUpdateContactRequest :: DB.Connection -> User -> Int64 -> InvitationId -> Profile -> Maybe XContactId -> ExceptT StoreError IO ContactOrRequest
-createOrUpdateContactRequest db user@User {userId} userContactLinkId invId Profile {displayName, fullName, image, contactLink, preferences} xContactId_ =
+createOrUpdateContactRequest :: DB.Connection -> User -> Int64 -> InvitationId -> Maybe VersionRange -> Profile -> Maybe XContactId -> ExceptT StoreError IO ContactOrRequest
+createOrUpdateContactRequest db user@User {userId} userContactLinkId invId chatVRange Profile {displayName, fullName, image, contactLink, preferences} xContactId_ =
```
epoberezkin, line 417: this would be a simple pattern match without Maybe

src/Simplex/Chat/Store/Shared.hs
```diff
@@ -150,12 +152,12 @@ toConnection ((connId, acId, connLevel, viaContact, viaUserContactLink, viaGroup
...
+toMaybeConnection ((Just connId, Just agentConnId, Just connLevel, viaContact, viaUserContactLink, Just viaGroupLink, groupLinkId, customUserProfileId, Just connStatus, Just connType, Just localAlias) :. (contactId, groupMemberId, sndFileId, rcvFileId, userContactLinkId) :. (Just createdAt, code_, verifiedAt_, Just authErrCounter, Just minVer, Just maxVer)) =
+  Just $ toConnection ((connId, agentConnId, connLevel, viaContact, viaUserContactLink, viaGroupLink, groupLinkId, customUserProfileId, connStatus, connType, localAlias) :. (contactId, groupMemberId, sndFileId, rcvFileId, userContactLinkId) :. (createdAt, code_, verifiedAt_, authErrCounter, minVer, maxVer))
 toMaybeConnection _ = Nothing
 
-createConnection_ :: DB.Connection -> UserId -> ConnType -> Maybe Int64 -> ConnId -> Maybe ContactId -> Maybe Int64 -> Maybe ProfileId -> Int -> UTCTime -> IO Connection
-createConnection_ db userId connType entityId acId viaContact viaUserContactLink customUserProfileId connLevel currentTs = do
+createConnection_ :: DB.Connection -> UserId -> ConnType -> Maybe Int64 -> ConnId -> Maybe VersionRange -> Maybe ContactId -> Maybe Int64 -> Maybe ProfileId -> Int -> UTCTime -> IO Connection
```
epoberezkin, line 159: here too, no need for it to be Maybe, it only complicates things

PR #2896, opened 2023-08-11
desktop: changing language without reload

avently:


apps/multiplatform/common/src/desktopMain/kotlin/chat/simplex/common/DesktopApp.kt
```diff
@@ -34,84 +35,87 @@ fun showApp() = application {
...
-      if (simplexWindowState.openDialog.isAwaiting) {
-        FileDialogChooser(
-          title = "SimpleX",
-          isLoad = true,
-          params = simplexWindowState.openDialog.params,
-          onResult = {
-            simplexWindowState.openDialog.onResult(it.firstOrNull())
-          }
-        )
+  // Dirty hack to reload all strings in all @Composable's after language change at runtime
```
epoberezkin, line 38:
```suggestion
  // Reload all strings in all @Composable's after language change at runtime
```

PR #2899, opened 2023-08-11
ios: only show "entity" notifications if the are enabled in the conversation

epoberezkin:


apps/ios/SimpleXChat/ChatTypes.swift
```diff
@@ -1466,6 +1466,8 @@ public struct SecurityCode: Decodable, Equatable {
 
 public struct UserContact: Decodable {
     public var userContactLinkId: Int64
+//    public var connReqContact: String
```
spaced4ndy, line 1469:
```suggestion
```

spaced4ndy (approved):

PR #2908, opened 2023-08-12
core: explicitly set encoding to utf-8 (fixes unicode filenames on mobile)

angerman:
GHC's rts tries to obtain the encoding from iconv, however this is not really available on iOS. It therefore defaults to US-ASCII, and then breaks with unicode data. We now explicitly set this to utf-8 here.


angerman:
Confirmed to be working on iOS through a custom built.

PR #2925, opened 2023-08-14
ios: improve group layout

epoberezkin:


apps/ios/Shared/Views/Chat/ChatItemView.swift
```diff
@@ -130,60 +126,31 @@ struct ChatItemContentView<Content: View>: View {
...
-            return nil
-        }
+    private var membersConnectedText: String? {
+        var ns = chatModel.getConnectedMemberNames(chatItem)
+        return ns.count > 3
+            ? "\(ns[0]), \(ns[1]) and \(ns.count - 2) other members connected"
+            : ns.count == 3
+            ? "\(ns[0] + ", " + ns[1]) and \(ns[2]) connected"
+            : ns.count == 2
+            ? "\(ns[0]) and \(ns[1]) connected"
```
spaced4ndy, line 144: why these shouldn't be NSLocalizedString?

apps/ios/Shared/Views/Chat/ChatItem/FramedItemView.swift
```diff
@@ -232,17 +231,27 @@ struct FramedItemView: View {
...
+        }
+        .padding(.top, 6)
         .padding(.horizontal, 12)
     }
     
+    private func ciQuotedMsgTextView(_ qi: CIQuote, lines: Int) -> some View {
+        MsgContentView(text: qi.text, formattedText: qi.formattedText)
+        .lineLimit(lines)
+        .font(.subheadline)
+        .padding(.bottom, 6)
```
spaced4ndy, line 252: formatting seems to be wrong

apps/ios/Shared/Views/Chat/ChatView.swift
```diff
@@ -430,63 +430,64 @@ struct ChatView: View {
...
-                    .padding(.leading, 8)
-                    .environmentObject(chat)
+                    .padding(.top, 5)
+                    .padding(.trailing)
+                    .padding(.leading, 12)
+                } else {
+                    chatItemWithMenu(ci, maxWidth)
+                    .padding(.top, 5)
+                    .padding(.trailing)
+                    .padding(.leading, memberImageSize + 8 + 12)
```
spaced4ndy, line 464: same

apps/ios/Shared/Views/Chat/ChatItemView.swift
epoberezkin: the question is why the should be. These are simpler.

PR #2934, opened 2023-08-15
android: members connected aggregated item; group layout

spaced4ndy:


apps/ios/SimpleXChat/ChatTypes.swift
```diff
@@ -2543,6 +2543,9 @@ public enum CIContent: Decodable, ItemContent {
         case .rcvDecryptionError: return true
         case .rcvGroupInvitation: return true
         case .sndChatPreference: return true
+        case .rcvChatPreference: return true
+        case .sndGroupFeature: return true
+        case .rcvGroupFeature: return true
```
spaced4ndy, line 2548: these items appear in groups and they don't have member name

PR #3003, opened 2023-09-02
ui: types and stubs to encrypt local files

epoberezkin:


apps/ios/Shared/Views/Chat/ComposeMessage/ComposeView.swift
```diff
@@ -656,10 +654,11 @@ struct ComposeView: View {
                 }
             case let .voicePreview(recordingFileName, duration):
                 stopPlayback.toggle()
-                chatModel.filesToDelete.remove(getAppFilePath(recordingFileName))
-                sent = await send(.voice(text: msgText, duration: duration), quoted: quoted, file: recordingFileName, ttl: ttl)
+                // TODO encrypt voice
```
spaced4ndy, line 657: it seems it's done in voiceCryptoFile
```suggestion
```

apps/ios/SimpleXChat/ChatTypes.swift
```diff
@@ -2112,6 +2112,17 @@ public struct ChatItem: Identifiable, Decodable {
...
     }
 
+    public var encryptedFile: Bool? {
+        guard let fileSource = file?.fileSource else { return nil }
+        return fileSource.cryptoArgs != nil
+    }
+
+    public var encryptLocalFile: Bool {
+        file?.fileProtocol == .xftp &&
+        content.msgContent?.isVideo == false &&
```
spaced4ndy, line 2122: weren't regular files going to not be encrypted initially too?

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/SimpleXAPI.kt
```diff
@@ -1413,7 +1413,8 @@ object ChatController {
             ((mc is MsgContent.MCImage && file.fileSize <= MAX_IMAGE_SIZE_AUTO_RCV)
                 || (mc is MsgContent.MCVideo && file.fileSize <= MAX_VIDEO_SIZE_AUTO_RCV)
                 || (mc is MsgContent.MCVoice && file.fileSize <= MAX_VOICE_SIZE_AUTO_RCV && file.fileStatus !is CIFileStatus.RcvAccepted))) {
-          withApi { receiveFile(r.user, file.fileId, auto = true) }
+          // TODO encrypt images and voice
+          withApi { receiveFile(r.user, file.fileId, encrypted = false, auto = true) }
```
spaced4ndy, line 1417: TODO left

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/item/CIImageView.kt
```diff
@@ -152,7 +152,8 @@ fun CIImageView(
           when (file.fileStatus) {
             CIFileStatus.RcvInvitation ->
               if (fileSizeValid()) {
-                receiveFile(file.fileId)
+                // TODO encrypt image
+                receiveFile(file.fileId, false)
```
spaced4ndy, line 156: + here

spaced4ndy (approved):

apps/ios/SimpleXChat/ChatTypes.swift
epoberezkin: they are now encrypted - it's the same logic as with voice - they are decrypted to temp folder and removed after sharing activity completes (in the completion handler)

epoberezkin: it also optimized the logic of copying the file over to app storage - previously it was fully read into memory instead of just copying, for reasons unknown...

epoberezkin: imaging trying to send 1gb file

PR #3027, opened 2023-09-06
android: changing a chat on user change

avently:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/App.kt
```diff
@@ -194,24 +193,27 @@ fun AndroidScreen(settingsState: SettingsViewState) {
...
       launch {
         snapshotFlow { chatModel.chatId.value }
           .distinctUntilChanged()
           .collect {
-            if (it != null) currentChatId = it
-            else onComposed()
+            if (it == null) {
+              onComposed(null)
+            }
+            currentChatId = it
```
epoberezkin, line 216:
```suggestion
            if (it == null) onComposed(null)
            currentChatId = it
```

hm, so the problem was that it was not resetting currentChatId to null?

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/App.kt
```diff
@@ -194,24 +193,27 @@ fun AndroidScreen(settingsState: SettingsViewState) {
...
         offset.animateTo(
-          if (chatModel.chatId.value == null) 0f else maxWidth.value,
+          if (chatId == null) 0f else maxWidth.value,
           chatListAnimationSpec()
         )
-        if (offset.value == 0f) {
-          currentChatId = null
-        }
-      }
+      }.join()
```
epoberezkin, line 206: what does it do?

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/App.kt
avently: No, the problem is much larger that this line. When the app setting chatId to null, and after a moment it sets chatId to some chat.id, the animation was still in progress and the way of setting currentChatId was wrong. So it just didn't react on re-setting chatId if this resetting was fast. Not it listens for every change and reacts accordingly.

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/App.kt
avently: There is a comment describing why it's here:
https://github.com/simplex-chat/simplex-chat/pull/3027/files/198873ecad601c0acfba8f8bf3c7aaa274cb54e8#diff-ac9b6f94f88dca783c51ee54710c5c043ed44b7c9a7e49e8187e9b26003a378cR197-R200

PR #3033, opened 2023-09-07
Fix broken RSS feed source URL

sethforprivacy:
Noticed that the blog source URL is incorrectly set and has a comma that breaks the URL.

image


epoberezkin:
thank you!

PR #3053, opened 2023-09-13
core: api to create contacts with group members

spaced4ndy:


src/Simplex/Chat.hs
```diff
@@ -1588,6 +1588,33 @@ processChatCommand = \case
...
+    -- create connection and contact
+    -- reuse member incognito profile
+    -- how to differentiate between ready contacts and pending member contacts in UI?
+    -- (it should affect whether send, attachment is enabled; invite button inside chat;
+    -- currently non ready contacts can't be opened in UI)
+    --   - special cases (via new field pending_contact_member_id?)
+    --   - or new entity - PendingMemberContact?
+    -- CRNewMemberContact
+    ok_
+  APISendMemberContactInvitation contactId msgContent_ -> do
```
epoberezkin, line 1603: maybe just use send message?

src/Simplex/Chat.hs
```diff
@@ -1588,6 +1588,33 @@ processChatCommand = \case
...
+    -- get group and group member by contact (via new field pending_contact_member_id?)
+    -- send XGrpDirectInv to member
+    -- optionally create chat item for contact
+    -- further messages / repeat invitation:
+    --   - flag that message was sent? (prohibit to send further messages until member accepts?)
+    --   - or allow to send pending messages to not ready contacts? (+ pending_direct_messages table)
```
epoberezkin, line 1609: I think just to prohibit initially?

src/Simplex/Chat.hs
```diff
@@ -4489,6 +4517,21 @@ processAgentMessageConn user@User {userId} corrId agentConnId agentMessage = do
           groupMsgToView g' m ci msgMeta
         createGroupFeatureChangedItems user cd CIRcvGroupFeature g g'
 
+    xGrpDirectInv :: GroupInfo -> GroupMember -> ConnReqInvitation -> Maybe MsgContent -> RcvMessage -> MsgMeta -> m ()
+    xGrpDirectInv g m connReq mContent_ msg msgMeta = do
+      -- check direct messages preference
+      -- auto-join / client setting? (new to_join field if auto join is in UI?)
```
epoberezkin, line 4523: yes, global setting with group overrides, on by default? We have some framework for them, they are in ChatSettings as separate fields.

src/Simplex/Chat.hs
```diff
@@ -4489,6 +4517,21 @@ processAgentMessageConn user@User {userId} corrId agentConnId agentMessage = do
...
+      -- create contact (PendingMemberContact?)
```
epoberezkin, line 4524: I don't yet see why another type is needed

src/Simplex/Chat.hs
```diff
@@ -4489,6 +4517,21 @@ processAgentMessageConn user@User {userId} corrId agentConnId agentMessage = do
...
+      -- if member already has contact, replace it (to not mix messages from connections)
```
epoberezkin, line 4525: didn't get it

src/Simplex/Chat.hs
```diff
@@ -4489,6 +4517,21 @@ processAgentMessageConn user@User {userId} corrId agentConnId agentMessage = do
...
+      -- create chat item for contact
+      -- CRNewMemberContactInvitation
+      -- how to display "contact invitation" in UI?
```
epoberezkin, line 4528: I thought as a special message in the group? But then maybe also use contact request view in the list as well?

src/Simplex/Chat.hs
```diff
@@ -4489,6 +4517,21 @@ processAgentMessageConn user@User {userId} corrId agentConnId agentMessage = do
...
+      --   - special interaction in chat list + can open chat?
+      --   - accept button inside chat?
+      -- how to differentiate initiating and joining contact? (joining contact should have "accept" interaction)
```
epoberezkin, line 4531: it does seem like accepting should be another type, unless it's auto-accepted

src/Simplex/Chat.hs
spaced4ndy: If contact already exists for member receiving XGrpDirectInv (and doesn't exist for sending member), contact reference on member record will be replaced for the receiving member. Otherwise messages from two connections will be mixed in one contact. Or receiving member could prohibit XGrpDirectInv but then initiating member would never be able to connect directly

src/Simplex/Chat/Protocol.hs
```diff
@@ -223,6 +223,7 @@ data ChatMsgEvent (e :: MsgEncoding) where
   XGrpLeave :: ChatMsgEvent 'Json
   XGrpDel :: ChatMsgEvent 'Json
   XGrpInfo :: GroupProfile -> ChatMsgEvent 'Json
+  XGrpDirectInv :: ConnReqInvitation -> Maybe MsgContent -> ChatMsgEvent 'Json
```
epoberezkin, line 226: how would it work when it is Nothing? simply show an empty chat with another member in the list? do we want it?

src/Simplex/Chat/Controller.hs
```diff
@@ -282,6 +282,10 @@ data ChatCommand
   | APIGroupLinkMemberRole GroupId GroupMemberRole
   | APIDeleteGroupLink GroupId
   | APIGetGroupLink GroupId
+  | APICreateMemberContact GroupId GroupMemberId
+  | APISendMemberContactInvitation {contactId :: ContactId, msgContent_ :: Maybe MsgContent}
+  | CreateMemberContact GroupName ContactName
+  | SendMemberContactInvitation {contactName :: ContactName, message_ :: Maybe Text}
```
epoberezkin, line 288: I don't think the terminal commands are needed here, it's not too useful. It's actually the argument to modify send to allow sending to members, or you'll have to put some logic in terminal send to do two things. Member have a unique local display name, so we need to preserve the ability to simply send messages to them. And if we do, we may as well create contact and send message in one go from the UI as well (although it's a separate consideration).

src/Simplex/Chat.hs
```diff
@@ -1588,6 +1588,40 @@ processChatCommand = \case
...
     pure $ CRGroupLink user gInfo groupLink mRole
+  APICreateMemberContact gId gMemberId -> withUser $ \user -> do
+    (g, m) <- withStore $ \db -> (,) <$> getGroupInfo db user gId <*> getGroupMember db user gId gMemberId
+    assertUserGroupRole g GRAuthor
+    unless (groupFeatureAllowed SGFDirectMessages g) $ throwChatError $ CECommandError "direct messages not allowed"
+    case memberConn m of
+      Just mConn@Connection {peerChatVRange} -> do
+        unless (isCompatibleRange (fromJVersionRange peerChatVRange) xGrpDirectInvVRange) $ throwChatError CEPeerChatVRangeIncompatible
+        let GroupMember {memberContactId} = m
+        when (isJust memberContactId) $ throwChatError $ CECommandError "member contact already exists"
```
epoberezkin, line 1599:
```suggestion
        when (isJust $ memberContactId m) $ throwChatError $ CECommandError "member contact already exists"
```

src/Simplex/Chat/View.hs
```diff
@@ -230,6 +230,9 @@ responseToView user_ ChatConfig {logLevel, showReactions, showReceipts, testView
   CRGroupLink u g cReq mRole -> ttyUser u $ groupLink_ "Group link:" g cReq mRole
   CRGroupLinkDeleted u g -> ttyUser u $ viewGroupLinkDeleted g
   CRAcceptingGroupJoinRequest _ g c -> [ttyFullContact c <> ": accepting request to join group " <> ttyGroup' g <> "..."]
+  CRNewMemberContact u Contact {localDisplayName = c} g m -> ttyUser u ["contact for member " <> ttyGroup' g <> " " <> ttyMember m <> " prepared, use " <> highlight ("/invite member contact @" <> c <> " <message>") <> " to send invitation"]
+  CRNewMemberContactSentInv u _ct g m -> ttyUser u ["sent invitation to connect directly to member " <> ttyGroup' g <> " " <> ttyMember m]
+  CRNewMemberContactReceivedInv u ct g m -> ttyUser u [ttyGroup' g <> " " <> ttyMember m <> " is creating direct contact " <> ttyContact' ct <> " with you"]
```
epoberezkin, line 235: I don't think it needs to be shown in terminal when there is a message

PR #3078, opened 2023-09-19
android: create contacts with group members

spaced4ndy:


apps/multiplatform/common/src/desktopMain/kotlin/chat/simplex/common/platform/PlatformTextField.desktop.kt
```diff
@@ -51,6 +52,13 @@ actual fun PlatformTextField(
     delay(50)
     keyboard?.show()
   }
+  LaunchedEffect(sendMsgEnabled) {
```
spaced4ndy, line 55: TODO test

PR #3081, opened 2023-09-19
core: update contact and member profiles for both sides when contact is created with member (WIP)

epoberezkin:
…is created with member (WIP)

### DIFF
--- a/src/Simplex/Chat.hs
+++ b/src/Simplex/Chat.hs
@@ -12,7 +12,6 @@
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeApplications #-}
-
 {-# OPTIONS_GHC -fno-warn-ambiguous-fields #-}
 
 module Simplex.Chat where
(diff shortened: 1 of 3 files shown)


src/Simplex/Chat.hs
```diff
@@ -3062,10 +3062,17 @@ processAgentMessageConn user@User {userId} corrId agentConnId agentMessage = do
               -- TODO update member profile
               -- [async agent commands] no continuation needed, but command should be asynchronous for stability
               allowAgentConnectionAsync user conn' confId XOk
-            XOk -> do
-              allowAgentConnectionAsync user conn' confId XOk
-              void $ withStore' $ \db -> resetMemberContactFields db ct
+            XInfo p' -> confXInfo ct conn' $ Just p'
+            XOk -> confXInfo ct conn' Nothing
```
spaced4ndy, line 3066: XOk is not necessary to process - it was only added for this feature

tests/ChatTests/Groups.hs
```diff
@@ -3047,3 +3048,65 @@ testMemberContactIncognito =
...
+
+      alice ##> "/p alisa Alisa"
+      alice <## "user profile is changed to alisa (Alisa) (your 1 contacts are notified)"
+      bob <## "contact alice changed to alisa (Alisa)"
+      bob <## "use @alisa <message> to send messages"
+
+      alice `send` "@cath hello"
+      -- TODO too many outputs here, and maybe should follow the output of the message?
+      -- two lines - connecting..., connected would be ok
+      alice <## "member #team cath does not have direct connection, creating"
```
spaced4ndy, line 3088: this output is only in CLI so that if there's error on creating contact, it's more clear what chat was trying to do. For example, there're tests where there's failure immediately due to version incompatibility and w/t this line it looks like:

"@cath hello"
"peer chat version range incompatible"

with this line it's at least clear chat was trying to create member contact

tests/ChatTests/Groups.hs
```diff
@@ -3047,3 +3048,65 @@ testMemberContactIncognito =
...
+      bob <# "#team alice> hello"
+      cath <# "#team alice> hello"
+      cath #> "#team hello too"
+      bob <# "#team cath> hello too"
+      alice <# "#team cath> hello too"
+
+      alice ##> "/p alisa Alisa"
+      alice <## "user profile is changed to alisa (Alisa) (your 1 contacts are notified)"
+      bob <## "contact alice changed to alisa (Alisa)"
+      bob <## "use @alisa <message> to send messages"
```
spaced4ndy, line 3083: cath should also change profile, they send update in different messages

PR #3092, opened 2023-09-21
Remote desktop

dpwiz:


src/Simplex/Chat/Controller.hs
```diff
@@ -420,7 +446,16 @@ data ChatCommand
   deriving (Show)
 
 data ChatResponse
-  = CRActiveUser {user :: User}
+  = CRZoneCreated {ident :: ZoneId}
+  | CRZoneList {zones :: [ZoneId]}
+  | CRZoneDetails {ident :: ZoneId, title :: Text, active :: Bool, kind :: ZoneKind}
```
epoberezkin, line 451: don't think you need to different responses - can just return list of objects, no need to get them one by one - just harder to use API. You need a type for this record.

src/Simplex/Chat/Migrations/M20230920_zones.hs
```diff
@@ -0,0 +1,21 @@
...
+module Simplex.Chat.Migrations.M20230920_zones where
+
+import Database.SQLite.Simple (Query)
+import Database.SQLite.Simple.QQ (sql)
+
+m20230920_zones :: Query
+m20230920_zones =
+  [sql|
+CREATE TABLE zones ( -- chat controller zones
+  zone_id INTEGER PRIMARY KEY,
```
epoberezkin, line 12: needs path...

PR #3100, opened 2023-09-22
core: optimize C apis

epoberezkin:


src/Simplex/Chat/Mobile/Shared.hs
```diff
@@ -1,19 +1,26 @@
...
+  pure $ BS fp $ fromIntegral len
 
 putByteString :: Ptr Word8 -> ByteString -> IO ()
-putByteString ptr bs@(PS fp offset _)  = do
-  let p = unsafeForeignPtrToPtr fp `plusPtr` offset
-  memcpy ptr p $ B.length bs
+putByteString ptr (BS fp len) =
+  withForeignPtr fp $ \p -> memcpy ptr p len
+
+newCStringFromBS :: ByteString -> IO CString
```
dpwiz, line 20: Why not [unsafePackCString](https://hackage.haskell.org/package/bytestring-0.12.0.2/docs/Data-ByteString-Unsafe.html#v:unsafePackCString) ?

epoberezkin: not sure why they do it like this everywhere? You mean to avoid any copying?

epoberezkin: but ByteString can be garbage collected before CString is used.

dpwiz: And the rest from Data.ByteString.Unsafe...

dpwiz: I see, the code gives out `Ptr`s for outside code.

PR #3104, opened 2023-09-22
core: remote host/controller types

dpwiz:


src/Simplex/Chat/Controller.hs
```diff
@@ -165,6 +166,8 @@ data ChatDatabase = ChatDatabase {chatStore :: SQLiteStore, agentStore :: SQLite
 
 data ChatController = ChatController
   { currentUser :: TVar (Maybe User),
+    remoteHosts :: TVar (Map RemoteHostId RemoteHostSession), -- All the active remote hosts
+    remoteController :: TMVar RemoteControllerSession, -- Supervisor process for hosted controllers
```
epoberezkin, line 170: why TMVar and not TVar Maybe?

src/Simplex/Chat/Controller.hs
```diff
@@ -193,6 +196,28 @@ data ChatController = ChatController
     logFilePath :: Maybe FilePath
   }
 
+data RemoteHostSession = RemoteHostSession
+  { -- | A process that relays the commands to its host
+    handler :: Async (),
+    -- | Path for local resources to be synchronized with host
+    path :: FilePath,
+    -- | Commands to be relayed to a host
+    command :: forall m . ChatMonad m => ByteString -> m ChatResponse,
```
epoberezkin, line 205: I don't think I'm a fan of this pattern when methods are captured as part of the state - why not just to pass session to the function that does it? This pattern feels opposite of functional, almost like doing Java in Haskell.

src/Simplex/Chat/Controller.hs
```diff
@@ -193,6 +196,28 @@ data ChatController = ChatController
...
+    fetchFile :: forall m . ChatMonad m => FilePath -> m ChatResponse,
+    -- | Store a file on a host to be sent
+    storeFile :: forall m . ChatMonad m => FilePath -> ByteString -> m ChatResponse
+  }
+
+-- | Host-side dual to RemoteHostSession, on-methods represent HTTP API.
+data RemoteControllerSession = RemoteControllerSession
+  { -- | A process that relays commands and data from remote controller
+    handler :: Async (),
+    onCommand :: forall m . ChatMonad m => ByteString -> m ChatResponse,
```
epoberezkin, line 216: same here - why they are not just functions?

epoberezkin: Lots of libraries are using and it makes it unnecessary hard to reason about what these functions do, as you need to find where this object is created instead of just finding the function by name. We should not use this pattern.

src/Simplex/Chat/Controller.hs
```diff
@@ -201,7 +226,20 @@ instance ToJSON HelpSection where
...
 
 data ChatCommand
-  = ShowActiveUser
+  = CreateRemoteHost -- ^ Configure a new remote host
+  | ListRemoteHosts
+  | StartRemoteHost RemoteHostId -- ^ Start and announce a remote host
+  | StopRemoteHost RemoteHostId -- ^ Shut down a running session
+  | DisposeRemoteHost RemoteHostId -- ^ Unregister remote host and remove its data
+  | StoreRemote FilePath RemoteHostId -- ^ Store a file at host, preparing for `SendFile`
+  | FetchRemote FilePath RemoteHostId -- ^ Fetch a file from host after `ReceiveFile`
```
epoberezkin, line 235: it's arguable whether these should be commands or just functions... We will see, ok to keep as commands for now.

epoberezkin: The reason why they should not be commands is because it would offload this responsibility to the UI (and we are back to doing it 4 times), while when desktop core receives send from UI it can automatically upload file to host before forwarding the actual send command, and likewise having received "file received" it may also automatically fetch it. Also, not sure why the actual file cannot be just attached as a body to the request or response and any commands are needed here at all.

Kind of convinced myself these commands are not needed, normally. The only case they may useful is when you load chat, and there is some file, and it's not present in the local storage. But again, the host could simply see it before sending the response, check if it has files locally, and send them with the response.

src/Simplex/Chat/Controller.hs
```diff
@@ -201,7 +226,20 @@ instance ToJSON HelpSection where
...
+  | RegisterRemoteController Text -- ^ Register OOB data for satellite discovery and handshake
+  | StartRemoteController -- ^ Start listening for satellite announcements
+  | ConfirmRemoteController -- ^ Confirm discovered data and store confirmation
+  | RejectRemoteController -- ^ Reject discovered data (and blacklist?)
+  | StopRemoteController -- ^ Stop listening for announcements or terminate an active session
+  | DisposeRemoteController Text -- ^ Remove all local data associated with a satellite session
```
epoberezkin, line 241: Above is not clear. There can be one remote controller (desktop) active at a time, but there can be multiple remote controllers registered with the host, so IDs seem missing in at least some commands. It's only when you add a new one you will get ID in response. Also there should be command for listing them and de-registering them. Above seems to be mixing sessions and known remote controllers.

src/Simplex/Chat/Controller.hs
```diff
@@ -420,7 +458,17 @@ data ChatCommand
...
+  = CRRemoteHostCreated {ident :: RemoteHostId, oobData :: Text}
+  | CRRemoteHostList {remoteHosts :: [(RemoteHostId, Text)]} -- XXX: RemoteHostInfo is mostly concerned with session setup
+  | CRRemoteHostStarted {ident :: RemoteHostId}
+  | CRRemoteHostStopped {ident :: RemoteHostId}
+  | CRRemoteHostDisposed {ident :: RemoteHostId}
+  | CRRemoteControllerRegistered
+  | CRRemoteControllerAccepted
+  | CRRemoteControllerRejected
+  | CRRemoteControllerConnected
+  | CRRemoteControllerDisconnected
```
epoberezkin, line 470: same here

src/Simplex/Chat/Controller.hs
```diff
@@ -967,6 +1017,21 @@ instance ToJSON SQLiteError where
 throwDBError :: ChatMonad m => DatabaseError -> m ()
 throwDBError = throwError . ChatErrorDatabase
 
+data RemoteError
```
epoberezkin, line 1020:
```suggestion
data RemoteHostError
```

src/Simplex/Chat/Migrations/M20230922_remote_hosts.hs
```diff
@@ -0,0 +1,22 @@
...
+
+m20230922_remote_hosts :: Query
+m20230922_remote_hosts =
+  [sql|
+CREATE TABLE remote_hosts (
+  zone_id INTEGER PRIMARY KEY,
+  display_name TEXT NOT NULL,
+  path TEXT NOT NULL,
+  ca_key BLOB NOT NULL,
+  ca_cert BLOB NOT NULL
```
epoberezkin, line 16: probably not all fields that are needed. We normally store public key separately, so you don't need to read private key when only public is needed, and the naming should be ca_private_key. Also you need the second key and cert, not to change TLS stack that expects 2-cert chain (with the fingerprint from the root).

src/Simplex/Chat/Mobile.hs
```diff
@@ -97,10 +99,18 @@ cChatMigrateInit fp key conf ctrl = do
 
 -- | send command to chat (same syntax as in terminal for now)
 cChatSendCmd :: StablePtr ChatController -> CString -> IO CJSONString
-cChatSendCmd cPtr cCmd = do
+cChatSendCmd cPtr = cChatSendCmdEx cPtr 0
+
+-- | send command to chat (same syntax as in terminal for now)
+cChatSendCmdEx :: StablePtr ChatController -> CInt -> CString -> IO CJSONString
+cChatSendCmdEx cPtr cZone cCmd = do
```
epoberezkin, line 106:
```suggestion
cChatSendCmdEx cPtr cRemoteHostId cCmd = do
```

src/Simplex/Chat/Types.hs
```diff
@@ -1527,3 +1532,24 @@ newtype JVersionRange = JVersionRange {fromJVersionRange :: VersionRange} derivi
 instance ToJSON JVersionRange where
   toJSON (JVersionRange (VersionRange minV maxV)) = J.object ["minVersion" .= minV, "maxVersion" .= maxV]
   toEncoding (JVersionRange (VersionRange minV maxV)) = J.pairs $ "minVersion" .= minV <> "maxVersion" .= maxV
+
+newtype RemoteHostId = RemoteHostId Int
```
epoberezkin, line 1536: why does it need to be newtype and not just synonym?

src/Simplex/Chat/Controller.hs
```diff
@@ -420,7 +458,17 @@ data ChatCommand
   deriving (Show)
 
 data ChatResponse
-  = CRActiveUser {user :: User}
+  = CRRemoteHostCreated {ident :: RemoteHostId, oobData :: Text}
```
epoberezkin, line 461:
```suggestion
  = CRRemoteHostCreated {remoteHostId :: RemoteHostId, oobData :: Text}
```

src/Simplex/Chat/Controller.hs
```diff
@@ -12,6 +12,7 @@
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE StrictData #-}
 {-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE RankNTypes #-}
```
epoberezkin, line 15: and you won't need this if that anti-pattern is removed.

src/Simplex/Chat/Controller.hs
```diff
@@ -617,8 +666,8 @@ logResponseToFile = \case
   _ -> False
 
 instance ToJSON ChatResponse where
-  toJSON = J.genericToJSON . sumTypeJSON $ dropPrefix "CR"
-  toEncoding = J.genericToEncoding . sumTypeJSON $ dropPrefix "CR"
+  toJSON = JT.genericToJSON . sumTypeJSON $ dropPrefix "CR"
+  toEncoding = JT.genericToEncoding . sumTypeJSON $ dropPrefix "CR"
```
epoberezkin, line 670:
```suggestion
  toJSON = J.genericToJSON . sumTypeJSON $ dropPrefix "CR"
  toEncoding = J.genericToEncoding . sumTypeJSON $ dropPrefix "CR"
```

epoberezkin: maybe I am missing why it was needed?

src/Simplex/Chat/Types.hs
```diff
@@ -1527,3 +1532,24 @@ newtype JVersionRange = JVersionRange {fromJVersionRange :: VersionRange} derivi
...
+  deriving stock (Show)
+  deriving newtype (Eq, Ord, FromJSON, ToJSON, FromField, ToField, StrEncoding)
+
+pattern LOCAL_HOST_ID :: Maybe RemoteHostId
+pattern LOCAL_HOST_ID = Nothing
+{-# COMPLETE LOCAL_HOST_ID, Just #-}
```
epoberezkin, line 1542: why can't we just use Nothing? seems confusing rather than helpful.

epoberezkin: We don't define what "Local host" means, and it kind of obvious that if remoteHostId is Nothing then it's not remote.

src/Simplex/Chat.hs
```diff
@@ -5355,7 +5390,20 @@ withStoreCtx ctx_ action = do
 chatCommandP :: Parser ChatCommand
 chatCommandP =
   choice
-    [ "/mute " *> ((`SetShowMessages` False) <$> chatNameP),
+    [ "/create remote host" $> CreateRemoteHost,
+      "/list remote hosts" $> ListRemoteHosts,
```
epoberezkin, line 5394: assuming that this will remove all know hosts with session statuses for each?

src/Simplex/Chat/Controller.hs
```diff
@@ -420,7 +458,17 @@ data ChatCommand
   deriving (Show)
 
 data ChatResponse
-  = CRActiveUser {user :: User}
+  = CRRemoteHostCreated {ident :: RemoteHostId, oobData :: Text}
+  | CRRemoteHostList {remoteHosts :: [(RemoteHostId, Text)]} -- XXX: RemoteHostInfo is mostly concerned with session setup
```
epoberezkin, line 462: but this seems to miss actual sessions - how would UI list sessions?

src/Simplex/Chat/Controller.hs
```diff
@@ -466,6 +514,7 @@ data ChatResponse
   | CRMsgIntegrityError {user :: User, msgError :: MsgErrorType}
   | CRCmdAccepted {corr :: CorrId}
   | CRCmdOk {user_ :: Maybe User}
+  | CRCmdRelayed -- ^ A command sent to a remote host instead of being interpreted locally
```
epoberezkin, line 517: why does UI need to know the difference? The response should be synchronous anyway, when it was.

src/Simplex/Chat/Mobile.hs
```diff
@@ -195,13 +205,19 @@ chatMigrateInit dbFilePrefix dbKey confirm = runExceptT $ do
...
 
-chatSendCmd :: ChatController -> ByteString -> IO JSONByteString
-chatSendCmd cc s = J.encode . APIResponse Nothing <$> runReaderT (execChatCommand s) cc
+-- XXX: duplicated multiple times in different APIs
+chatSendCmd :: ChatController -> Maybe RemoteHostId -> String -> IO JSONByteString
+chatSendCmd cc z s = J.encode . APIResponse Nothing z <$> runReaderT handler cc
+  where
+    handler :: B.ByteString -> ReaderT ChatController IO ChatResponse
+    handler = case z of
+      LOCAL_HOST_ID -> execChatCommand
```
epoberezkin, line 214:
```suggestion
      Nothing -> execChatCommand
```

src/Simplex/Chat/Mobile.hs
```diff
@@ -227,7 +243,7 @@ chatPasswordHash pwd salt = either (const "") passwordHash salt'
     salt' = U.decode salt
     passwordHash = U.encode . C.sha512Hash . (pwd <>)
 
-data APIResponse = APIResponse {corr :: Maybe CorrId, resp :: ChatResponse}
+data APIResponse = APIResponse {corr :: Maybe CorrId, zone :: Maybe RemoteHostId, resp :: ChatResponse}
```
epoberezkin, line 246:
```suggestion
data APIResponse = APIResponse {corr :: Maybe CorrId, remoteHostId :: Maybe RemoteHostId, resp :: ChatResponse}
```

src/Simplex/Chat/Controller.hs
dpwiz: The 3 commands without arguments are working on a current session, initiated by getting OOB from UI.

The de-registering command is `Dispose`. It works on a persisted state, thus needs an identifier.

src/Simplex/Chat/Controller.hs
dpwiz: The errors appear to be the same for the controller and host.

src/Simplex/Chat/Migrations/M20230922_remote_hosts.hs
dpwiz: I didn't encounter the need for public key. This pair only used to derive an session certificate which is ephemeral.

src/Simplex/Chat/Controller.hs
dpwiz: TBD. The only use for this right now is CA fingerprint.

PR #3129, opened 2023-09-26
core: add FromJSON instance to ChatResponse

dpwiz:


src/Simplex/Chat/Controller.hs
```diff
@@ -438,10 +445,10 @@ data ChatCommand
...
   = CRActiveUser {user :: User}
   | CRUsersList {users :: [UserInfo]}
-  | CRChatStarted
-  | CRChatRunning
-  | CRChatStopped
-  | CRChatSuspended
+  | CRChatStarted {_nullary :: Maybe Int}
+  | CRChatRunning {_nullary :: Maybe Int}
+  | CRChatStopped {_nullary :: Maybe Int}
+  | CRChatSuspended {_nullary :: Maybe Int}
```
dpwiz, line 451: `Maybe Void`?

src/Simplex/Chat/Messages/CIContent.hs
```diff
@@ -722,3 +696,31 @@ ciCallInfoText status duration = case status of
   CISCallProgress -> "in progress " <> durationText duration
   CISCallEnded -> "ended " <> durationText duration
   CISCallError -> "error"
+
+$(JQ.deriveJSON (enumJSON $ dropPrefix "CISCall") ''CICallStatus)
```
dpwiz, line 700: No need for ()s, just `$JQ.derive ...`

epoberezkin: It's a common style I think

src/Simplex/Chat/Controller.hs
dpwiz: It wouldn't. Void makes `Just` unrepresentable, making Nothing the only option (to be eliminated later).

src/Simplex/Chat/Types.hs
```diff
@@ -1550,6 +1574,24 @@ instance ToJSON ChatVersionRange where
...
+
+$(JQ.deriveJSON defOpts ''User)
+
+$(JQ.deriveJSON defOpts ''NewUser)
+
+$(JQ.deriveJSON defOpts ''UserInfo)
+
+$(JQ.deriveJSON defOpts ''Contact)
+
+$(JQ.deriveJSON defOpts ''ContactRef)
```
dpwiz, line 1597: Missing:
```
$(JQ.deriveJSON defOpts ''Notification)
```

PR #3142, opened 2023-09-28
core: set database journal mode to WAL

epoberezkin:


src/Simplex/Chat.hs
```diff
@@ -334,9 +334,13 @@ stopChatController ChatController {chatStore, smpAgent, agentAsync = s, sndFiles
...
-  when closeStore $ liftIO $ do
-    closeSQLiteStore chatStore
-    closeSQLiteStore $ agentClientStore smpAgent
+  liftIO $ do
+    let agentStore = agentClientStore smpAgent
+    checkpointSQLiteStore chatStore
+    checkpointSQLiteStore agentStore
+    when closeStore $ do
+      closeSQLiteStore chatStore
+      closeSQLiteStore agentStore
```
spaced4ndy, line 343: can't something occur between checkpoint and close?

spaced4ndy: also close already does checkpoint itself

PR #3146, opened 2023-09-29
core: close stores before import/delete/encryption operations to make compatible with windows, make encryption more resilient

epoberezkin:


src/Simplex/Chat/Archive.hs
```diff
@@ -94,53 +94,60 @@ copyDirectoryFiles fromDir toDir = do
 
 deleteStorage :: ChatMonad m => m ()
 deleteStorage = do
-  StorageFiles {chatDb, agentDb, filesPath} <- storageFiles
-  removeFile chatDb
-  removeFile agentDb
+  fs@StorageFiles {filesPath} <- storageFiles
```
spaced4ndy, line 97: we could catch exception in deleteStorage, currently it crashes the app if file doesn't exist

src/Simplex/Chat/Archive.hs
```diff
@@ -94,53 +94,60 @@ copyDirectoryFiles fromDir toDir = do
...
+  liftIO $ closeSQLiteStore `withStores` fs
+  removeFile `withDBs` fs
   mapM_ removePathForcibly filesPath
   tmpPath <- readTVarIO =<< asks tempDirectory
   mapM_ removePathForcibly tmpPath
 
 data StorageFiles = StorageFiles
   { chatDb :: FilePath,
-    chatEncrypted :: TVar Bool,
+    chatStore :: SQLiteStore,
```
spaced4ndy, line 106: if it's being refactored anyway, why keep chatDb, agentDb separately in StorageFiles

spaced4ndy: withDBs could still exist and take the from stores

PR #3166, opened 2023-10-03
desktop (windows): adapting Windows build to new libs

avently:


avently:
Included into https://github.com/simplex-chat/simplex-chat/pull/3169, so can be closed

PR #3168, opened 2023-10-03
ios: profile names (remove full name)

epoberezkin:


apps/ios/Shared/Views/Onboarding/CreateProfile.swift
```diff
@@ -9,175 +9,244 @@
...
-                        if canCreateProfile() { focusFullName = true }
-                        else { focusDisplayName = true }
-                    }
+            Group {
+                Text("Create your profile")
+                    .font(.largeTitle)
+                    .bold()
+                Text("Your profile, contacts and delivered messages are stored on your device.")
+                    .foregroundColor(.secondary)
+                Text("The profile is only shared with your contacts.")
```
spaced4ndy, line 91: users confuse it with contact address
```suggestion
                Text("The profile is only shared with your SimpleX Chat contacts.")
```

spaced4ndy: same for profile update

PR #3206, opened 2023-10-12
rfc: remote UI implementation

epoberezkin:


docs/rfcs/2023-10-12-remote-ui.md
```diff
@@ -0,0 +1,88 @@
...
+
+When mobile is connected to the remote desktop it will receive no events from the core (other than remote session connection status changes), and the core will accept no commands from the UI (other than to terminate the remote session).
+
+As currently mobile maintains the list of connection statuses for all profiles, this state will have to be maintained in Haskell to be able to send it to the remote desktop UI when it is connected. It will also be used to update mobile UI when control is returned to the mobile.
+
+The switching between remote host role and local UI role should prevent the possibility of any race conditions.
+
+To swith to remote host role:
+- UI: will show the screen with "connected" status and "cancel" button, with disabled sleep timer on iOS, as iOS app will have to stay in foregro. Android will be able to function in background in this state.
+- core: stop adding events to the output queues (via some flag).
```
dpwiz, line 21: We can't just discard events, can we? Better state this explicitly: "The events should go into the «remote» queue instead."

docs/rfcs/2023-10-12-remote-ui.md
```diff
@@ -0,0 +1,88 @@
...
+- UI: process remaining events in the output queue and stop receiving them.
```
dpwiz, line 22: Please clarify. Does "stop receiving them" only states the observation that no more local events would come? Or the app should stop issuing polling calls?

docs/rfcs/2023-10-12-remote-ui.md
```diff
@@ -0,0 +1,88 @@
...
+- core: stop sending events to and accepting commands from UI.
```
dpwiz, line 23: "stop sending events" looks already covered above by "stop adding events to the output queues".

What does it mean to stop accepting commands - is that an `ChatError` response like "pls no local commands, I'm busy with remote" for all the `chat_send_cmd` calls with suppressed commands?

docs/rfcs/2023-10-12-remote-ui.md
```diff
@@ -0,0 +1,88 @@
...
+- core: send current list of profiles, coversations, and connections statuses to the remote desktop.
+- core: start sending events to and accepting commands from remote desktop.
+- core: start adding events to remote output queue.
```
dpwiz, line 26: how "adding events to remote output queue" is different from "sending event to / remote desktop"?..

docs/rfcs/2023-10-12-remote-ui.md
epoberezkin: yes, the idea is to stop polling

docs/rfcs/2023-10-12-remote-ui.md
epoberezkin: > "stop sending events" looks already covered above by "stop adding events to the output queues".

"stop adding" means blocking them completely

> response like "pls no local commands, I'm busy with remote"

yes

PR #3211, opened 2023-10-12
core: track network statuses, use in commands/events

epoberezkin:


src/Simplex/Chat.hs
```diff
@@ -1688,6 +1693,8 @@ processChatCommand = \case
         (connId, cReq) <- withAgent $ \a -> createConnection a (aUserId user) True SCMInvitation Nothing subMode
         -- [incognito] reuse membership incognito profile
         ct <- withStore' $ \db -> createMemberContact db user connId cReq g m mConn subMode
+        -- TODO not sure it is correct to set connections status here?
```
spaced4ndy, line 1696: I think we did it in UI as a workaround for the fact it's not separately subscribed to avoid spinning circle on this contact

src/Simplex/Chat.hs
```diff
@@ -2687,12 +2695,30 @@ subscribeUserConnections onlyNeeded agentBatchSubscribe user@User {userId} = do
...
+            addStatus :: ConnId -> Contact -> [(AgentConnId, NetworkStatus)] -> [(AgentConnId, NetworkStatus)]
+            addStatus connId ct =
+              let ns = (contactAgentConnId ct, netStatus $ resultErr connId rs)
+              in (ns :)
+            netStatus :: Maybe ChatError -> NetworkStatus
+            netStatus = maybe NSConnected $ NSError . errorNetworkStatus
+            errorNetworkStatus :: ChatError -> String
+            errorNetworkStatus = \case
+              ChatErrorAgent (BROKER _ NETWORK) _ -> "network"
+              ChatErrorAgent (SMP SMP.AUTH) _ -> "contact deleted"
```
spaced4ndy, line 2719: in other places we don't write it as 100% conclusion
```suggestion
              ChatErrorAgent (SMP SMP.AUTH) _ -> "authorization (contact may be deleted)"
```

src/Simplex/Chat/Controller.hs
```diff
@@ -1044,6 +1050,27 @@ chatWriteVar :: ChatMonad' m => (ChatController -> TVar a) -> a -> m ()
...
+-- CRNetworkStatuses <- + CRContactSubSummary: setContactNetworkStatus(sub.contact, .connected) / processContactSubError(sub.contact, err)
+-- CRNetworkStatuses <- + CRContactSubError: processContactSubError(contact, chatError)
+-- keep <- + CRConnectedToGroupMember: m.setContactNetworkStatus(contact, .connected)
+-- CRNetworkStatus <- + CRContactsSubscribed: updateContactsStatus(contactRefs, status: .connected)
+-- CRNetworkStatus <- + CRContactsDisconnected: updateContactsStatus(contactRefs, status: .disconnected)
+
+-- + updateContactsStatus(): networkStatuses[c.agentConnId] = status
+-- + processContactSubError(): m.setContactNetworkStatus(contact, .error(err))
+
+-- Below seems pre-mature? maybe it was not connected yet? It will send contact connected and it will set the status, that will be spinning until then?
```
spaced4ndy, line 1071: possibly.. it may seem counterintuitive that user should send message when there's spinner

src/Simplex/Chat.hs
epoberezkin: it's what it is now in the UI, don't think it needs to change. I'd rather we do the opposite in the UI where it says something else.

src/Simplex/Chat/Controller.hs
epoberezkin: possibly

PR #3249, opened 2023-10-19
flake.nix: better cache discoverability

cab404:


shumvgolove:
This will hardcore nix config preferences and will break `build-android.sh`/F-Droid policy. Please see my original comment in https://github.com/simplex-chat/simplex-chat/commit/05c4a6c68297418226e3a1fa186925444e25482a :
> This change to flake.nix breaks build-android.sh script by forcing user
to input y/n. AFAIK, this cannot be automated and I rather not include
workarounds like piping "yes n | nix build ...".

cab404:
> Hey.
> 
> I'm not entirely against this change per se, but this will break [`build-android.sh`](https://github.com/simplex-chat/simplex-chat/blob/master/scripts/android/build-android.sh) script and [F-Droid policy](https://gitlab.com/fdroid/fdroiddata/-/merge_requests/11545#note_1059768446) about nix caches (see my original comment in [05c4a6c](https://github.com/simplex-chat/simplex-chat/commit/05c4a6c68297418226e3a1fa186925444e25482a)). I can include a workaround (`yes n | nix build ...`) in our Android build script to mitigate this, but could you clarify what's the point of including zw3rk cache repo, and what value it would bring? Haskell's builds are not deterministic anyway.

Hi!

A friend of mine encountered something that looked like a GHC rebuild — thus I helped them to configure this cache, and by extension posted it as a patch here.

I guess that it is not worth it to add nixConfig to core flake. if that adds requirement for such hacks. At least not before `--ignore-flake-substituters` will be added)

Thanks for the time!

PR #3275, opened 2023-10-24
core: derive JSON with TH

epoberezkin:


src/Simplex/Chat/Controller.hs
```diff
@@ -996,22 +906,18 @@ data CoreVersionInfo = CoreVersionInfo
     simplexmqVersion :: String,
     simplexmqCommit :: String
   }
-  deriving (Show, Generic, FromJSON)
```
epoberezkin, line 999: FromJSON instance is not used, removed

spaced4ndy (approved):

PR #3283, opened 2023-10-25
implement some of the robust discovery rfc

dpwiz:
- new announce and OOB types
- starting on random port
- socket/announce address matching


src/Simplex/Chat/Remote.hs
```diff
@@ -287,33 +289,36 @@ handleGetFile path reply = do
 -- TODO the problem with this code was that it wasn't clear where the recursion can happen,
 -- by splitting receiving and processing to two functions it becomes clear
 discoverRemoteCtrls :: ChatMonad m => TM.TMap C.KeyHash TransportHost -> m ()
-discoverRemoteCtrls discovered = Discovery.withListener $ receive >=> process
+discoverRemoteCtrls discovered = do
```
dpwiz, line 292: Needs to actually use annouce data like counters, addr etc.

Only sig is checked (via Encoding instance) and the old fingerprint-matching is used.

PR #3288, opened 2023-10-27
core: improve group link protocol (immediately establish group connection without first creating contact)

spaced4ndy:


src/Simplex/Chat/Controller.hs
```diff
@@ -474,6 +474,7 @@ data ChatResponse
   | CRUserContactLinkUpdated {user :: User, contactLink :: UserContactLink}
   | CRContactRequestRejected {user :: User, contactRequest :: UserContactRequest}
   | CRUserAcceptedGroupSent {user :: User, groupInfo :: GroupInfo, hostContact :: Maybe Contact}
+  | CRGroupConnecting {user :: User, groupInfo :: GroupInfo}
```
spaced4ndy, line 477: process same as CRUserAcceptedGroupSent in UI

PR #3290, opened 2023-10-27
android, desktop: block members

avently:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
```diff
@@ -1695,6 +1791,15 @@ data class ChatItem (
   }
 }
 
+enum class CIMergeCategory {
+  MemberConnected,
```
epoberezkin, line 1795: this can be removed, it's not used

PR #3323, opened 2023-11-07
ios: connect with contact via address (for preset simplex contact)

spaced4ndy:


apps/ios/Shared/Views/ChatList/ChatListNavLink.swift
```diff
@@ -411,6 +432,17 @@ struct ChatListNavLink: View {
                     .environment(\EnvironmentValues.refresh as! WritableKeyPath<EnvironmentValues, RefreshAction?>, nil)
             }
     }
+
+    private func connectContactViaAddress_(_ contact: Contact, _ incognito: Bool) {
+        Task {
+            let ok = await connectContactViaAddress(contact.contactId, true)
```
epoberezkin, line 438:
```suggestion
            let ok = await connectContactViaAddress(contact.contactId, incognito)
```

PR #3332, opened 2023-11-08
core: handle remote control session setup errors

dpwiz:


src/Simplex/Chat/Remote.hs
```diff
@@ -131,7 +131,13 @@ startRemoteHost' rh_ = do
   withRemoteHostSession_ rhKey $ maybe (Right ((), Just RHSessionStarting)) (\_ -> Left $ ChatErrorRemoteHost rhKey RHEBusy)
   ctrlAppInfo <- mkCtrlAppInfo
   (invitation, rchClient, vars) <- withAgent $ \a -> rcConnectHost a pairing (J.toJSON ctrlAppInfo) multicast
-  rhsWaitSession <- async $ waitForSession rhKey remoteHost_ rchClient vars
+  rhsWaitSession <- async $ do
+    currentKey <- newTVarIO rhKey
```
epoberezkin, line 135: it is probably better to create  it before async?

src/Simplex/Chat/Remote.hs
```diff
@@ -131,7 +131,13 @@ startRemoteHost' rh_ = do
...
+    waitForSession rhKey remoteHost_ currentKey vars `catchChatError` \err -> do
```
epoberezkin, line 136: and then put handler into function to make it more readable?

PR #3369, opened 2023-11-14
core: add remote controller discovery with multicast

dpwiz:


src/Simplex/Chat/Controller.hs
```diff
@@ -1061,6 +1061,8 @@ data RemoteCtrlError
   | RCEBadState -- ^ A session is in a wrong state for the current operation
   | RCEBusy -- ^ A session is already running
   | RCETimeout
+  | RCENoKnownControllers -- ^ No previously-contacted controllers to discover
+  | RCEControllerMismatch -- ^ Attempting to confirm a found controller with another ID
```
epoberezkin, line 1065: RCEBadState?

src/Simplex/Chat/Controller.hs
```diff
@@ -1076,6 +1078,10 @@ data ArchiveError
 -- | Host (mobile) side of transport to process remote commands and forward notifications
 data RemoteCtrlSession
   = RCSessionStarting
+  | RCSessionSearching
```
epoberezkin, line 1081: RCSessionDiscovery?

src/Simplex/Chat/Controller.hs
```diff
@@ -1101,6 +1107,7 @@ data RemoteCtrlSession
 
 data RemoteCtrlSessionState
   = RCSStarting
+  | RCSSearching
```
epoberezkin, line 1110: same

src/Simplex/Chat/Remote.hs
```diff
@@ -342,19 +341,58 @@ liftRH rhId = liftError (ChatErrorRemoteHost (RHId rhId) . RHEProtocolError)
 
 -- * Mobile side
 
+-- ** Multicast
+
 findKnownRemoteCtrl :: ChatMonad m => m ()
-findKnownRemoteCtrl = undefined -- do
+findKnownRemoteCtrl = handleCtrlError "findKnownRemoteCtrl" $ do
+  knownCtrls <- withStore' getRemoteCtrls
+  pairings <- maybe (throwError $ ChatErrorRemoteCtrl RCENoKnownControllers) (pure . fmap (\RemoteCtrl {ctrlPairing} -> ctrlPairing)) $ nonEmpty knownCtrls
```
epoberezkin, line 349: some mess here a bit...

src/Simplex/Chat/Remote.hs
```diff
@@ -342,19 +341,58 @@ liftRH rhId = liftError (ChatErrorRemoteHost (RHId rhId) . RHEProtocolError)
...
+  withRemoteCtrlSession_ $ maybe (Right ((), Just RCSessionStarting)) (\_ -> Left $ ChatErrorRemoteCtrl RCEBusy)
+  foundCtrl <- newEmptyTMVarIO
+  cmdOk <- newEmptyTMVarIO
+  action <- async $ handleCtrlError "findKnownRemoteCtrl.discover" $ do
+    atomically $ takeTMVar cmdOk
+    (RCCtrlPairing {ctrlFingerprint}, inv) <- timeoutThrow (ChatErrorRemoteCtrl RCETimeout) (networkIOTimeout * 4) . withAgent $ \a -> rcDiscoverCtrl a pairings
```
epoberezkin, line 355: let's just make another constant - no need to couple them

src/Simplex/Chat/Remote.hs
```diff
@@ -342,19 +341,58 @@ liftRH rhId = liftError (ChatErrorRemoteHost (RHId rhId) . RHEProtocolError)
...
+    rc_ <- withStore' $ \db -> getRemoteCtrlByFingerprint db ctrlFingerprint
+    rc <- maybe (throwChatError $ CEInternalError "connecting with a stored ctrl") pure rc_
```
epoberezkin, line 357: this maybe (throw) is hard to read... I thin it's better just use \case in this case, without moving Maybe to ExceptT territory, they are different kind of error.

```suggestion
    withStore' (`getRemoteCtrlByFingerprint` ctrlFingerprint) >>= \case
      Nothing -> throwChatError $ CEInternalError "connecting with a stored ctrl"
      Just rc -> do
```

same as above

src/Simplex/Chat/Remote.hs
```diff
@@ -554,11 +576,12 @@ verifyRemoteCtrlSession execChatCommand sessCode' = handleCtrlError "verifyRemot
...
 
-handleCtrlError :: ChatMonad m => Text -> m a -> m a
-handleCtrlError name action = action `catchChatError` \e -> do
-  logError $ name <> " remote ctrl error: " <> tshow e
-  cancelActiveRemoteCtrl True
-  throwError e
+handleCtrlError :: (HasCallStack, ChatMonad m) => Text -> m a -> m a
+handleCtrlError name action =
+  action `catchChatError` \e -> do
+    withFrozenCallStack $ logError $ name <> " remote ctrl error: " <> tshow e
```
epoberezkin, line 582: what's that

src/Simplex/Chat/Controller.hs
dpwiz: It was Discovery, but View displays it as "(searching)". Renamed to match other states in "-ing" form.

src/Simplex/Chat/Remote.hs
dpwiz: fixing provenance in log messages. `handleCtrlError` is used multiple times, but source position tells the message comes from `handleCtrlError` itself which is useless.

this actually obsoletes the `name` argument.

PR #3378, opened 2023-11-15
desktop: add image pasting from clipboard

jessehorne:
The following PR adds the ability, on Desktop, to paste images from the clipboard into the send message input box.

Tested On:
1. Ubuntu

Supported image formats (all I've tested so far):
1. png
2. jpg
3. webp
4. bmp

Unsupported image formats:
1. svg
2. gif (It displays but does not transition between frames)


avently:
Added some optimizations to the code. We have processPickedMedia function that do some more checks like file size and correctness of the image. Also turns out previous logic of getting image from clipboard takes 39 seconds on my machine (this image: https://miro.medium.com/v2/resize:fit:4800/format:webp/1*gbpDpWISH80JkVzmQnqWvQ.jpeg). Changes I made here make loading of it faster: 8 seconds in total. I checked on ArchLinux

avently:
Checked on Windows too. Same image is loading for about 4 seconds.

@jessehorne, thank you for this contribution!

PR #3406, opened 2023-11-20
Update app-settings.md

qvsojBJGiEnR:
https://github.com/simplex-chat/simplex-chat/pull/3379#issue-1996386175


epoberezkin:
> One question: How can I update this PR before it gets merged?

you can make additional commits (or if you edit via UI there will be an option to update the same branch rather than create a new one)

PR #3431, opened 2023-11-22
core: shuffle members when sending messages and introductions; send to admins and owners first

spaced4ndy:


src/Simplex/Chat/Util.hs
```diff
@@ -30,3 +33,11 @@ encryptFile fromPath toPath cfArgs = do
...
 chunkSize = 65536
 {-# INLINE chunkSize #-}
+
+shuffle :: [a] -> IO [a]
+shuffle xs = do
+  randomNumbers <- mapM (\_ -> randomRIO (0, 65535) :: IO Word16) xs
+  let zipped = zip randomNumbers xs
+      sorted = sortOn fst zipped
+      shuffled = map snd sorted
+  pure shuffled
```
epoberezkin, line 43:
```suggestion
  pure $ map snd . sortOn fst . zip randomNumbers xs
```

src/Simplex/Chat/Util.hs
```diff
@@ -30,3 +33,11 @@ encryptFile fromPath toPath cfArgs = do
 chunkSize :: Num a => a
 chunkSize = 65536
 {-# INLINE chunkSize #-}
+
+shuffle :: [a] -> IO [a]
+shuffle xs = do
+  randomNumbers <- mapM (\_ -> randomRIO (0, 65535) :: IO Word16) xs
+  let zipped = zip randomNumbers xs
+      sorted = sortOn fst zipped
```
dpwiz, line 41: `sortBy (comparing fst)`?

`sortOn` does extra wrapping/unwrapping for no gain as the keys are numbers already.

dpwiz: why not decorate original data while generating randoms?

epoberezkin: it seems the opposite: https://hackage.haskell.org/package/base-4.19.0.0/docs/Data-List.html#v:sortOn

[sortOn](https://hackage.haskell.org/package/base-4.19.0.0/docs/Data-List.html#v:sortOn) f is equivalent to [sortBy](https://hackage.haskell.org/package/base-4.19.0.0/docs/Data-List.html#v:sortBy) ([comparing](https://hackage.haskell.org/package/base-4.19.0.0/docs/Data-Ord.html#v:comparing) f), but has the performance advantage of only evaluating f once for each element in the input list.

epoberezkin: I think I understand what you say now - it's better to evaluate fst twice than to move to another tuple... Interesting.

PR #3440, opened 2023-11-23
desktop: close remote host connecting screen on stop of host

avently:


dpwiz (approved):
This works.

But the "disconnect" message could be better. It says "device disconnected", but it wasn't connected in a first place.

avently:
> but it wasn't connected in a first place.

How backend knows that it needs to show Verify code screen? Because a mobile device connected to it. How it knows that "remoteHostStopped"? Because mobile device "disconnected" from it. 

As a user, why would you understand the meaning of the phrase "device disconnected"? You clicked on button that is called "Disconnect" or on a button that shows "wifi off" icon, which means disconnect.

PR #3464, opened 2023-11-25
desktop: use dropdowns for group setting message duration instead of wheel picker

jessehorne:
duration


jessehorne:
I think the text should be bigger for the dropdowns.

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/helpers/CustomTimePicker.kt
```diff
@@ -183,18 +174,41 @@ fun CustomTimePickerDialog(
...
-            timeUnitsLimits
-          )
 
-          TextButton(onClick = { confirmButtonAction(selection.value) }) {
-            Text(
-              confirmButtonText,
-              fontSize = 18.sp,
-              color = MaterialTheme.colors.primary
-            )
+          fun getUnitValues(unit: CustomTimeUnit, selectedValue: Int): List<Int> {
```
epoberezkin, line 178: please move it to avoid diff

epoberezkin:
let's please move it within the boundaries of app window and avoid rectangular white background that sticks behind rounded corners

epoberezkin:
... so it looks like the first menu:

image

PR #3465, opened 2023-11-25
desktop: remember window position and size

jessehorne:


apps/multiplatform/common/src/desktopMain/kotlin/chat/simplex/common/DesktopApp.kt
```diff
@@ -31,10 +30,31 @@ import java.io.File
 val simplexWindowState = SimplexWindowState()
 
 fun showApp() = application {
+  val state = getStoredWindowState() // creates file if not exists; comes with proper defaults
+
   // For some reason on Linux actual width will be 10.dp less after specifying it here. If we specify 1366,
   // it will show 1356. But after that we can still update it to 1366 by changing window state. Just making it +10 now here
-  val width = if (desktopPlatform.isLinux()) 1376.dp else 1366.dp
-  val windowState = rememberWindowState(placement = WindowPlacement.Floating, width = width, height = 768.dp)
```
avently, line 37: since you moved the code, move the comment too

apps/multiplatform/common/src/desktopMain/kotlin/chat/simplex/common/DesktopApp.kt
```diff
@@ -31,10 +30,31 @@ import java.io.File
...
+    height = state.height.dp,
+    position = WindowPosition(state.x.dp, state.y.dp)
+  )
+
+  LaunchedEffect(
+    windowState.position.x.value,
+    windowState.position.y.value,
+    windowState.size.width.value,
+    windowState.size.height.value
+  ) {
```
avently, line 49: will it still work if you just use 
```kotlin 
KeyChangeEffect(windowState) {
...
}
```
? It should, because WindowPositionSize is a data class.

apps/multiplatform/common/src/desktopMain/kotlin/chat/simplex/common/DesktopApp.kt
```diff
@@ -31,10 +30,31 @@ import java.io.File
 val simplexWindowState = SimplexWindowState()
 
 fun showApp() = application {
+  val state = getStoredWindowState() // creates file if not exists; comes with proper defaults
```
avently, line 33: Better to place a comment on top of the code line to make it more visible and consistent with other comments in codebase

PR #3467, opened 2023-11-25
desktop: remove slash from update member role alert in group settings

jessefromearth:
`The role will be changed to \"%s\".`

becomes

`The role will be changed to %s.`


epoberezkin:
The intention was to have it like this:

The role will be changed to "%s".

That is, in quotes. As otherwise it blends in into the text.

PR #3471, opened 2023-11-26
add remote host bindings

dpwiz:


src/Simplex/Chat/Controller.hs
```diff
@@ -426,7 +427,7 @@ data ChatCommand
   | SetGroupTimedMessages GroupName (Maybe Int)
   | SetLocalDeviceName Text
   | ListRemoteHosts
-  | StartRemoteHost (Maybe (RemoteHostId, Bool)) -- Start new or known remote host with optional multicast for known host
+  | StartRemoteHost (Maybe (RemoteHostId, Bool)) (Maybe Text) (Maybe TransportHost) (Maybe Word16) -- Start new or known remote host with optional multicast for known host
```
epoberezkin, line 430:
```
| StartRemoteHost {remoteHost_ :: Maybe (RemoteHostId, Bool), address :: Maybe RemoteCtrlAddress, port :: Maybe Word16}

data RemoteCtrlAddress = RemoteCtrlAddress {interface: String, address: String}
```

Then UI can pass the whole structure as is.

network-info returns String as interface, why does it need to be Text here? I don't think explicit usage of Text is justified, for short strings it can be counterproductive

src/Simplex/Chat/View.hs
```diff
@@ -1713,8 +1713,12 @@ viewRemoteHosts = \case
   [] -> ["No remote hosts"]
   hs -> "Remote hosts: " : map viewRemoteHostInfo hs
   where
-    viewRemoteHostInfo RemoteHostInfo {remoteHostId, hostDeviceName, sessionState} =
-      plain $ tshow remoteHostId <> ". " <> hostDeviceName <> maybe "" viewSessionState sessionState
+    viewRemoteHostInfo RemoteHostInfo {remoteHostId, hostDeviceName, sessionState, bindAddress_, bindPort_} =
+      plain $ tshow remoteHostId <> ". " <> hostDeviceName <> todo'viewBinds bindAddress_ bindPort_ <> maybe "" viewSessionState sessionState
+    todo'viewBinds rca_ port_ =
+      if isJust rca_ || isJust port_
+        then " [addr=127.0.0.1 iface=lo port=52230]" -- alternatively: bind=lo,127.0.0.1:52230
```
epoberezkin, line 1720:
```suggestion
        then " [addr=127.0.0.1 iface=lo port=52230]" -- alternatively: bind=lo,127.0.0.1:52230
```

maybe just 127.0.0.1:52230

PR #3481, opened 2023-11-28
desktop: select text from chat items

jessefromearth:
https://github.com/simplex-chat/simplex-chat/assets/2800532/c45995d5-a80c-4cd9-96d1-5783c290907e

In this PR I've done the following and limited the changes to desktop...

1) Disabled dragging of chat items to reply
2) You can select text now from chat items, including the reply headers
3) You can still right click anywhere on the chat item to open up the menu


epoberezkin:
Context menu is almost impossible to show with this change, it should capture long clicks on text too.

Issue #3350, opened 2023-11-12
[Bug]: Missing scrollbar


sectore:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Platform

Linux

### OS version

Ubuntu 22.04

### App version

v5.4.0-beta.3

### Current Behavior

Content can be scrolled by mouse, but there is no scrollbar available.

In main view (chat message history), side panel (chat group selector, settings), console.

Not sure if this a missing feature of the GUI or a bug. For the former, change bug -> feature request.

### Expected Behavior

Scroll bar should appear for content higher than window height.

### Steps To Reproduce

Open a group chat with a long chat history.

PR #3484, opened 2023-11-28
desktop: add vertical scrollbar to chat list

jessefromearth:
This PR adds a scroll bar to the chat list. If we merge this we can close issue #3350


epoberezkin:
It's only added to chat view, where it was less of a problem, not to chat list.

It needs to be added to all other scrollable windows then too - it's a lot of them, throughout the app.

Also, don't think the layout in chat list will accommodate scroll bar (although maybe it'll get wider if you add it?).

PR #3498, opened 2023-12-01
desktop: allow changing listening ip and port of remote

avently:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/SimpleXAPI.kt
```diff
@@ -4354,6 +4362,8 @@ data class SomeRemoteCtrl(
   val appVersion: String
 )
 
+
+
```
epoberezkin, line 4366:
```suggestion
```

epoberezkin, line 4366:
```suggestion
```

PR #3505, opened 2023-12-03
core: chat list pagination

dpwiz:


src/Simplex/Chat.hs
```diff
@@ -1821,9 +1821,8 @@ processChatCommand = \case
     quotedItemId <- withStore $ \db -> getGroupChatItemIdByText db user groupId cName quotedMsg
     let mc = MCText msg
     processChatCommand . APISendMessage (ChatRef CTGroup groupId) False Nothing $ ComposedMessage Nothing (Just quotedItemId) mc
-  LastChats count_ -> withUser' $ \user -> do
-    chats <- withStore' $ \db -> getChatPreviews db user False
-    pure $ CRChats $ maybe id take count_ chats
+  LastChats count_ -> withUser' $ \user ->
+    CRChats <$> withStore' (\db -> getChatPreviews db user False (CPLastTs <$> count_))
```
epoberezkin, line 1825: 👍

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -563,94 +584,56 @@ getDirectChatPreviews_ db user@User {userId} = do
             )
             OR c.connection_id IS NULL
           )
+          AND (
+            ct.local_display_name LIKE '%' || :search || '%'
+            OR cp.display_name LIKE '%' || :search || '%'
+            OR cp.full_name LIKE '%' || :search || '%'
+            OR cp.local_alias LIKE '%' || :search || '%'
+            OR c.local_alias LIKE '%' || :search || '%'
```
spaced4ndy, line 592: Need to double check, but connection alias is only used until contact profile is received, after which it's transferred onto profile and afterwards alias is only updated for profile. So likely search shouldn't be made by connection alias because it could've changed on profile by then. (Connection alias may still have to be used in search when loading pending contact connections previews)

src/Simplex/Chat/Controller.hs
```diff
@@ -247,7 +247,7 @@ data ChatCommand
   | ExecChatStoreSQL Text
   | ExecAgentStoreSQL Text
   | SlowSQLQueries
-  | APIGetChats {userId :: UserId, pendingConnections :: Bool}
+  | APIGetChats {userId :: UserId, pendingConnections :: Bool, paginationTs_ :: Maybe ChatPaginationTs, favorite :: Bool, unread :: Bool, search_ :: Maybe String}
```
spaced4ndy, line 250: search doesn't take into account filters so this has higher dimensionality than necessary (you can't search **and** filter, but api suggests you can), maybe time to make a type for this:

```haskell
data ChatListQuery
  = CLQNoFilter
  | CLQSearch String
  | CLQFilter ChatListFilters
  -- or for current UI the last one could be simply
  -- | CLQFavoriteOrUnread
  -- and if separate are allowed later we could then add
  -- | CLQFavorite
  -- | CLQUnread

data ChatListFilters = ChatListFilters
  { favorite :: Bool,
    unread :: Bool
  }
```

actually even CLQNoFilter and CLQFilter could be united and if all filters in ChatListFilters are false then it's "no filter"

spaced4ndy: maybe it's a bad idea if we plan to support search while filtering later

src/Simplex/Chat/Messages.hs
```diff
@@ -719,6 +719,12 @@ data ChatPagination
   | CPBefore ChatItemId Int
   deriving (Show)
 
+data ChatPaginationTs
+  = CPLastTs Int
+  | CPAfterTs ChatItemTs Int
+  | CPBeforeTs ChatItemTs Int
```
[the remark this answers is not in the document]
spaced4ndy: it may be not chat list specific but then it could be called Pagination instead of ChatPagination, and chat pagination already exists and is done for items inside chat by id, not timestamp, so it's somewhat misleading. Actually even synonym ChatItemTs is incorrect here semantically because chat list is paginated by chat timestamp, not chat item timestamp - it could be simply UTCTime.

Issue #3469, opened 2023-11-26
[Bug]: Cabal install error on rPI4


cthulhubuddha:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Platform

Linux

### OS version

PRETTY_NAME="Debian GNU/Linux 11 (bullseye)" NAME="Debian GNU/Linux" VERSION_ID="11" VERSION="11 (bullseye)" VERSION_CODENAME=bullseye ID=debian HOME_URL="https://www.debian.org/" SUPPORT_URL="https://www.debian.org/support" BUG_REPORT_URL="https://bugs.debian.org/"

### App version

Latest stable

### Current Behavior

Compiling from source and following these instructions - https://github.com/simplex-chat/simplex-chat/blob/stable/docs/CLI.md

Everything compiles as expected, but on the final `cabal install simplex-chat` it comes back with

```
~/simplex-chat# cabal install simplex-chat
cabal: Ambiguous target 'simplex-chat'. It could be:
exe:simplex-chat (component)
lib:simplex-chat (component)
```

### Expected Behavior

It should install correctly

### Steps To Reproduce

On Rpi4 bullseye:

```
sudo apt install build-essential curl libffi-dev libffi7 libgmp-dev libgmp10 libncurses-dev libncurses5 libtinfo5
curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
git clone https://github.com/simplex-chat/simplex-chat
cd simplex-chat
git checkout stable
apt-get update && apt-get install -y build-essential libgmp3-dev zlib1g-dev
cp scripts/cabal.project.local.linux cabal.project.local
cabal update
cabal install simplex-chat
```

dpwiz:
> cabal: Ambiguous target 'simplex-chat'. It could be:

Hm.. Looks like there's some difference between `cabal-install` versions.

Try giving it what it asks for:
```
cabal install simplex-chat:exe:simplex-chat
```

cthulhubuddha:
That did the trick! Now running into cabal and ghc version issues, but those are different issues. Thanks!

eyeinsky:
Made PR here https://github.com/simplex-chat/simplex-chat/pull/3528

[...]

PR #3528, opened 2023-12-07
Add component name to install command (#3469)

eyeinsky:


epoberezkin:
it's probably related to https://github.com/simplex-chat/simplex-chat/pull/3559, and it doesn't fix that problem too, sadly...

PR #3529, opened 2023-12-08
android, desktop: rework UX of creating new connection

avently:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ChatListView.kt
```diff
@@ -298,15 +313,120 @@ fun connectIfOpenedViaUri(rhId: Long?, uri: URI, chatModel: ChatModel) {
...
+        hideSearchOnBack()
+      }
+    } else {
+      Row {
+        val padding = if (appPlatform.isDesktop) 0.dp else 7.dp
+        val clipboard = LocalClipboardManager.current
+        val clipboardHasText = remember(focused) { chatModel.clipboardHasText }.value
+        if (clipboardHasText) {
+          IconButton(
+            onClick = { searchText.value = searchText.value.copy(clipboard.getText()?.text ?: return@IconButton) },
```
spaced4ndy, line 351: when pasting a link to a known contact or group, it's filtered in chat list for a moment, then search is cleared and filter is reset.

This happens with links to others' contacts (known via address) and groups (known via group link); link to own group correctly stays filtered.

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/newchat/NewChatView.kt
```diff
@@ -0,0 +1,435 @@
...
+@Composable
+private fun InviteView(rhId: Long?, connReqInvitation: String, contactConnection: MutableState<PendingContactConnection?>) {
+  SectionView(stringResource(MR.strings.share_this_1_time_link).uppercase()) {
+    LinkTextView(connReqInvitation, true)
+  }
+
+  Spacer(Modifier.height(10.dp))
+
+  SectionView(stringResource(MR.strings.or_show_this_qr_code).uppercase()) {
+    SimpleXLinkQRCode(connReqInvitation)
```
spaced4ndy, line 208: when qr code is shared, it shouldn't offer to delete unused invitation

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/newchat/NewChatView.kt
```diff
@@ -0,0 +1,435 @@
...
+  }
+
+  Column(
+    Modifier.fillMaxSize(),
+  ) {
+    Box(contentAlignment = Alignment.Center) {
+      val bottomPadding = DEFAULT_PADDING
+      AppBarTitle(stringResource(MR.strings.new_chat), hostDevice(rh?.remoteHostId), bottomPadding = bottomPadding)
+      Column(Modifier.align(Alignment.CenterEnd).padding(bottom = bottomPadding, end = DEFAULT_PADDING)) {
+        AddContactLearnMoreButton()
```
spaced4ndy, line 106: when an info page is closed, it always goes back to the view that was initially shown, not the view that was shown before opening info:

meaning if you enter via "Add contact", go to Connect via link tab, open info, then close info - New chat would show Add contact tab; and vice versa if you open via scan qr code button in search, go to Add contact tab, open then close info - New chat would show Connect via link tab.

Same goes for Incognito info when opening via qr code.

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ChatListView.kt
```diff
@@ -298,15 +313,120 @@ fun connectIfOpenedViaUri(rhId: Long?, uri: URI, chatModel: ChatModel) {
...
 }
 
+@Composable
+private fun ChatListSearchBar(listState: LazyListState, searchText: MutableState<TextFieldValue>, searchShowingSimplexLink: MutableState<Boolean>, searchChatFilteredBySimplexLink: MutableState<String?>) {
+  Row(verticalAlignment = Alignment.CenterVertically) {
+    val focusRequester = remember { FocusRequester() }
+    var focused by remember { mutableStateOf(false) }
+    Icon(painterResource(MR.images.ic_search), null, Modifier.padding(horizontal = DEFAULT_PADDING_HALF), tint = MaterialTheme.colors.secondary)
+    SearchTextField(
+      Modifier.weight(1f).onFocusChanged { focused = it.hasFocus }.focusRequester(focusRequester),
```
spaced4ndy, line 328: closing keyboard doesn't remove cursor from search, considering search can't even be hidden it's quite annoying in chat list.. is there an easy fix possibly?

avently:
@spaced4ndy  all done

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/newchat/ScanToConnectView.kt
```diff
@@ -1,57 +1,51 @@
 package chat.simplex.common.views.newchat
```
spaced4ndy, line 1: should be good to move to NewChatView, or do you want to do it in a separate pr to keep diff?

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/newchat/NewChatView.kt
```diff
@@ -0,0 +1,435 @@
...
+  if (appPlatform.isAndroid) {
+    Spacer(Modifier.height(10.dp))
+
+    SectionView(stringResource(MR.strings.or_scan_qr_code).uppercase()) {
+      QRCodeScanner(showQRCodeScanner) { text ->
+        withBGApi {
+          val res = verify(rhId, text, close)
+          if (!res) {
+            AlertManager.shared.showAlertMsg(
+              title = generalGetString(MR.strings.incorrect_code)
```
spaced4ndy, line 268: reuses key from VerifyCodeView, should say something about link instead, e.g.
title: "Invalid QR code", message: "The code you scanned is not a SimpleX link QR code."

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/newchat/ScanToConnectView.kt
avently: I would prefer we both to have this code separated from other classes. Like ConnectPlan.{kt,swift}. What do you think?
- it's unrelated to this PR and to its idea
- it has some logic nobody really cares about
- it has no UI, just logic.

Let's just move it to separate file (in my case I can just rename it, in you case you can move the whole big part of code to the file ConnectPlan.swift).

I don't really like to mix such big parts of UI and logic into one SQLite-like file

spaced4ndy: Ok, fine to rename too, or even keep as is, doesn't matter much. In swift it's slightly more tied to views with those ugly Alert and Sheet enums, but frankly I just don't want to bother moving second time

avently: Ah, ok, didn't think about how tied they are. Let me rename my file than and will keep it like this

avently: renamed

PR #3535, opened 2023-12-10
ios: support fully disabling notifications

epoberezkin:


apps/ios/Shared/Model/SuspendChat.swift
```diff
@@ -103,11 +104,32 @@ func activateChat(appState: AppState = .active) {
 func initChatAndMigrate(refreshInvitations: Bool = true) {
     let m = ChatModel.shared
     if (!m.chatInitialized) {
+        m.v3DBMigration = v3DBMigrationDefault.get()
+        if AppChatState.shared.value == .stopped {
+            AlertManager.shared.showAlert(Alert(
+                title: Text("Start chat?"),
+                message: Text("Chat is stopped. Do not start chat if you already used this database on another device."),
```
spaced4ndy, line 111:
```suggestion
                message: Text("Chat is stopped. If you already used this database on another device, you should transfer it back before starting chat."),
```

spaced4ndy (approved):

PR #3546, opened 2023-12-13
directory: better search, allow both simplex:/ and simplex.chat links in description

epoberezkin:


apps/simplex-directory-service/src/Directory/Events.hs
```diff
@@ -120,14 +132,20 @@ deriving instance Show ADirectoryCmd
 
 directoryCmdP :: Parser ADirectoryCmd
 directoryCmdP =
-  (A.char '/' *> cmdStrP) <|> (ADC SDRUser . DCSearchGroup <$> A.takeText)
+  (A.char '/' *> cmdStrP)
+    <|> (A.char '.' $> ADC SDRUser DCSearchNext)
```
spaced4ndy, line 136: why special case it.. it's not a normal user friendly interface anyway

PR #3561, opened 2023-12-15
ios: delay suspendChat in NSE, background schedule depends on notifications mode

epoberezkin:


apps/ios/Shared/AppDelegate.swift
```diff
@@ -81,7 +81,7 @@ class AppDelegate: NSObject, UIApplicationDelegate {
                 }
             } else if let checkMessages = ntfData["checkMessages"] as? Bool, checkMessages {
                 logger.debug("AppDelegate: didReceiveRemoteNotification: checkMessages")
-                if  m.ntfEnablePeriodic && allowBackgroundRefresh() {
+                if  m.ntfEnablePeriodic && allowBackgroundRefresh() &&  BGManager.shared.shouldRun {
```
spaced4ndy, line 84:
```suggestion
                if  m.ntfEnablePeriodic && allowBackgroundRefresh() && BGManager.shared.shouldRun {
```

apps/ios/SimpleX NSE/NotificationService.swift
```diff
@@ -326,7 +332,13 @@ class NotificationService: UNNotificationServiceExtension {
...
             threadId = nil
             if NSEThreads.shared.endThread(t) {
-                suspendChat(nseSuspendTimeout)
+                logger.debug("NotificationService.deliverBestAttemptNtf: will suspend")
+                DispatchQueue.global().asyncAfter(deadline: .now() + nseSuspendDelay) {
+                    if NSEThreads.shared.noThreads {
+                        logger.debug("NotificationService.deliverBestAttemptNtf: suspending...")
+                        suspendChat(nseSuspendTimeout)
+                    }
+                }
```
spaced4ndy, line 341: could check noThreads first, then delay and check again. It's probably unnecessary to delay if there's already noThreads

apps/ios/Shared/Model/BGManager.swift
```diff
@@ -34,35 +39,49 @@ class BGManager {
...
         if !ChatModel.shared.ntfEnableLocal {
             logger.debug("BGManager.handleRefresh: disabled")
             return
         }
         logger.debug("BGManager.handleRefresh")
-        schedule()
-        if allowBackgroundRefresh() {
+        let shouldRun_ = shouldRun
+        if allowBackgroundRefresh() && shouldRun_ {
+            schedule()
```
spaced4ndy, line 77: I don't fully understand logic here.. first, why not pass directly `interval: TimeInterval?` instead of `soon` to `schedule`? It seems as unnecessary indirection as in the else below `soon` itself depends on `runInterval`.. Second, why not always pass it to `schedule`? In if it would be `schedule(runInverval)`, in else - `schedule(shouldRun ? bgRefreshInterval : runInterval)`.

schedule could actually be even moved outside of case it seems, as it's being called in both cases now. interval parameter would be:

| allowBackgroundRefresh() && shouldRun -> runInterval (also, why not bgRefreshInterval?)
| shouldRun -> bgRefreshInterval
| otherwise -> runInterval

apps/ios/Shared/Model/BGManager.swift
```diff
@@ -34,35 +39,49 @@ class BGManager {
...
 
+    var runInterval: TimeInterval {
+        switch ChatModel.shared.notificationMode {
+        case .instant: maxBgRefreshInterval
+        case .periodic: periodicBgRefreshInterval
+        case .off: bgRefreshInterval
+        }
+    }
+
+    var shouldRun: Bool {
```
spaced4ndy, line 65:
```suggestion
    var lastRanLongAgo: Bool {
```

`shouldRun` isn't very descriptive name

apps/ios/SimpleX NSE/NotificationService.swift
epoberezkin: add comment

PR #3567, opened 2023-12-17
core: allow deleting last user

epoberezkin:


tests/ChatClient.hs
```diff
@@ -276,15 +276,15 @@ getTermLine cc =
   5000000 `timeout` atomically (readTQueue $ termQ cc) >>= \case
     Just s -> do
       -- remove condition to always echo virtual terminal
-      -- when True $ do
-      when (printOutput cc) $ do
+      when True $ do
+      -- when (printOutput cc) $ do
```
spaced4ndy, line 280:
```suggestion
      -- when True $ do
      when (printOutput cc) $ do
```

tests/ChatTests/Direct.hs
```diff
@@ -1519,28 +1519,25 @@ testDeleteUser =
...
-      alice ##> "/delete user alisa"
-      alice <## "cannot delete active user"
-
-      alice ##> "/users"
-      alice <## "alisa (active)"
-
-      alice <##> cath
-
-      -- delete user deleting SMP queues
+      -- cannot delete new active user wher
```
spaced4ndy, line 1522:
```suggestion
      -- cannot delete active user when there is another user
```

tests/ChatTests/Direct.hs
```diff
@@ -1553,6 +1550,16 @@ testDeleteUser =
 
       alice <##> dan
 
+      -- delete last active user deleting SMP queues
+
+      alice ##> "/delete user alisa2 del_smp=off"
```
spaced4ndy, line 1555: queues actually are not deleted here
```suggestion
      -- delete last active user

      alice ##> "/delete user alisa2 del_smp=off"
```

tests/ChatTests/Direct.hs
```diff
@@ -2064,17 +2082,14 @@ testUserPrivacy =
       userVisible alice ""
       alice ##> "/_hide user 2 \"another_password\""
       userHidden alice ""
-      alice ##> "/user alisa another_password"
-      showActiveUser alice "alisa"
-      alice ##> "/user alice"
-      showActiveUser alice "alice (Alice)"
```
spaced4ndy, line 2070: why is this deleted?

PR #3519, opened 2023-12-06
core: support batch sending in groups, batch introductions; send recent message history to new members

spaced4ndy:

PR #3568, opened 2023-12-18
core: add notes chat type

dpwiz:


src/Simplex/Chat/Messages.hs
```diff
@@ -69,6 +70,7 @@ data ChatRef = ChatRef ChatType Int64
 data ChatInfo (c :: ChatType) where
   DirectChat :: Contact -> ChatInfo 'CTDirect
   GroupChat :: GroupInfo -> ChatInfo 'CTGroup
+  NotesChat :: NotesFolder -> ChatInfo 'CTNotes
```
spaced4ndy, line 73: why not same order as in ChatType, or vice versa move it there

src/Simplex/Chat/Migrations/M20231219_notes_folders.hs
```diff
@@ -0,0 +1,41 @@
...
+m20231219_notes_folders :: Query
+m20231219_notes_folders =
+  [sql|
+    CREATE TABLE notes_folders (
+      notes_folder_id INTEGER PRIMARY KEY AUTOINCREMENT,
+      user_id INTEGER NOT NULL REFERENCES users ON DELETE CASCADE,
+      local_display_name TEXT NOT NULL,
+      created_at TEXT NOT NULL,
+      updated_at TEXT NOT NULL,
+      chat_ts TEXT,
```
spaced4ndy, line 17: since it's a new table it can be NOT NULL
```suggestion
      chat_ts TEXT NOT NULL,
```

src/Simplex/Chat/Migrations/M20231219_notes_folders.hs
```diff
@@ -0,0 +1,41 @@
...
+      favorite INTEGER NOT NULL DEFAULT 0,
+      unread_chat INTEGER DEFAULT 0 NOT NULL,
```
spaced4ndy, line 19: doesn't seem to make sense to mark your own notes as unread

spaced4ndy, line 19: doesn't seem to make sense to mark your own notes as unread
```suggestion
```

src/Simplex/Chat/Migrations/M20231219_notes_folders.hs
```diff
@@ -0,0 +1,41 @@
...
+      chat_item_id INTEGER DEFAULT NULL REFERENCES chat_items ON DELETE SET NULL,
```
spaced4ndy, line 20: this field is unnecessary - if I remember correctly it's only used in groups to refer to invitation item
```suggestion
```

src/Simplex/Chat/Migrations/M20231219_notes_folders.hs
```diff
@@ -0,0 +1,41 @@
...
+      FOREIGN KEY (user_id, local_display_name)
+        REFERENCES display_names (user_id, local_display_name)
+        ON DELETE CASCADE
+        ON UPDATE CASCADE,
+    );
+
+    CREATE UNIQUE INDEX idx_notes_user_local_display_name ON notes_fodler (
+      user_id,
+      local_display_name
+    );
```
spaced4ndy, line 30:
```suggestion
    CREATE UNIQUE INDEX idx_notes_folders_user_id_local_display_name ON notes_folders (
      user_id,
      local_display_name
    );
```

src/Simplex/Chat/Types.hs
```diff
@@ -327,6 +327,8 @@ type ContactName = Text
 
 type GroupName = Text
 
+type FolderName = Text
```
spaced4ndy, line 330: too similar to file system terminology..
```suggestion
type NotesFolderName = Text
```

src/Simplex/Chat/Types.hs
```diff
@@ -1520,6 +1522,21 @@ data XGrpMemIntroCont = XGrpMemIntroCont
...
   deriving (Show)
 
+data NotesFolder = NotesFolder
+  { notesFolderId :: NotesFolderId,
+    userId :: UserId,
+    localDisplayName :: FolderName,
+    chatItemId :: Maybe Int64,
+    createdAt :: UTCTime,
+    updatedAt :: UTCTime,
+    chatTs :: Maybe UTCTime,
```
spaced4ndy, line 1532:
```suggestion
    chatTs :: UTCTime,
```

src/Simplex/Chat/Types.hs
```diff
@@ -1520,6 +1522,21 @@ data XGrpMemIntroCont = XGrpMemIntroCont
...
+    favorite :: Bool,
+    unread :: Bool
```
spaced4ndy, line 1534:
```suggestion
```

src/Simplex/Chat/Migrations/M20231219_notes_folders.hs
```diff
@@ -0,0 +1,41 @@
...
+import Database.SQLite.Simple (Query)
+import Database.SQLite.Simple.QQ (sql)
+
+m20231219_notes_folders :: Query
+m20231219_notes_folders =
+  [sql|
+    CREATE TABLE notes_folders (
+      notes_folder_id INTEGER PRIMARY KEY AUTOINCREMENT,
+      user_id INTEGER NOT NULL REFERENCES users ON DELETE CASCADE,
+      local_display_name TEXT NOT NULL,
```
spaced4ndy, line 14: also needs at least a display_name (full profile doesn't make sense)

src/Simplex/Chat/Types.hs
```diff
@@ -1520,6 +1522,21 @@ data XGrpMemIntroCont = XGrpMemIntroCont
   }
   deriving (Show)
 
+data NotesFolder = NotesFolder
+  { notesFolderId :: NotesFolderId,
+    userId :: UserId,
+    localDisplayName :: FolderName,
```
spaced4ndy, line 1528: + displayName

src/Simplex/Chat/Migrations/M20231219_notes_folders.hs
dpwiz: "Dear future me, this folder contains some urgent stuff for you to resolve ASAP"

dpwiz: also, to have unified pagination/search/filter code with direct/group

src/Simplex/Chat.hs
```diff
@@ -759,6 +763,12 @@ processChatCommand = \case
             quoteData ChatItem {chatDir = CIGroupSnd, content = CISndMsgContent qmc} membership' = pure (qmc, CIQGroupSnd, True, membership')
             quoteData ChatItem {chatDir = CIGroupRcv m, content = CIRcvMsgContent qmc} _ = pure (qmc, CIQGroupRcv $ Just m, False, m)
             quoteData _ _ = throwChatError CEInvalidQuote
+    CTLocal -> do
+      nf@NoteFolder {noteFolderId} <- withStore $ \db -> getNoteFolder db user chatId
+      -- TODO: files, voice, etc.
+      msg <- createSndMessage (XMsgNew . MCSimple $ extMsgContent mc Nothing) (NoteFolderId noteFolderId)
+      ci <- saveSndChatItem user (CDLocalSnd nf) msg (CISndMsgContent mc)
```
spaced4ndy, line 770: It's not necessary to create records in `messages` table, only chat items. messages are only used to store "service messages", which are linked via msg_deliveries to messages as sent to / received from agent connections. Since we don't need any connections here, it's ok to simply put msg content (mc) into chat items, w/t even wrapping them into protocol events and encoding.

spaced4ndy: it may cause problems though if we ever allow to forward them to conversations with connections, because there they'd have to be wrapped. Perhaps in notes we should check for lower limit than protocol message limit (it wasn't present anywhere in chat code explicitly until #3519, in that branch it's maxChatMsgSize).

src/Simplex/Chat.hs
```diff
@@ -5659,11 +5689,11 @@ saveSndChatItem' user cd msg@SndMessage {sharedMsgId} content ciFile quotedItem
     pure ciId
   liftIO $ mkChatItem cd ciId content ciFile quotedItem (Just sharedMsgId) itemTimed live createdAt Nothing createdAt
 
-saveRcvChatItem :: ChatMonad m => User -> ChatDirection c 'MDRcv -> RcvMessage -> UTCTime -> CIContent 'MDRcv -> m (ChatItem c 'MDRcv)
+saveRcvChatItem :: (ChatMonad m, ChatTypeQuotable c) => User -> ChatDirection c 'MDRcv -> RcvMessage -> UTCTime -> CIContent 'MDRcv -> m (ChatItem c 'MDRcv)
```
spaced4ndy, line 5692: same here, not sure we need to even add constraints to these functions, because it might be simpler to create specific function for local chats, or use `createInternalChatItem`. Local chat items are "snd"/"rcv" only in appearance, they won't really be sent or received, and so they don't require `msg`, `sharedMsgId_` used in saveRcv/SndChatItem functions, meaning these functions won't fit the use case.

src/Simplex/Chat/Messages.hs
```diff
@@ -790,7 +831,7 @@ data PendingGroupMessage = PendingGroupMessage
 
 type MessageId = Int64
 
-data ConnOrGroupId = ConnectionId Int64 | GroupId Int64
+data ConnOrGroupId = ConnectionId Int64 | GroupId Int64 | NoteFolderId Int64
```
spaced4ndy, line 834: this is only used in "deliver to connection" functions and maybe some other "service messages" related functions, so it doesn't need a new constructor

src/Simplex/Chat/Migrations/M20231219_note_folders.hs
```diff
@@ -0,0 +1,45 @@
...
+import Database.SQLite.Simple (Query)
+import Database.SQLite.Simple.QQ (sql)
+
+m20231219_note_folders :: Query
+m20231219_note_folders =
+  [sql|
+    CREATE TABLE note_folders (
+      note_folder_id INTEGER PRIMARY KEY AUTOINCREMENT,
+      user_id INTEGER NOT NULL REFERENCES users ON DELETE CASCADE,
+      display_name TEXT NOT NULL,
```
spaced4ndy, line 14: + image
@epoberezkin do we need images for these?

src/Simplex/Chat/Migrations/M20231219_note_folders.hs
```diff
@@ -0,0 +1,45 @@
...
+        ON UPDATE CASCADE
+    );
+
+    CREATE UNIQUE INDEX idx_note_folders_user_id_local_display_name ON note_folders (
+      user_id,
+      local_display_name
+    );
+
+    ALTER TABLE chat_items ADD COLUMN note_folder_id INTEGER DEFAULT NULL REFERENCES note_folders ON DELETE CASCADE;
+    ALTER TABLE messages ADD COLUMN note_folder_id INTEGER DEFAULT NULL REFERENCES note_folders ON DELETE CASCADE;
```
spaced4ndy, line 33:
```suggestion
```

src/Simplex/Chat/Migrations/M20231219_note_folders.hs
```diff
@@ -0,0 +1,45 @@
...
+    ALTER TABLE chat_item_reactions ADD COLUMN note_folder_id INTEGER DEFAULT NULL REFERENCES note_folders ON DELETE CASCADE;
+|]
+
+down_m20231219_note_folders :: Query
+down_m20231219_note_folders =
+  [sql|
+DROP INDEX idx_note_folders_user_id_local_display_name;
+DROP TABLE note_folders;
+ALTER TABLE chat_items DROP COLUMN note_folder_id;
+ALTER TABLE messages DROP COLUMN note_folder_id;
```
spaced4ndy, line 43:
```suggestion
```

src/Simplex/Chat/Migrations/M20231219_note_folders.hs
spaced4ndy: for same reasons as above

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -724,6 +734,50 @@ getGroupChatPreview_ db user (GroupChatPD _ groupId lastItemId_ stats) = do
...
   pure $ AChat SCTGroup (Chat (GroupChat groupInfo) lastItem stats)
 
+findLocalChatPreview_ :: DB.Connection -> User -> PaginationByTime -> ChatListQuery -> IO [AChatPreviewData]
+findLocalChatPreview_ db User {userId} pagination clq = pure []
+
+getLocalChatPreview_ :: DB.Connection -> User -> ChatPreviewData 'CTLocal -> ExceptT StoreError IO AChat
+getLocalChatPreview_ = error "TODO: getLocalChatPreview_"
+
+-- this function can be changed so it never fails, not only avoid failure on invalid json
+toLocalChatItem :: UTCTime -> ChatItemRow -> Either StoreError (CChatItem 'CTLocal)
```
spaced4ndy, line 744: what are the differences between this and toDirectChatItem, except this not having QuoteRow? Could this reuse most of the functionality of the latter? Or are "direct" and "note" chat items also somehow conceptually different?

spaced4ndy: for example, direct chat items already include internal chat items, which can't be quoted. So possibly there could be a toChatItemWtQuote (needs a better name), and toDirectChatItem could reuse it

src/Simplex/Chat.hs
dpwiz: `getChatItemQuote_` will have to "throw" some error from store operation in IO that isn't gonna even happen

src/Simplex/Chat/Store/Messages.hs
dpwiz: `maybeCIFile`, `ciMeta` and `ciTimed` can be reused even for groups. Not sure if that would be worth collecting all the closure variables to arguments.

src/Simplex/Chat/Migrations/M20231219_note_folders.hs
epoberezkin: yes

src/Simplex/Chat/Migrations/M20231219_notes_folders.hs
epoberezkin: unread won't be available in the UI until we support multiple Note folders (we don't need API to create/delete now, it'll be just created for each user profile when profile is created).

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -1824,6 +2166,19 @@ getGroupCIReactions db GroupInfo {groupId} itemMemberId itemSharedMsgId =
       |]
       (groupId, itemMemberId, itemSharedMsgId)
 
+getLocalCIReactions :: DB.Connection -> NoteFolder -> SharedMsgId -> IO [CIReactionCount]
```
epoberezkin, line 2169: remove, can't have reactions

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -1918,6 +2281,38 @@ setGroupReaction db GroupInfo {groupId} m itemMemberId itemSharedMId sent reacti
         |]
         (groupId, groupMemberId' m, itemSharedMId, itemMemberId, sent, reaction)
 
+getLocalReactions :: DB.Connection -> NoteFolder -> SharedMsgId -> Bool -> IO [MsgReaction]
```
epoberezkin, line 2284: remove

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -1918,6 +2281,38 @@ setGroupReaction db GroupInfo {groupId} m itemMemberId itemSharedMId sent reacti
...
+    <$> DB.query
+      db
+      [sql|
+        SELECT reaction
+        FROM chat_item_reactions
+        WHERE note_folder_id = ? AND shared_msg_id = ? AND reaction_sent = ?
+      |]
+      (noteFolderId, itemSharedMId, sent)
+
+setLocalReaction :: DB.Connection -> NoteFolder -> SharedMsgId -> MsgReaction -> Bool -> UTCTime -> IO ()
```
epoberezkin, line 2296: remove

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -1714,14 +1957,109 @@ getGroupChatItemIdByText' db User {userId} groupId msg =
       |]
       (userId, groupId, msg <> "%")
 
+getLocalCIWithReactions :: DB.Connection -> User -> NoteFolder -> ChatItemId -> ExceptT StoreError IO (CChatItem 'CTLocal)
+getLocalCIWithReactions db user nf@NoteFolder {noteFolderId} itemId = do
+  liftIO . localCIWithReactions db nf =<< getLocalChatItem db user noteFolderId itemId
+
+localCIWithReactions :: DB.Connection -> NoteFolder -> CChatItem 'CTLocal -> IO (CChatItem 'CTLocal)
```
epoberezkin, line 1964: remove

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -1714,14 +1957,109 @@ getGroupChatItemIdByText' db User {userId} groupId msg =
       |]
       (userId, groupId, msg <> "%")
 
+getLocalCIWithReactions :: DB.Connection -> User -> NoteFolder -> ChatItemId -> ExceptT StoreError IO (CChatItem 'CTLocal)
```
epoberezkin, line 1960: remove

src/Simplex/Chat.hs
```diff
@@ -1901,9 +1992,13 @@ processChatCommand' vr = \case
                     updateRcvFileAgentId db fileId Nothing
                   getChatItemByFileId db vr user fileId
                 pure $ CRRcvFileCancelled user ci ftr
-  FileStatus fileId -> withUser $ \user -> do
+  FileStatus fileId -> withUser $ \user@User {userId} -> do
     ci@(AChatItem _ _ _ ChatItem {file}) <- withStore $ \db -> getChatItemByFileId db vr user fileId
     case file of
+      Just CIFile {fileProtocol = FPLocal} -> do
+        -- XXX: not stricly a file transfer but the file *has* some status
```
epoberezkin, line 1999: probably should just throw error

src/Simplex/Chat.hs
```diff
@@ -904,6 +940,20 @@ processChatCommand' vr = \case
               r = ACIReaction SCTGroup SMDSnd (GroupChat g) $ CIReaction CIGroupSnd ci' createdAt reaction
           pure $ CRChatItemReaction user add r
         _ -> throwChatError $ CECommandError "reaction not possible - no shared item ID"
+    CTLocal -> do
```
epoberezkin, line 943: not needed, remove

src/Simplex/Chat.hs
```diff
@@ -6321,6 +6444,9 @@ chatCommandP =
       "/_invite member contact @" *> (APISendMemberContactInvitation <$> A.decimal <*> optional (A.space *> msgContentP)),
       (">#" <|> "> #") *> (SendGroupMessageQuote <$> displayName <* A.space <*> pure Nothing <*> quotedMsg <*> msgTextP),
       (">#" <|> "> #") *> (SendGroupMessageQuote <$> displayName <* A.space <* char_ '@' <*> (Just <$> displayName) <* A.space <*> quotedMsg <*> msgTextP),
+      -- "/notes" $> ListNoteFolders, -- TODO
+      -- "/_new local chat " *> (APINewLocalChat <$> A.decimal <*> jsonP),
```
spaced4ndy, line 6448: remove?

src/Simplex/Chat.hs
```diff
@@ -785,6 +790,19 @@ processChatCommand' vr = \case
       unzipMaybe3 :: Maybe (a, b, c) -> (Maybe a, Maybe b, Maybe c)
       unzipMaybe3 (Just (a, b, c)) = (Just a, Just b, Just c)
       unzipMaybe3 _ = (Nothing, Nothing, Nothing)
+  APICreateChatItem folderId (ComposedMessage file_ quotedItemId_ mc) -> withUser $ \user -> do
+    forM_ quotedItemId_ $ \_ -> throwError $ ChatError $ CECommandError "not supported"
+    nf <- withStore $ \db -> getNoteFolder db user folderId
+    -- TODO: assertLocalAllowed user MDSnd nf XMsgNew_
```
spaced4ndy, line 796: XMsgNew doesn't make sense in this context, it's a protocol message. Api already requires msg content so only content items can be created
```suggestion
```

src/Simplex/Chat.hs
```diff
@@ -785,6 +790,19 @@ processChatCommand' vr = \case
...
+    ci'@ChatItem {meta = CIMeta {itemId, itemTs}} <- createLocalChatItem user (CDLocalSnd nf) (CISndMsgContent mc) Nothing
+    ciFile_ <- forM file_ $ \cf@CryptoFile {filePath, cryptoArgs} -> do
+      fsFilePath <- toFSFilePath filePath -- XXX: only used for size?..
```
spaced4ndy, line 799: it also appends files folder (it changes in mobile apps between updates/installations so we have to work with relative paths)
```suggestion
      fsFilePath <- toFSFilePath filePath
```

src/Simplex/Chat.hs
```diff
@@ -830,6 +848,21 @@ processChatCommand' vr = \case
                 else pure $ CRChatItemNotChanged user (AChatItem SCTGroup SMDSnd (GroupChat gInfo) ci)
             _ -> throwChatError CEInvalidChatItemUpdate
         CChatItem SMDRcv _ -> throwChatError CEInvalidChatItemUpdate
+    CTLocal -> do
+      (nf@NoteFolder {noteFolderId}, cci) <- withStore $ \db -> (,) <$> getNoteFolder db user chatId <*> getLocalChatItem db user chatId itemId
+      case cci of
+        CChatItem SMDRcv _ -> throwChatError CEInvalidChatItemUpdate
+        CChatItem SMDSnd ci@ChatItem {meta = CIMeta {editable}, content = ciContent} -> do
```
spaced4ndy, line 855: doesn't really make sense to limit edit window in case of local items (see mkCIMeta)

spaced4ndy: basically can just throw away check of `editable` from case

src/Simplex/Chat.hs
```diff
@@ -1004,8 +1063,18 @@ processChatCommand' vr = \case
                       withStore' (\db -> setContactDeleted db user ct)
                         `catchChatError` (toView . CRChatError (Just user))
                       pure $ map aConnId conns
+    CTLocal -> do
+      nf <- withStore $ \db -> getNoteFolder db user chatId
+      filesInfo <- withStore' $ \db -> getNoteFolderFileInfo db user nf
+      withChatLock "deleteChat local" . procCmd $ do
+        mapM_ (deleteFile user) filesInfo
+        -- functions below are called in separate transactions to prevent crashes on android
+        -- (possibly, race condition on integrity check?)
```
spaced4ndy, line 1072: we don't know anymore even for groups at the moment, maybe the issue is gone. for notes we haven't even UI yet :)
```suggestion
```

spaced4ndy: not saying it's necessary to unite transactions, just remove comment

src/Simplex/Chat.hs
```diff
@@ -5876,6 +5974,12 @@ deleteGroupCI user gInfo ci@ChatItem {file} byUser timed byGroupMember_ deletedT
   where
     gItem = AChatItem SCTGroup msgDirection (GroupChat gInfo)
 
+deleteLocalCI :: (ChatMonad m, MsgDirectionI d) => User -> NoteFolder -> ChatItem 'CTLocal d -> Bool -> Bool -> m ChatResponse
+deleteLocalCI user nf ci@ChatItem {file} byUser timed = do
+  deleteCIFile user file
+  withStoreCtx' (Just "deleteLocalCI, deleteLocalChatItem") $ \db -> deleteLocalChatItem db user nf ci
```
spaced4ndy, line 5980:
```suggestion
  withStore' $ \db -> deleteLocalChatItem db user nf ci
```

src/Simplex/Chat.hs
```diff
@@ -6014,13 +6118,29 @@ sameGroupProfileInfo p p' = p {groupPreferences = Nothing} == p' {groupPreferenc
 
 createInternalChatItem :: forall c d m. (ChatTypeI c, MsgDirectionI d, ChatMonad m) => User -> ChatDirection c d -> CIContent d -> Maybe UTCTime -> m ()
 createInternalChatItem user cd content itemTs_ = do
+  ci <- createInternalChatItem_ user cd content itemTs_
+  toView $ CRNewChatItem user (AChatItem (chatTypeI @c) (msgDirection @d) (toChatInfo cd) ci)
+
+createInternalChatItem_ :: (MsgDirectionI d, ChatMonad m) => User -> ChatDirection c d -> CIContent d -> Maybe UTCTime -> m (ChatItem c d)
```
spaced4ndy, line 6124: why both createInternalChatItem_ and createLocalChatItem are required? probably refactoring of the former can be reverted. so there's only createInternalChatItem without underscore and createLocalChatItem

src/Simplex/Chat.hs
```diff
@@ -6014,13 +6118,29 @@ sameGroupProfileInfo p p' = p {groupPreferences = Nothing} == p' {groupPreferenc
...
+createLocalChatItem user cd content itemTs_ = do
+  createdAt <- liftIO getCurrentTime
+  let itemTs = fromMaybe createdAt itemTs_
+  gVar <- asks random
+  ciId <- withStore $ \db -> do
+    when (ciRequiresAttention content) . liftIO $ updateChatTs db user cd createdAt
+    createWithRandomId gVar $ \sharedMsgId ->
+      let smi_ = Just (SharedMsgId sharedMsgId)
+       in createNewChatItem_ db user cd Nothing smi_ content (Nothing, Nothing, Nothing, Nothing, Nothing) Nothing False itemTs Nothing createdAt
+  liftIO $ mkChatItem cd ciId content Nothing Nothing Nothing Nothing False itemTs Nothing createdAt
```
spaced4ndy, line 6143: how does it work when chat item has file if it always passes Nothing to Maybe CIFile? it seems files will not be displayed in UI, probably need to add tests where items with files are created and chat is read. see tests by search `"messages with files"`

src/Simplex/Chat/Controller.hs
```diff
@@ -555,8 +559,11 @@ data ChatResponse
   | CRUserDeletedMember {user :: User, groupInfo :: GroupInfo, member :: GroupMember}
   | CRGroupsList {user :: User, groups :: [(GroupInfo, GroupSummary)]}
   | CRSentGroupInvitation {user :: User, groupInfo :: GroupInfo, contact :: Contact, member :: GroupMember}
+  | CRNoteFolderCreated {user :: User, noteFolder :: NoteFolder}
+  | CRNoteFolderDeleted {user :: User, noteFolder :: NoteFolder}
   | CRFileTransferStatus User (FileTransfer, [Integer]) -- TODO refactor this type to FileTransferStatus
   | CRFileTransferStatusXFTP User AChatItem
+  | CRLocalFileStatus User LocalFileMeta
```
spaced4ndy, line 566:
```suggestion
```

src/Simplex/Chat/Controller.hs
```diff
@@ -257,6 +257,7 @@ data ChatCommand
   | APIGetChatItems ChatPagination (Maybe String)
```
spaced4ndy, line 257: need api to change notes "profile" (name and image)

src/Simplex/Chat/Store/Files.hs
```diff
@@ -912,24 +919,70 @@ getFileTransferMeta_ db userId fileId =
...
+          file_name, file_path, file_size,
+          file_crypto_key, file_crypto_nonce,
+          chunk_size, file_inline, ci_file_status, protocol, created_at, updated_at
+        )
+      VALUES (?,?,?,?,?,?,?,?,?,?,?,?,?,?)
+    |]
+    ( (userId, noteFolderId, chatItemId)
+        :. (takeFileName filePath, filePath, fileSize)
+        :. maybe (Nothing, Nothing) (\(CFArgs key nonce) -> (Just key, Just nonce)) cryptoArgs
+        :. (65536 :: Int, Nothing :: Maybe InlineFileMode, fileStatus, FPLocal, itemTs, itemTs)
```
spaced4ndy, line 938: is 65536 chunkSize? better use fileChunkSize

spaced4ndy: would be best to use null but not possible w/t schema change.. at least it'd be the same as everywhere. It's can't be changed anyway

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -697,6 +715,133 @@ getGroupChatPreview_ db vr user (GroupChatPD _ groupId lastItemId_ stats) = do
...
+
+getLocalChatPreview_ :: DB.Connection -> User -> ChatPreviewData 'CTLocal -> ExceptT StoreError IO AChat
+getLocalChatPreview_ db user (LocalChatPD _ noteFolderId lastItemId_ stats) = do
+  nf <- getNoteFolder db user noteFolderId
+  lastItem <- case lastItemId_ of
+    Just lastItemId -> (: []) <$> getLocalChatItem db user noteFolderId lastItemId
+    Nothing -> pure []
+  pure $ AChat SCTLocal (Chat (LocalChat nf) lastItem stats)
+
+-- this function can be changed so it never fails, not only avoid failure on invalid json
```
spaced4ndy, line 807: clarify or remove. is it comment for pr and do you propose to remove `badItem`?

src/Simplex/Chat/Store/NoteFolders.hs
```diff
@@ -0,0 +1,96 @@
+{-# LANGUAGE DuplicateRecordFields #-}
```
spaced4ndy, line 1: + method to update profile (name, image)

src/Simplex/Chat/Types.hs
```diff
@@ -1523,6 +1535,22 @@ data XGrpMemIntroCont = XGrpMemIntroCont
   }
   deriving (Show)
 
+-- | An entity for local chats
+data NoteFolder = NoteFolder
+  { noteFolderId :: NoteFolderId,
+    userId :: UserId,
+    displayName :: NoteFolderName,
```
spaced4ndy, line 1542: \+ image

src/Simplex/Chat/View.hs
```diff
@@ -262,6 +263,8 @@ responseToView hu@(currentRH, user_) ChatConfig {logLevel, showReactions, showRe
   CRMemberSubError u g m e -> ttyUser u [ttyGroup' g <> " member " <> ttyMember m <> " error: " <> sShow e]
   CRMemberSubSummary u summary -> ttyUser u $ viewErrorsSummary (filter (isJust . memberError) summary) " group member errors"
   CRGroupSubscribed u g -> ttyUser u $ viewGroupSubscribed g
+  CRNoteFolderCreated u NoteFolder {displayName} -> ttyUser u ["new note folder created, write to $" <> plain displayName <> " to add notes"]
```
spaced4ndy, line 266: I think more common language in view is "use", also use `highlight` for command
```suggestion
  CRNoteFolderCreated u NoteFolder {displayName} -> ttyUser u ["new note folder created, use $" <> plain displayName <> " to add notes"]
```

src/Simplex/Chat/View.hs
```diff
@@ -1871,6 +1908,7 @@ viewChatError logLevel testView = \case
     SEDuplicateGroupMessage {groupId, sharedMsgId}
       | testView -> ["duplicate group message, group id: " <> sShow groupId <> ", message id: " <> sShow sharedMsgId]
       | otherwise -> []
+    SENoteFolderNotFoundByName f -> ["no folder " <> ttyLocal f]
```
spaced4ndy, line 1911: + check language in other places in view - to not confuse with file system, common to call directories as folders
```suggestion
    SENoteFolderNotFoundByName f -> ["no notes folder " <> ttyLocal f]
```

tests/ChatTests/Local.hs
```diff
@@ -0,0 +1,92 @@
...
+
+testFiles :: FilePath -> IO ()
+testFiles tmp = withNewTestChat tmp "alice" aliceProfile $ \alice -> do
+  createFolder alice "self"
+
+  alice #$> ("/_files_folder ./tests/tmp/app_files", id, "ok")
+  copyFile "./tests/fixtures/test.jpg" "./tests/tmp/app_files/test.jpg"
+  alice ##> "/_create $1 json {\"filePath\": \"test.jpg\", \"msgContent\": {\"text\":\"\",\"type\":\"image\",\"image\":\"data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAgAAAAIAQMAAAD+wSzIAAAABlBMVEX///+/v7+jQ3Y5AAAADklEQVQI12P4AIX8EAgALgAD/aNpbtEAAAAASUVORK5CYII=\"}}"
+  alice <# "$self file 1 (test.jpg)"
+  alice ##> "/tail"
```
spaced4ndy, line 80: why does it return zero chat items

tests/ChatTests/Local.hs
```diff
@@ -0,0 +1,92 @@
...
+
+import ChatClient
+import ChatTests.Utils
+import Test.Hspec
+import System.Directory (copyFile)
+
+chatLocalTests :: SpecWith FilePath
+chatLocalTests = do
+  fdescribe "note folders" $ do
+    it "create folders, add notes, read, search" testNotes
```
spaced4ndy, line 14: tests are too sparse, add tests for following functionality:
- messages with files (both text and file), should get chat and check list of messages via chatF
- create several (at least 2) folders, create messages in both, get chat and check list of messages (currently a single test creates two folders and doesn't create messages in 2nd)
- pagination inside notes chat - not only last, before and after should test for different scenarios (e.g. after but not till end, before and not till beginning)
- pagination of chat list with notes included (e.g. in testPaginationAllChatTypes)

possibly other scenarios

tests/ChatTests/Local.hs
```diff
@@ -0,0 +1,92 @@
...
+  alice <# "$self Greetings."
+
+  alice ##> "/delete $self"
+  alice <## "note folder self deleted"
+
+testUserNotes :: FilePath -> IO ()
+testUserNotes tmp = withNewTestChat tmp "alice" aliceProfile $ \alice -> do
+  createFolder alice "self"
+
+  alice #> "$self keep in mind"
```
epoberezkin, line 57: re api - we won't allow adding folders, there will only be one folder for notes without names, so it'll be just $

src/Simplex/Chat/Store/Messages.hs
dpwiz: IDK, it's a boilerplate comment for the 3 toXxxChat functions.

tests/ChatTests/Local.hs
spaced4ndy: made edits according to recent discussion to get rid of names and api to create

src/Simplex/Chat.hs
dpwiz: ciFile_ gets attached later with a record update

src/Simplex/Chat.hs
spaced4ndy: I think it's because chat has no reason to read its contents, but size is used for metadata (e.g. sent in invitation, or in this case only displayed on chat item vi CIFile)

src/Simplex/Chat/Store/Messages.hs
spaced4ndy: ok then :)

src/Simplex/Chat.hs
spaced4ndy: ok, I see now. In case of local file it may be better to save both chat item and file in a single transaction, it seems there's no work happening in between (like preparation of file invitation / requesting agent to send xftp file / preparation of sharedMsgId).

spaced4ndy: You wouldn't have to go around this function then and make a record update. This may become a trap if mkChatItem gets some additional logic later. These record updates (or lack of them) after creating a record in database were an often source of bugs at some point.. we try to avoid it now

tests/ChatTests/Local.hs
```diff
@@ -0,0 +1,128 @@
...
+  alice ##> "/_delete item $1 2 internal"
+  alice <## "message deleted"
+  doesFileExist stored2 `shouldReturn` False
+  doesFileExist stored `shouldReturn` True
+
+  alice ##> "/clear -"
+  alice ##> "/fs 1"
+  alice <## "chat db error: SEChatItemNotFoundByFileId {fileId = 1}"
+  alice ##> "/tail"
+  doesFileExist stored `shouldReturn` False
```
spaced4ndy, line 123: also test that that clearing notes chat doesn't remove files from other conversations (it does use some new function to retrieve list of files I think). maybe it's better to split this second part into one or two separate tests to easier identify what fails when something breaks tests.

tests/ChatTests/Local.hs
```diff
@@ -0,0 +1,128 @@
...
+  alice /- "last"
+  tsE <- iso8601Show <$> getCurrentTime
+
+  -- there's only one folder that got updated after tsM and before tsE
+  getChats_ alice "count=3" [("-", "last")]
+  getChats_ alice ("after=" <> tsE <> " count=10") []
+  getChats_ alice ("after=" <> tsS <> " count=10") [("-", "last")]
+  getChats_ alice ("before=" <> tsM <> " count=10") []
+  getChats_ alice ("before=" <> tsE <> " count=10") [("-", "last")]
+  getChats_ alice ("before=" <> tsS <> " count=10") []
```
spaced4ndy, line 80: this only tests chat list pagination, should also test chat pagination (pagination of chat items inside chat). also when you always take count=10 it doesn't show that "after" behaves correctly and not the same as "last" - chat should have more items and test should check after some early item but not until the end

src/Simplex/Chat.hs
```diff
@@ -785,6 +790,22 @@ processChatCommand' vr = \case
...
+    let content = CISndMsgContent mc
+    let cd = CDLocalSnd nf
+    ciId <- createLocalChatItem user cd content createdAt
+    ciFile_ <- forM file_ $ \cf@CryptoFile {filePath, cryptoArgs} -> do
+      fsFilePath <- toFSFilePath filePath
+      fileSize <- liftIO $ CF.getFileContentsSize $ CryptoFile fsFilePath cryptoArgs
+      chunkSize <- asks $ fileChunkSize . config
+      withStore' $ \db -> do
+        fileId <- createLocalFile CIFSSndComplete db user nf ciId createdAt cf fileSize chunkSize
+        pure CIFile {fileId, fileName = takeFileName filePath, fileSize, fileSource = Just cf, fileStatus = CIFSSndComplete, fileProtocol = FPLocal}
```
spaced4ndy, line 806: maybe it's better to use CIFSSndStored, it should avoid some unnecessary effects in UI (e.g. possibly checkmark). Also logically they never move from this state comparing to regular files, not that they complete some transfer
```suggestion
        pure CIFile {fileId, fileName = takeFileName filePath, fileSize, fileSource = Just cf, fileStatus = CIFSSndStored, fileProtocol = FPLocal}
```

src/Simplex/Chat/Help.hs
```diff
@@ -106,6 +106,7 @@ filesHelpInfo =
     [ green "File transfer commands:",
       indent <> highlight "/file @<contact> <file_path>       " <> " - send file to contact",
       indent <> highlight "/file #<group> <file_path>         " <> " - send file to group",
+      indent <> highlight "/file $<folder> <file_path>        " <> " - add file to folder",
```
spaced4ndy, line 109: this probably doesn't work now since APISendMessage prohibits notes

src/Simplex/Chat/Migrations/M20240102_note_folders.hs
```diff
@@ -0,0 +1,42 @@
...
+
+m20240102_note_folders :: Query
+m20240102_note_folders =
+  [sql|
+    CREATE TABLE note_folders (
+      note_folder_id INTEGER PRIMARY KEY AUTOINCREMENT,
+      user_id INTEGER NOT NULL REFERENCES users ON DELETE CASCADE,
+      created_at TEXT NOT NULL,
+      updated_at TEXT NOT NULL,
+      chat_ts TEXT NOT NULL,
```
spaced4ndy, line 16:
```suggestion
      created_at TEXT NOT NULL DEFAULT(datetime('now')),
      updated_at TEXT NOT NULL DEFAULT(datetime('now')),
      chat_ts TEXT NOT NULL DEFAULT(datetime('now')),
```

src/Simplex/Chat/Migrations/M20240102_note_folders.hs
```diff
@@ -0,0 +1,42 @@
...
+      favorite INTEGER NOT NULL DEFAULT 0,
+      unread_chat INTEGER DEFAULT 0 NOT NULL
```
spaced4ndy, line 18:
```suggestion
      favorite INTEGER NOT NULL DEFAULT 0,
      unread_chat INTEGER NOT NULL DEFAULT 0
```

src/Simplex/Chat/Migrations/M20240102_note_folders.hs
```diff
@@ -0,0 +1,42 @@
...
+    INSERT INTO note_folders
+      SELECT
+        NULL as note_folder_id,
+        u.user_id as user_id,
+        datetime('now') as created_at,
+        datetime('now') as updated_at,
+        datetime('now') as chat_ts,
+        0 as favorite,
+        0 as unread_chat
+      FROM users u;
```
spaced4ndy, line 33: explicitly specify list of columns names in INSERT, wouldn't require aliases then. It's a common syntax in other queries in code and this case doesn't seem to warrant new syntax. Also inserting NULL into primary key isn't necessary, it's autofilled.
```suggestion
    INSERT INTO note_folders
      (user_id, .....)
      SELECT
        u.user_id as user_id,
        datetime('now') as created_at,
        datetime('now') as updated_at,
        datetime('now') as chat_ts,
        0 as favorite,
        0 as unread_chat
      FROM users u;
```

src/Simplex/Chat/Store/Profiles.hs
```diff
@@ -120,6 +121,11 @@ createUserRecordAt db (AgentUserId auId) Profile {displayName, fullName, image,
       (profileId, displayName, userId, True, currentTs, currentTs, currentTs)
     contactId <- insertedRowId db
     DB.execute db "UPDATE users SET contact_id = ? WHERE user_id = ?" (contactId, userId)
+    DB.execute
```
spaced4ndy, line 124: as this isn't a direct part of user record, it's better to create separately in CreateActiveUser and getCreateActiveUser, same as `createContact db user simplexContactProfile` in the former.

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -697,6 +717,123 @@ getGroupChatPreview_ db vr user (GroupChatPD _ groupId lastItemId_ stats) = do
...
+          ( baseQuery
+              <> [sql|
+                    WHERE nf.user_id = :user_id
+                      AND (nf.favorite = 1
+                        OR nf.unread_chat = 1 OR ChatStats.UnreadCount > 0)
+                  |]
+              <> pagQuery
+          )
+          ([":user_id" := userId, ":rcv_new" := CISRcvNew] <> pagParams)
+      CLQSearch {} -> pure []
```
spaced4ndy, line 789: probably not implemented, api seems to support search

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -967,11 +1104,88 @@ getGroupChatBefore_ db user@User {userId} g@GroupInfo {groupId} beforeChatItemId
...
+    getLocalChatItemIdsLast_ :: IO [ChatItemId]
+    getLocalChatItemIdsLast_ =
+      map fromOnly
+        <$> DB.query
+          db
+          [sql|
+            SELECT chat_item_id
+            FROM chat_items
+            WHERE user_id = ? AND note_folder_id = ? AND item_text LIKE '%' || ? || '%'
+            ORDER BY item_ts DESC, chat_item_id DESC
```
spaced4ndy, line 1132: better to repeat logic from direct chats here and in other chat pagination functions (especially it affects before and after queries by removing some extra queries and filters)
```suggestion
            ORDER BY created_at DESC, chat_item_id DESC
```

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -967,11 +1104,88 @@ getGroupChatBefore_ db user@User {userId} g@GroupInfo {groupId} beforeChatItemId
...
+            LIMIT ?
+          |]
+          (userId, noteFolderId, search, count)
+
+getLocalChatAfter_ :: DB.Connection -> User -> NoteFolder -> ChatItemId -> Int -> String -> ExceptT StoreError IO (Chat 'CTLocal)
+getLocalChatAfter_ db user@User {userId} nf@NoteFolder {noteFolderId} afterChatItemId count search = do
+  let stats = ChatStats {unreadCount = 0, minUnreadItemId = 0, unreadChat = False}
+  afterChatItem <- getLocalChatItem db user noteFolderId afterChatItemId
```
spaced4ndy, line 1140: this query can be avoided

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -967,11 +1104,88 @@ getGroupChatBefore_ db user@User {userId} g@GroupInfo {groupId} beforeChatItemId
...
+    getLocalChatItemIdsAfter_ afterChatItemTs =
+      map fromOnly
+        <$> DB.query
+          db
+          [sql|
+            SELECT chat_item_id
+            FROM chat_items
+            WHERE user_id = ? AND note_folder_id = ? AND item_text LIKE '%' || ? || '%'
+              AND (item_ts > ? OR (item_ts = ? AND chat_item_id > ?))
+            ORDER BY item_ts ASC, chat_item_id ASC
```
spaced4ndy, line 1155:
```suggestion
              AND chat_item_id > ?
            ORDER BY created_at ASC, chat_item_id ASC
```

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -967,11 +1104,88 @@ getGroupChatBefore_ db user@User {userId} g@GroupInfo {groupId} beforeChatItemId
...
+            LIMIT ?
+          |]
+          (userId, noteFolderId, search, afterChatItemTs, afterChatItemTs, afterChatItemId, count)
+
+getLocalChatBefore_ :: DB.Connection -> User -> NoteFolder -> ChatItemId -> Int -> String -> ExceptT StoreError IO (Chat 'CTLocal)
+getLocalChatBefore_ db user@User {userId} nf@NoteFolder {noteFolderId} beforeChatItemId count search = do
+  let stats = ChatStats {unreadCount = 0, minUnreadItemId = 0, unreadChat = False}
+  beforeChatItem <- getLocalChatItem db user noteFolderId beforeChatItemId
```
spaced4ndy, line 1163: same, it seems this query can be avoided

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -967,11 +1104,88 @@ getGroupChatBefore_ db user@User {userId} g@GroupInfo {groupId} beforeChatItemId
...
+    getLocalChatItemIdsBefore_ beforeChatItemTs =
+      map fromOnly
+        <$> DB.query
+          db
+          [sql|
+            SELECT chat_item_id
+            FROM chat_items
+            WHERE user_id = ? AND note_folder_id = ? AND item_text LIKE '%' || ? || '%'
+              AND (item_ts < ? OR (item_ts = ? AND chat_item_id < ?))
+            ORDER BY item_ts DESC, chat_item_id DESC
```
spaced4ndy, line 1178:
```suggestion
              AND chat_item_id < ?
            ORDER BY created_at DESC, chat_item_id DESC
```

src/Simplex/Chat/Store/Messages.hs
spaced4ndy: groups have more difficult pagination logic because messages arrive from different connections in arbitrary order, and since we don't maintain causality graph there yet, there's currently no better ordering criteria than item_ts (which is also not perfect but better than created_at)

src/Simplex/Chat/Types.hs
```diff
@@ -1171,6 +1182,7 @@ fileTransferCancelled (FTSnd FileTransferMeta {cancelled} _) = cancelled
 fileTransferCancelled (FTRcv RcvFileTransfer {cancelled}) = cancelled
 
 -- For XFTP file transfers FSConnected means "uploaded to XFTP relays"
+-- Local files are always FSComplete
```
spaced4ndy, line 1185: FileStatus only relates to file transfers I think (maybe even only SMP file transfers? not sure)
```suggestion
```

src/Simplex/Chat/Migrations/M20240102_note_folders.hs
```diff
@@ -0,0 +1,42 @@
...
+      user_id INTEGER NOT NULL REFERENCES users ON DELETE CASCADE,
+      created_at TEXT NOT NULL,
+      updated_at TEXT NOT NULL,
+      chat_ts TEXT NOT NULL,
+      favorite INTEGER NOT NULL DEFAULT 0,
+      unread_chat INTEGER DEFAULT 0 NOT NULL
+    );
+
+    ALTER TABLE chat_items ADD COLUMN note_folder_id INTEGER DEFAULT NULL REFERENCES note_folders ON DELETE CASCADE;
+    ALTER TABLE files ADD COLUMN note_folder_id INTEGER DEFAULT NULL REFERENCES note_folders ON DELETE CASCADE;
```
spaced4ndy, line 22: add indexes:
- chat_items(note_folder_id)
- files(note_folder_id)
- note_folders(user_id)? probably excessive but we don't limit number of users and it won't hurt
- index for notes chat pagination - probably it would be smth like chat_items(user_id, note_folder_id, chat_item_id) once queries are changed there, but check last/before/after queries plans
- queries for chat list pagination would probably reuse existing indexes but also better to check plans for a few example queries

src/Simplex/Chat/Store/Messages.hs
spaced4ndy: nvm, I guess it's due to removal of names

tests/ChatTests/Local.hs
dpwiz: There's only one "chat" here, so any `count` above `1` is the same.

src/Simplex/Chat.hs
```diff
@@ -6513,6 +6598,7 @@ chatCommandP =
           " member" $> GRMember,
           " observer" $> GRObserver
         ]
+    notesNameP = A.char '-' $> ChatName CTLocal "-"
```
epoberezkin, line 6601:
```suggestion
    notesNameP = A.char '-' $> ChatName CTLocal ""
```

epoberezkin: "-" is a name prefix, should be used in all APIs

src/Simplex/Chat.hs
```diff
@@ -6317,6 +6400,7 @@ chatCommandP =
       ("/leave " <|> "/l ") *> char_ '#' *> (LeaveGroup <$> displayName),
       ("/delete #" <|> "/d #") *> (DeleteGroup <$> displayName),
       ("/delete " <|> "/d ") *> char_ '@' *> (DeleteContact <$> displayName),
+      "/clear -" $> ClearNoteFolder "-",
```
epoberezkin, line 6403:
```suggestion
      "/clear -" $> ClearNoteFolder "",
```

src/Simplex/Chat.hs
```diff
@@ -6440,7 +6525,7 @@ chatCommandP =
     incognitoOnOffP = (A.space *> "incognito=" *> onOffP) <|> pure False
     imagePrefix = (<>) <$> "data:" <*> ("image/png;base64," <|> "image/jpg;base64,")
     imageP = safeDecodeUtf8 <$> ((<>) <$> imagePrefix <*> (B64.encode <$> base64P))
-    chatTypeP = A.char '@' $> CTDirect <|> A.char '#' $> CTGroup <|> A.char ':' $> CTContactConnection
+    chatTypeP = A.char '@' $> CTDirect <|> A.char '#' $> CTGroup <|> A.char '$' $> CTLocal <|> A.char ':' $> CTContactConnection
```
epoberezkin, line 6528:
```suggestion
    chatTypeP = A.char '@' $> CTDirect <|> A.char '#' $> CTGroup <|> A.char '-' $> CTLocal <|> A.char ':' $> CTContactConnection
```

src/Simplex/Chat.hs
```diff
@@ -6350,22 +6434,23 @@ chatCommandP =
       ("/connect" <|> "/c") *> (Connect <$> incognitoP <* A.space <*> ((Just <$> strP) <|> A.takeTill isSpace $> Nothing)),
       ("/connect" <|> "/c") *> (AddContact <$> incognitoP),
       SendMessage <$> chatNameP <* A.space <*> msgTextP,
+      "/- " *> (SendMessage (ChatName CTLocal "-") <$> msgTextP),
```
epoberezkin, line 6437:
```suggestion
      "/- " *> (SendMessage (ChatName CTLocal "") <$> msgTextP),
```

tests/ChatTests/Local.hs
spaced4ndy: exactly, that's why I'm saying that chat should have more items - currently test doesn't check different cases of chat pagination

spaced4ndy: also as far as I could tell there're only tests for chat list pagination and no tests of chat pagination

dpwiz: I added chat items test.

This one I don't get. This test has only notes as mixing with the others is tested in the combined chat list test.
There are 2 notes and some of the `[]` results are where the notes folder got hidden entirely instead of showing "first" note. And this is correct behavior since the folder timestamp is indeed outside the range despite having older items in there (otherwise chat list pagination wouldn't work, showing all kinds of stuff all the time).

spaced4ndy: > I added chat items test.

ok, I was referring to the initial comment before this test was added

tests/ChatClient.hs
```diff
@@ -191,7 +192,10 @@ groupLinkViaContactVRange = mkVersionRange 1 2
 createTestChat :: FilePath -> ChatConfig -> ChatOpts -> String -> Profile -> IO TestCC
 createTestChat tmp cfg opts@ChatOpts {coreOptions = CoreChatOpts {dbKey}} dbPrefix profile = do
   Right db@ChatDatabase {chatStore} <- createChatDatabase (tmp </> dbPrefix) dbKey False MCError
-  Right user <- withTransaction chatStore $ \db' -> runExceptT $ createUserRecord db' (AgentUserId 1) profile True
+  Right user <- withTransaction chatStore $ \db' -> runExceptT $ do
+    user <- createUserRecord db' (AgentUserId 1) profile True
+    createNoteFolder db' user
+    pure user
```
spaced4ndy, line 198: probably better to revert, this would cause multiple tests to fail and a lot of noise to fix them

src/Simplex/Chat/Migrations/M20240102_note_folders.hs
```diff
@@ -0,0 +1,41 @@
...
+      favorite INTEGER NOT NULL DEFAULT 0,
+      unread_chat INTEGER NOT NULL DEFAULT 0
+    );
+
+    ALTER TABLE chat_items ADD COLUMN note_folder_id INTEGER DEFAULT NULL REFERENCES note_folders ON DELETE CASCADE;
+    ALTER TABLE files ADD COLUMN note_folder_id INTEGER DEFAULT NULL REFERENCES note_folders ON DELETE CASCADE;
+
+    CREATE INDEX chat_items_note_folder_id ON chat_items(note_folder_id);
+    CREATE INDEX files_note_folder_id ON files(note_folder_id);
+    CREATE UNIQUE INDEX note_folders_user_id ON note_folders(user_id); -- drop UNIQUE for more folders per user
```
spaced4ndy, line 26: non unique was ok, also no need to check constraint in query - we plan to add more note chats soon (tips)
```suggestion
    CREATE INDEX note_folders_user_id ON note_folders(user_id);
```

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -967,11 +1104,102 @@ getGroupChatBefore_ db user@User {userId} g@GroupInfo {groupId} beforeChatItemId
...
+          i.chat_item_id, i.item_ts, i.item_sent, i.item_content, i.item_text, i.item_status, i.shared_msg_id, i.item_deleted, i.item_deleted_ts, i.item_edited, i.created_at, i.updated_at, i.timed_ttl, i.timed_delete_at, i.item_live,
+          -- CIFile
+          f.file_id, f.file_name, f.file_size, f.file_path, f.file_crypto_key, f.file_crypto_nonce, f.ci_file_status, f.protocol
+        FROM chat_items i
+        LEFT JOIN files f ON f.chat_item_id = i.chat_item_id
+        WHERE i.user_id = ? AND i.note_folder_id = ? AND i.item_text LIKE '%' || ? || '%'
+        ORDER BY i.created_at DESC, i.chat_item_id DESC
+        LIMIT ?
+      |]
+      (userId, noteFolderId, search, count)
```
spaced4ndy, line 1141: no need to select all items fields - the initial approach of selecting only IDs and reusing getLocalChatItem in this commit https://github.com/simplex-chat/simplex-chat/pull/3568/commits/211a440e7892021b9c97a629baa8f7a5b01752d1 was ok. I only proposed to change filters and ORDER BY, not fully change queries. The direct chat queries were just never refactored same as groups that's why they still have full lists of fields

src/Simplex/Chat/Store/NoteFolders.hs
```diff
@@ -0,0 +1,65 @@
...
+
+createNoteFolder :: DB.Connection -> User -> ExceptT StoreError IO ()
+createNoteFolder db User {userId} =
+  checkConstraint SENoteFolderAlreadyCreated . liftIO $
+    DB.execute db [sql| INSERT INTO note_folders (user_id) VALUES (?) |] (Only userId)
+
+getNoteFolderIdByName :: DB.Connection -> User -> NoteFolderName -> ExceptT StoreError IO NoteFolderId
+getNoteFolderIdByName db User {userId} ldn =
+  ExceptT . firstRow fromOnly (SENoteFolderNotFoundByName ldn) $
+    DB.query db [sql| SELECT note_folder_id FROM note_folders WHERE user_id = ? AND "-" = ? |] (userId, ldn)
```
spaced4ndy, line 26:
```suggestion
    DB.query db "SELECT note_folder_id FROM note_folders WHERE user_id = ? AND "-" = ?" (userId, ldn)
```

spaced4ndy: what's this filter? `"-" = ?`

spaced4ndy: actually the whole function doesn't seem to make sense right now as they don't have name

src/Simplex/Chat/Store/NoteFolders.hs
```diff
@@ -0,0 +1,65 @@
...
+import Simplex.Chat.Store.Shared (StoreError (..), checkConstraint)
+import Simplex.Chat.Types (NoteFolder (..), NoteFolderId, NoteFolderName, User (..))
+import Simplex.Messaging.Agent.Protocol (UserId)
+import Simplex.Messaging.Agent.Store.SQLite (firstRow)
+import qualified Simplex.Messaging.Agent.Store.SQLite.DB as DB
+
+createNoteFolder :: DB.Connection -> User -> ExceptT StoreError IO ()
+createNoteFolder db User {userId} =
+  checkConstraint SENoteFolderAlreadyCreated . liftIO $
+    DB.execute db [sql| INSERT INTO note_folders (user_id) VALUES (?) |] (Only userId)
```
spaced4ndy, line 21: I think everywhere where it's single line current code uses just quotes, why change it
```suggestion
    DB.execute db "INSERT INTO note_folders (user_id) VALUES (?)" (Only userId)
```

tests/ChatTests/Local.hs
```diff
@@ -0,0 +1,145 @@
...
+  alice /- "memento mori"
+  alice /- "knock-knock"
+  alice /- "who's there?"
+
+  alice #$> ("/_get chat $1 count=100", chat, [(1, "hello world"), (1, "memento mori"), (1, "knock-knock"), (1, "who's there?")])
+  alice #$> ("/_get chat $1 count=1", chat, [(1, "who's there?")])
+  alice #$> ("/_get chat $1 after=2 count=10", chat, [(1, "knock-knock"), (1, "who's there?")])
+  alice #$> ("/_get chat $1 after=2 count=2", chat, [(1, "knock-knock"), (1, "who's there?")])
+  alice #$> ("/_get chat $1 before=3 count=10", chat, [(1, "hello world"), (1, "memento mori")])
+  alice #$> ("/_get chat $1 before=3 count=2", chat, [(1, "hello world"), (1, "memento mori")])
```
spaced4ndy, line 95: still doesn't seem to test that after works differently from last
```suggestion
  alice #$> ("/_get chat $1 after=2 count=10", chat, [(1, "knock-knock"), (1, "who's there?")])
  alice #$> ("/_get chat $1 after=2 count=2", chat, [(1, "knock-knock"), (1, "who's there?")])
  alice #$> ("/_get chat $1 after=1 count=2", chat, [(1, "memento mori"), (1, "knock-knock")])
  alice #$> ("/_get chat $1 before=3 count=10", chat, [(1, "hello world"), (1, "memento mori")])
  alice #$> ("/_get chat $1 before=3 count=2", chat, [(1, "hello world"), (1, "memento mori")])
  alice #$> ("/_get chat $1 before=4 count=2", chat, [(1, "memento mori"), (1, "knock-knock")])
```

src/Simplex/Chat/Store/NoteFolders.hs
dpwiz: They have have the name, it just happens to be empty. I left the check specifically to prevent future confusion when non-empty names would be added.

src/Simplex/Chat/Migrations/M20240102_note_folders.hs
dpwiz: But it *is* unique. And this check caught me a few bugs already.

src/Simplex/Chat/Store/Messages.hs
dpwiz: Ouch... Got confused by "better to repeat logic from direct chats here".

src/Simplex/Chat/Migrations/M20240102_note_folders.hs
spaced4ndy: It's single for now but making it unique seems wrong - we'll soon add tips and user would have multiple local chats. So it's not the record in table that would be one per user making user_id unique, but this specific type of local chats (notes) that would, but we don't even have the field to differentiate them by yet. That some functions had bugs now doesn't mean we should keep constraint in schema and plan to recreate restrictive index later.

dpwiz: ok, moved to explicit check in `createNoteFolder`.

src/Simplex/Chat/Migrations/M20240102_note_folders.hs
dpwiz: It is using `user_id` index, then "temorary b-tree for ordering". Looks legit.

src/Simplex/Chat/Messages.hs
```diff
@@ -351,11 +351,13 @@ data CIMeta (c :: ChatType) (d :: MsgDirection) = CIMeta
...
-  let editable = case itemContent of
-        CISndMsgContent _ -> diffUTCTime currentTs itemTs < nominalDay && isNothing itemDeleted
-        _ -> False
+mkCIMeta :: ChatItemId -> CIDirection c d -> CIContent d -> Text -> CIStatus d -> Maybe SharedMsgId -> Maybe (CIDeleted c) -> Bool -> Maybe CITimed -> Maybe Bool -> UTCTime -> ChatItemTs -> Maybe GroupMemberId -> UTCTime -> UTCTime -> CIMeta c d
+mkCIMeta itemId cd itemContent itemText itemStatus itemSharedMsgId itemDeleted itemEdited itemTimed itemLive currentTs itemTs forwardedByMember createdAt updatedAt =
+  let editable = case cd of
+        CILocalSnd -> isNothing itemDeleted
+        _ -> case itemContent of
+          CISndMsgContent _ -> diffUTCTime currentTs itemTs < nominalDay && isNothing itemDeleted
+          _ -> False
```
epoberezkin, line 360:
```suggestion
mkCIMeta :: forall c d. ChatTypeI c => ChatItemId -> CIContent d -> Text -> CIStatus d -> Maybe SharedMsgId -> Maybe (CIDeleted c) -> Bool -> Maybe CITimed -> Maybe Bool -> UTCTime -> ChatItemTs -> Maybe GroupMemberId -> UTCTime -> UTCTime -> CIMeta c d
mkCIMeta itemId itemContent itemText itemStatus itemSharedMsgId itemDeleted itemEdited itemTimed itemLive currentTs itemTs forwardedByMember createdAt updatedAt =
  let editable =  case itemContent of
          CISndMsgContent _ | chatTypeI @c == SCTLocal -> isNothing itemDeleted
          CISndMsgContent _ -> diffUTCTime currentTs itemTs < nominalDay && isNothing itemDeleted
          _ -> False
```

PR #3588, opened 2023-12-23
core: use version from config, add tests

epoberezkin:


spaced4ndy (dismissed):

src/Simplex/Chat/Protocol.hs
```diff
@@ -52,9 +52,13 @@ import Simplex.Messaging.Parsers (defaultJSON, dropPrefix, fromTextField_, fstTo
 import Simplex.Messaging.Util (eitherToMaybe, safeDecodeUtf8, (<$?>))
 import Simplex.Messaging.Version hiding (version)
 
+-- This should not be used directly in code, instead use `maxVersion chatVRange` from ChatConfig.
+-- This indirection is needed for backward/forward compatibility testing.
+-- Testing with real app versions is still needed, as the test only test current code with different version ranges, not old code.
```
spaced4ndy, line 57:
```suggestion
-- Testing with real app versions is still needed, as the tests only test current code with different version ranges, not old code.
```

PR #3598, opened 2023-12-25
android, desktop: self destruct becomes better

avently:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/SimpleXAPI.kt
```diff
@@ -2141,7 +2141,15 @@ class SharedPreference<T>(val get: () -> T, set: (T) -> Unit) {
...
-      _state.value = value
+      try {
+        _state.value = value
+      } catch (e: IllegalStateException) {
+        // Can be `Reading a state that was created after the snapshot was taken or in a snapshot that has not yet been applied`
+        Log.i(TAG, e.stackTraceToString())
+        withApi {
+          _state.value = value
+        }
+      }
```
epoberezkin, line 2152: 🥺

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/database/DatabaseView.kt
```diff
@@ -459,6 +459,24 @@ suspend fun deleteChatAsync(m: ChatModel) {
   m.controller.apiDeleteStorage()
   DatabaseUtils.ksDatabasePassword.remove()
   m.controller.appPrefs.storeDBPassphrase.set(true)
+  deleteChatManually()
+}
+
+fun deleteChatManually() {
```
epoberezkin, line 465:
```suggestion
fun deleteChatDatabaseFiles() {
```

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/SimpleXAPI.kt
```diff
@@ -2141,7 +2141,15 @@ class SharedPreference<T>(val get: () -> T, set: (T) -> Unit) {
   init {
     this.set = { value ->
       set(value)
-      _state.value = value
+      try {
+        _state.value = value
+      } catch (e: IllegalStateException) {
+        // Can be `Reading a state that was created after the snapshot was taken or in a snapshot that has not yet been applied`
+        Log.i(TAG, e.stackTraceToString())
+        withApi {
```
epoberezkin, line 2149: some strange hack...

avently: state set can only happen in main thread but the function to set state is called in different threads. But not always. So this hack allows to continue using the code like we use. The downside of it is, for example, if we set state and then read it afterwards from .state variable. It will be incorrect for a period of launching coroutine in main thread. But it's not a big deal since to get a state statically we use get(), and we use .state for composing composables so it's ok

avently: maybe the whole SharedPreference<T> should be reformatted to using MutableStateFlow or similar thing but it will require rewrite a lot of code

PR #3626, opened 2023-12-30
cli: prompt for database key entry if required

epoberezkin:


src/Simplex/Chat/Terminal.hs
```diff
@@ -40,18 +44,29 @@ terminalChatConfig =
...
+    run opts@ChatOpts {coreOptions = coreOptions@CoreChatOpts {dbKey}} =
+      handle checkDBKeyError . simplexChatCore cfg opts $ \u cc -> do
+        ct <- newChatTerminal t opts
+        when (firstTime cc) . printToTerminal ct $ chatWelcome u
+        runChatTerminal ct cc opts
+      where
+        checkDBKeyError :: SQLError -> IO ()
+        checkDBKeyError e = case sqlError e of
+          DB.ErrorNotADatabase -> do
+            putStrLn $ "Database file is invalid or " <> if BA.null dbKey then "encrypted." else "you passed an incorrect encryption key."
```
spaced4ndy, line 58:
```suggestion
            if BA.null dbKey
              then putStrLn "Database file is invalid or encrypted."
              else putStrLn "Database file is invalid or you passed an incorrect encryption key."
```

PR #3639, opened 2024-01-03
rfc: handling multiple files/attachments in a message

epoberezkin:


docs/rfcs/2024-01-02-multiple-attachments.md
```diff
@@ -0,0 +1,39 @@
...
+
+The immediate problem we need to solve is encrypting local videos, the absense of which creates a lot of confusion and perception that the videos are sent unencrypted.
+
+The reason videos are not encrypted is because they are usually large files that would be slow to decrypt in order to play, and they are also used to generate previews on the fly. The latter is slow anyway, and causes bad rendering experience.
+
+Videos on iOS are compressed from 5.4.2, and the solution to keep them encrypted is to keep video preview in a separate file. It can be done either on the receiving side, when the video is first rendered, or on the sending side, in which case we can send the preview as a separate file.
+
+In general, attaching multiple files to a message could be beneficial for other cases, such as sending hi- and low-res images at the same time, or sending long-form messages with attachments.
+
+## Solutions
```
spaced4ndy, line 13: another approach could be to use preview from MCVideo. Is the concern that it's too low res?

pros are that it doesn't require any additions to protocol, or implementing attachments until use cases are clear. Possibly they could be blurred in UI a bit prettier than current implementation

PR #3640, opened 2024-01-03
ios: self destruct improvements

avently:


apps/ios/SimpleXChat/FileUtils.swift
```diff
@@ -69,13 +69,29 @@ func fileModificationDate(_ path: String) -> Date? {
     }
 }
 
+public func deleteAppDatabaseAndFiles() {
+    let fm = FileManager.default
+    let dbPath = getAppDatabasePath().path
+    do {
+        try fm.removeItem(atPath: dbPath + CHAT_DB)
+        try fm.removeItem(atPath: dbPath + AGENT_DB)
```
spaced4ndy, line 77: why are only these in try catch?

epoberezkin: others won't through too. Not sure why like this :)

avently: Because in case error happens here, would be useful to see it in logs. Bak files might not exist yet, so don't care about possible error of deleting them

apps/ios/Shared/ContentView.swift
```diff
@@ -73,8 +73,7 @@ struct ContentView: View {
                     showSetPasscode = false
                     alertManager.showAlert(laPasscodeNotSetAlert())
                 }
-            }
-            if chatModel.chatDbStatus == nil {
+            } else if chatModel.chatDbStatus == nil && AppChatState.shared.value != .stopped {
```
spaced4ndy, line 76: this or something else breaks authentication somehow.. try to fail authentication and app will show initialization view instead of unlock button

epoberezkin: this happens if you close the app, not just put to background

avently: https://github.com/simplex-chat/simplex-chat/pull/3640/commits/4b1b325896d27801be8fc78540d0f3c5750a877d


Looks like without another flag is impossible to distinguish states properly to not lose initialization view when it's needed. Now it should be ok

apps/ios/Shared/ContentView.swift
```diff
@@ -61,6 +62,9 @@ struct ContentView: View {
             }
             if !showSettings, let la = chatModel.laRequest {
                 LocalAuthView(authRequest: la)
+                    .onDisappear {
+                        shouldRetryAuth = !accessAuthenticated
```
spaced4ndy, line 66:
```suggestion
                        // this flag is separate from accessAuthenticated to show initializationView while we wait for authentication
                        shouldRetryAuth = !accessAuthenticated
```

apps/ios/Shared/ContentView.swift
```diff
@@ -73,8 +77,7 @@ struct ContentView: View {
                     showSetPasscode = false
                     alertManager.showAlert(laPasscodeNotSetAlert())
                 }
-            }
-            if chatModel.chatDbStatus == nil {
+            } else if chatModel.chatDbStatus == nil && AppChatState.shared.value != .stopped && !shouldRetryAuth {
```
spaced4ndy, line 81:
```suggestion
            } else if chatModel.chatDbStatus == nil && AppChatState.shared.value != .stopped && !authAttemptedAndFailed {
```

apps/ios/Shared/ContentView.swift
```diff
@@ -31,6 +31,7 @@ struct ContentView: View {
     @State private var showWhatsNew = false
     @State private var showChooseLAMode = false
     @State private var showSetPasscode = false
+    @State private var shouldRetryAuth = false
```
spaced4ndy, line 34:
```suggestion
    @State private var authAttemptedAndFailed = false
```

apps/ios/Shared/ContentView.swift
```diff
@@ -61,6 +62,10 @@ struct ContentView: View {
             }
             if !showSettings, let la = chatModel.laRequest {
                 LocalAuthView(authRequest: la)
+                    .onDisappear {
+                        // this flag is separate from accessAuthenticated to show initializationView while we wait for authentication
+                        shouldRetryAuth = !accessAuthenticated
```
spaced4ndy, line 67:
```suggestion
                        authAttemptedAndFailed = !accessAuthenticated
```

PR #3648, opened 2024-01-05
xftp: add controlPort field

dpwiz:
build fix for https://github.com/simplex-chat/simplexmq/pull/951


epoberezkin:
added already...

PR #3654, opened 2024-01-06
core: support deleting the last profile (always create user record in agent when user is created)

epoberezkin:


src/Simplex/Chat/Core.hs
```diff
@@ -44,3 +70,59 @@ sendChatCmdStr cc s = runReaderT (execChatCommand Nothing . encodeUtf8 $ T.pack
...
+                    withTransaction st (`setActiveUser` userId)
+                    pure $ Just user
+
+createActiveUser :: ChatController -> IO User
+createActiveUser cc = do
+  putStrLn
+    "No user profiles found, it will be created now.\n\
+    \Please choose your display name and your full name.\n\
+    \They will be sent to your contacts when you connect.\n\
+    \They are only stored on your device and you can change them later."
```
spaced4ndy, line 109:
```suggestion
    "No user profiles found, it will be created now.\n\
    \Please choose your display name.\n\
    \It will be sent to your contacts when you connect.\n\
    \It is only stored on your device and you can change it later."
```

src/Simplex/Chat/Terminal.hs
```diff
@@ -2,6 +2,7 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TupleSections #-}
```
spaced4ndy, line 5:
```suggestion
```

tests/ChatClient.hs
```diff
@@ -190,7 +192,8 @@ groupLinkViaContactVRange = mkVersionRange 1 2
 
 createTestChat :: FilePath -> ChatConfig -> ChatOpts -> String -> Profile -> IO TestCC
 createTestChat tmp cfg opts@ChatOpts {coreOptions = CoreChatOpts {dbKey}} dbPrefix profile = do
-  Right db@ChatDatabase {chatStore} <- createChatDatabase (tmp </> dbPrefix) dbKey False MCError
+  Right db@ChatDatabase {chatStore, agentStore} <- createChatDatabase (tmp </> dbPrefix) dbKey False MCError
+  withTransaction agentStore (`DB.execute_` "INSERT INTO users (user_id) VALUES (1);")
   Right user <- withTransaction chatStore $ \db' -> runExceptT $ createUserRecord db' (AgentUserId 1) profile True
```
spaced4ndy, line 197: why not call api as in createActiveUser? probably it has to be moved to where controller is created then..

PR #3655, opened 2024-01-06
ios: fix callkit calls via NSE

epoberezkin:


apps/ios/Shared/Views/Call/CallController.swift
```diff
@@ -142,45 +142,57 @@ class CallController: NSObject, CXProviderDelegate, PKPushRegistryDelegate, Obse
...
-        logger.debug("CallController: initializing chat")
+        if AppChatState.shared.value == .stopped {
+            self.reportExpiredCall(payload: payload, completion)
+            return
+        }
         if (!ChatModel.shared.chatInitialized) {
-            initChatAndMigrate(refreshInvitations: false)
+            logger.debug("CallController: initializing chat")
+            do {
+                try initializeChat(start: true, refreshInvitations: false, withChatData: false)
```
spaced4ndy, line 161: doesn't compile with extra argument error
```suggestion
                try initializeChat(start: true, refreshInvitations: false)
```

apps/ios/Shared/Model/SuspendChat.swift
```diff
@@ -134,20 +136,33 @@ func initChatAndMigrate(refreshInvitations: Bool = true) {
     }
 }
 
-func startChatAndActivate(dispatchQueue: DispatchQueue = DispatchQueue.main, _ completion: @escaping () -> Void) {
+func startChatForCall() {
```
spaced4ndy, line 139: add a comment that the main difference with startChatAndActivate is that it doesn't wait for NSE? If I understand correctly. A bit hard to see behind different calls to activateChat(), via activate() in startChatAndActivate which does the same but with completion()

apps/ios/SimpleX.xcodeproj/xcshareddata/xcschemes/SimpleX (iOS).xcscheme
```diff
@@ -41,7 +41,7 @@
       </Testables>
    </TestAction>
    <LaunchAction
-      buildConfiguration = "Release"
+      buildConfiguration = "Debug"
```
spaced4ndy, line 44: don't forget to revert

epoberezkin: better keep as is

PR #3661, opened 2024-01-09
ios: fix "mark read" chat preview button not always marking unread items as read

spaced4ndy:


apps/ios/Shared/Model/SimpleXAPI.swift
```diff
@@ -844,7 +844,7 @@ func apiRejectContactRequest(contactReqId: Int64) async throws {
     throw r
 }
 
-func apiChatRead(type: ChatType, id: Int64, itemRange: (Int64, Int64)) async throws {
+func apiChatRead(type: ChatType, id: Int64, itemRange: (Int64, Int64)? = nil) async throws {
```
epoberezkin, line 847: Don't think it's a good idea to use this API, we don't use it.

There is minUnreadItemId in ChatStats, and the last one is known from preview, so we can use them.

spaced4ndy:
perhaps a more thorough rework of "mark read" logic is required, a few issues to consider:
- query used on api without ranges uses idx_chat_items_user_id_item_status index - without contact_id/group_id
- adding new indexes (e.g. user_id, contact_id, item_status) changes plan for chat pagination
- minUnreadItemId is not updated on new item in iOS, and is seemingly incorrectly updated in Android
- it's overall not correct to base ranges solely on IDs because pagination is based on timestamps, at least timestamps have to be read by IDs similar to group chat pagination (direct chat pagination should be fixed in similar way)
- in case of "mark read" ranges situation is worse than pagination, because fromId is often 0 (no update on ios / coalesce to 0 when read), so it should be translated into "before" without limit
- an example of adequate approach could be:
  - remove range based based api, always use api to read all (buttons) or single item/list of items (scrolling)
  - add necessary indexes without breaking chat pagination (direct chat pagination has also to be fixed so it'd require a different index from the one used now)
  - remove minUnreadItemId from chat stats, remove "aboveItem" logic

PR #3668, opened 2024-01-10
android: don't show alert when not needed after changing user concurrently

avently:


avently:
unneeded since I made https://github.com/simplex-chat/simplex-chat/pull/3669 which makes this problem absent

Safasalxsa16:
Hello My beautiful man! good luck for work

PR #3677, opened 2024-01-12
core: correct pagination (wip)

spaced4ndy:


spaced4ndy:
- pagination shouldn't be based on chat_item_id, because it has gaps
- pagination on timestamp also has its downsides, e.g. creating multiple records on a single timestamp, clock change
- perhaps we should add separate counter tables
TBC in larger scope of pagination and scroll imrpovements

PR #3689, opened 2024-01-15
core: block member for all

spaced4ndy:


src/Simplex/Chat.hs
```diff
@@ -1724,6 +1733,30 @@ processChatCommand' vr = \case
                 ci <- saveSndChatItem user (CDGroupSnd gInfo) msg (CISndGroupEvent gEvent)
                 toView $ CRNewChatItem user (AChatItem SCTGroup SMDSnd (GroupChat gInfo) ci)
           pure CRMemberRoleUser {user, groupInfo = gInfo, member = m {memberRole = memRole}, fromRole = mRole, toRole = memRole}
+  APIBlockMemberForAll groupId gmId blocked -> withUser $ \user -> do
+    Group gInfo members <- withStore $ \db -> getGroup db vr user groupId
+    let (targetMember, remainingMembers) = partition ((== gmId) . groupMemberId') members
```
epoberezkin, line 1738: something like `span` that needs to be used here, not `partition`.

epoberezkin: `removeQP` in simplexmq is similar, can be done on a normal list too.

src/Simplex/Chat.hs
```diff
@@ -1724,6 +1733,30 @@ processChatCommand' vr = \case
...
+              bm' <- withStore $ \db -> do
+                liftIO $ updateGroupMemberBlockedForAll db user groupId gmId byGMId_
+                getGroupMember db user groupId gmId
+              toggleNtf user bm' (not blocked)
+              pure CRMemberBlockedForAllUser {user, groupInfo = gInfo, member = bm', blocked}
+            else throwChatError $ CECommandError $ if blocked then "already blocked" else "already unblocked"
+        where
+          byGMId_ = if blocked then Just (groupMemberId' membership) else Nothing
+          GroupInfo {membership} = gInfo
+          GroupMember {memberId = bmMemberId, memberRole = bmRole, memberProfile = bmp, blockedByGroupMemberId} = blockedMember
```
epoberezkin, line 1758: this seems wrong to overuse this matching in where when it fits well in the case

src/Simplex/Chat.hs
```diff
@@ -1724,6 +1733,30 @@ processChatCommand' vr = \case
...
+              toView $ CRNewChatItem user (AChatItem SCTGroup SMDSnd (GroupChat gInfo) ci)
+              bm' <- withStore $ \db -> do
+                liftIO $ updateGroupMemberBlockedForAll db user groupId gmId byGMId_
+                getGroupMember db user groupId gmId
+              toggleNtf user bm' (not blocked)
+              pure CRMemberBlockedForAllUser {user, groupInfo = gInfo, member = bm', blocked}
+            else throwChatError $ CECommandError $ if blocked then "already blocked" else "already unblocked"
+        where
+          byGMId_ = if blocked then Just (groupMemberId' membership) else Nothing
+          GroupInfo {membership} = gInfo
```
epoberezkin, line 1757: even more so with this one - move to to where it is

src/Simplex/Chat.hs
epoberezkin: that some long pattern matches can to be done separately, doesn't mean that it has to always be done like that -the default style is to match in place.

epoberezkin: blockedByGroupMemberId, in particular, can be just used as a predicate to `m`

src/Simplex/Chat.hs
```diff
@@ -1724,6 +1733,30 @@ processChatCommand' vr = \case
...
+        assertUserGroupRole gInfo $ max GRAdmin bmRole
+        withChatLock "blockForAll" . procCmd $
+          if blocked /= isJust blockedByGroupMemberId
+            then do
+              (msg, _) <- sendGroupMessage' user gInfo remainingMembers $ XGrpMemBlock bmMemberId blocked
+              let ciContent = CISndGroupEvent $ SGEMemberBlocked gmId (fromLocalProfile bmp) blocked
+              ci <- saveSndChatItem user (CDGroupSnd gInfo) msg ciContent
+              toView $ CRNewChatItem user (AChatItem SCTGroup SMDSnd (GroupChat gInfo) ci)
+              bm' <- withStore $ \db -> do
+                liftIO $ updateGroupMemberBlockedForAll db user groupId gmId byGMId_
```
epoberezkin, line 1750: 6 words

src/Simplex/Chat.hs
```diff
@@ -4505,43 +4544,54 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+      gInfo@GroupInfo {groupId}
+      m@GroupMember {memberId, memberRole, blockedByGroupMemberId}
+      mc
+      msg@RcvMessage {sharedMsgId_}
+      brokerTs
+      forwarded
+        | isVoice content && not (groupFeatureAllowed SGFVoice gInfo) = rejected GFVoice
+        | not (isVoice content) && isJust fInv_ && not (groupFeatureAllowed SGFFiles gInfo) = rejected GFFiles
+        | otherwise =
+            case blockedByGroupMemberId of
```
epoberezkin, line 4557: selector

src/Simplex/Chat.hs
```diff
@@ -4505,43 +4544,54 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+              Just blockedByGMId ->
+                withStore' (\db -> runExceptT $ getGroupMember db user groupId blockedByGMId) >>= \case
+                  Right blockedByGM -> applyModeration' timed' live' blockedByGM Nothing brokerTs
+                  Left _ -> createItem timed' live'
```
epoberezkin, line 4561: this could be a function, so it's clear what you do

src/Simplex/Chat.hs
```diff
@@ -5315,6 +5365,29 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
     checkHostRole GroupMember {memberRole, localDisplayName} memRole =
       when (memberRole < GRAdmin || memberRole < memRole) $ throwChatError (CEGroupContactRole localDisplayName)
 
+    xGrpMemBlock :: GroupInfo -> GroupMember -> MemberId -> Bool -> RcvMessage -> UTCTime -> m ()
+    xGrpMemBlock gInfo@GroupInfo {groupId, membership} m@GroupMember {memberRole = senderRole} memId blocked msg brokerTs
+      | membershipMemId == memId =
+          -- member shouldn't receive this message about themselves
+          messageError "x.grp.mem.block: admin blocks you"
```
epoberezkin, line 5372: this would show in terminal, which is probably not something we want? although may be ok as low priority error that will be hidden by default...

src/Simplex/Chat.hs
```diff
@@ -5315,6 +5365,29 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+      | otherwise =
+          withStore' (\db -> runExceptT $ getGroupMemberByMemberId db user gInfo memId) >>= \case
```
epoberezkin, line 5374: as you fail on absent member anyway, maybe no need to handle it here, and can just use withStore without runExceptT?

epoberezkin: also, can there be a race between member being informed about all members and some members being blocked? In which case do we need to store these blockings? or, at least, include the information about member being blocked in introductions? also the member can be blocked before new member joins - I don't see it handled.

epoberezkin: So you still end up with the need to communicate blocking as part of profile?

tests/ChatTests/Groups.hs
```diff
@@ -141,6 +141,11 @@ chatGroupTests = do
     it "member contact is deleted silently, then considered disabled" testMembershipProfileUpdateContactDisabled
     it "profile update without change is ignored" testMembershipProfileUpdateNoChangeIgnored
     it "change of profile contact link is ignored" testMembershipProfileUpdateContactLinkIgnored
+  describe "block member for all" $ do
+    it "messages are marked as moderated" testBlockForAllModerated
+    it "messages are fully deleted" testBlockForAllFullDelete
+    it "another admin can unblock" testBlockForAllAnotherAdminUnblocks
+    it "can't repeat block/unblock" testBlockForAllCantRepeat
```
epoberezkin, line 148: does it test for members blocked before new member joins?

src/Simplex/Chat.hs
spaced4ndy: span/break seem to do a different thing. Could use `find` but then would need to separately filter out element for remaining members

src/Simplex/Chat.hs
spaced4ndy: matching in place results in very long lines. moved to `let`s

src/Simplex/Chat.hs
spaced4ndy: this shouldn't happen, it may also indicate a bug to us

tests/ChatTests/Groups.hs
spaced4ndy: no, I've missed this case

src/Simplex/Chat/View.hs
```diff
@@ -431,21 +434,16 @@ responseToView hu@(currentRH, user_) ChatConfig {logLevel, showReactions, showRe
     unmuted' :: User -> ChatInfo c -> CIDirection c d -> Bool -> [StyledString] -> [StyledString]
     unmuted' u chat chatDir mention s
       | chatDirNtf u chat chatDir mention = s
+      | testView = map (<> " <muted>") s
       | otherwise = []
 
 userNtf :: User -> Bool
 userNtf User {showNtfs, activeUser} = showNtfs || activeUser
 
-chatNtf :: User -> ChatInfo c -> Bool -> Bool
```
spaced4ndy, line 439: was unused

src/Simplex/Chat/Types.hs
```diff
@@ -675,10 +720,14 @@ incognitoMembershipProfile GroupInfo {membership = m@GroupMember {memberProfile}
 memberSecurityCode :: GroupMember -> Maybe SecurityCode
 memberSecurityCode GroupMember {activeConn} = connectionCode =<< activeConn
 
+memberBlocked' :: GroupMember -> Bool
```
spaced4ndy, line 723: todo rename blockedByAdmin

PR #3692, opened 2024-01-16
ios: profile updated chat items

spaced4ndy:


apps/ios/SimpleXChat/ChatTypes.swift
```diff
@@ -3352,14 +3366,43 @@ public enum CIGroupInvitationStatus: String, Decodable {
...
+    case profileUpdated(fromProfile: Profile, toProfile: Profile)
 
     var text: String {
         switch self {
         case .contactDeleted: return NSLocalizedString("deleted contact", comment: "rcv direct event chat item")
+        case let .profileUpdated(fromProfile, toProfile): return profileUpdatedText(fromProfile, toProfile)
+        }
+    }
+
+    private func profileUpdatedText(_ from: Profile, _ to: Profile) -> String {
```
epoberezkin, line 3378:
```suggestion
    private func profileUpdatedText(_ from: Profile, _ to: Profile) -> LocalizeStringKey {
```

or am I missing something?

apps/ios/SimpleXChat/ChatTypes.swift
```diff
@@ -3352,14 +3366,43 @@ public enum CIGroupInvitationStatus: String, Decodable {
...
+            defaultProfileUpdatedText
         }
     }
 }
 
+private var profileUpdatedRemovedPictureText: String = NSLocalizedString("removed profile picture", comment: "rcv direct or group event chat item")
+
+private var profileUpdatedSetNewPictureText: String = NSLocalizedString("set new profile picture", comment: "rcv direct or group event chat item")
+
+private var defaultProfileUpdatedText: String = NSLocalizedString("updated profile", comment: "rcv direct or group event chat item")
```
epoberezkin, line 3404: just define them in-place, they are used twice, and NSLocalizedString is probably not needed there?

apps/ios/SimpleXChat/ChatTypes.swift
```diff
@@ -3390,6 +3434,22 @@ public enum RcvGroupEvent: Decodable {
...
+            if to.image == nil {
+                profileUpdatedRemovedPictureText
+            } else {
+                profileUpdatedSetNewPictureText
+            }
+        } else {
+            // shouldn't happen if backend correctly creates item; UI should be synchronized with backend
+            defaultProfileUpdatedText
         }
     }
```
epoberezkin, line 3454:
```suggestion
    private func profileUpdatedText(_ from: Profile, _ to: Profile) -> LocalizedStringKey {
        to.displayName != from.displayName || to.fullName != from.fullName
        ? "member \(from.profileViewName) changed to \(to.profileViewName)"
        : to.image == from.image
        ? "updated profile" // won't happen
        : to.image == nil
        ? "removed profile picture"
        : "set new profile picture"
    }
```

No need for indirection, returns, ifs here.

Same for function above.

apps/ios/SimpleXChat/ChatTypes.swift
```diff
@@ -2245,7 +2250,16 @@ public struct ChatItem: Identifiable, Decodable {
     public var memberDisplayName: String? {
         get {
             if case let .groupRcv(groupMember) = chatDir {
-                return groupMember.chatViewName
+                switch content {
+                case let .rcvGroupEvent(rcvGroupEvent: .memberProfileUpdated(fromProfile, toProfile)):
+                    if toProfile.displayName != fromProfile.displayName || toProfile.fullName != fromProfile.fullName {
+                        return nil
```
epoberezkin, line 2256: remove `returns` (and `get`, as it's rewritten anyway)

epoberezkin: the first if can be ternary

apps/ios/SimpleXChat/ChatTypes.swift
epoberezkin: It probably needs to be String as others are strings annoyingly, but it's easier to just repeat NSLocalizedString here with a comment than use constants, also easier to make them different if needed.

spaced4ndy: ternary is harder to read as conditions and values are on the same level without indentation.

also this suggestion moves default case in the middle for optimization of one less comparison but makes it hard to read or change

spaced4ndy: this more or less repeats logic in terminal view - check what's changed, then how
https://github.com/simplex-chat/simplex-chat/blob/master/src/Simplex/Chat/View.hs#L1429
except terminal view doesn't account for images and starts with default case

apps/ios/SimpleXChat/ChatTypes.swift
spaced4ndy: again ternary would make everything a single unindented block harder to understand what's a condition and what's a value

PR #3739, opened 2024-01-23
ui: fix chat preview showing incorrect timestamp when chat is empty

spaced4ndy:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
```diff
@@ -1530,7 +1545,8 @@ class NoteFolder(
       favorite = false,
       unread = false,
       createdAt = Clock.System.now(),
-      updatedAt = Clock.System.now()
+      updatedAt = Clock.System.now(),
```
epoberezkin, line 1548: this is probably a bad idea to put current time in sample data... No need to change it now

cc @avently ?

epoberezkin (approved):

PR #3773, opened 2024-01-30
Fix broken english

du82:
This PR fixes multiple issues with the english in the readme so the project is more presentable


jermanuts (commented):
>We are replying the questions manually, so it is not instant – it can take up to 24 hours.

We are replying to the questions manually, so it is not instant – it can take up to 24 hours.

>Please note: The groups below are created for the users to be able to ask questions, make suggestions and ask questions about SimpleX Chat only.

Please note: The groups below are created for the users to be able to ask questions, make suggestions and inquire about SimpleX Chat only. 

>domain, - so anybody

domain,  so anybody

---

Tbh, I think the Readme deserves a clean-up. You can keep stuff that doesn't already exist in the website's docs in the readme. For example: the translators table.

PR #3784, opened 2024-02-06
desktop: change width and height to safe values for storing and restoring

avently:


avently:
We decided to ignore this optimization back then

PR #3796, opened 2024-02-11
docs: update privacy policy

epoberezkin:


PRIVACY.md
```diff
@@ -1,134 +1,154 @@
...
-SimpleX Chat security assessment was done in October 2022 by [Trail of Bits](https://www.trailofbits.com/about), and most fixes were released in v4.2.0 – see [the announcement](./blog/20221108-simplex-chat-v4.2-security-audit-new-website.md).
+SimpleX Chat security assessment was done in October 2022 by [Trail of Bits](https://www.trailofbits.com/about), and most fixes were released in v4.2 – see [the announcement](./blog/20221108-simplex-chat-v4.2-security-audit-new-website.md).
 
-### Information you provide
+### Your information
 
 #### User profiles
 
-We do not store user profiles. The profile you create in the app is local to your device.
+Servers used by SimpleX Chat apps do not create, store or identify user profiles. The profiles you can create in the app are local to your device, and can be removed at any time.
```
easthvan, line 29: Can you please mention SimpleX addresses, 1-time invitation links and the user's "contact list" which are currently the SimpleX addresses of their contacts (if they get it from them) and/or the conversations itself if they are based on 1-time link. 

What part of these three things SMP servers store and needed for clients to reconnect in case they reinstall client app or migrate phones or their orig SMPs go offline etc.

PRIVACY.md
```diff
@@ -1,134 +1,166 @@
...
+
+When you create the local profile, no records are created on any of the relay servers, and infrastructure providers, whether SimpleX Chat Ltd or any other, have no access to any part of your information, and even to the fact that you created a profile - it is a local record stored only on your device. That means that if you delete the app, and have no backup, you will permanently lose all your data and the private connections you created with other users.
 
-When you create a user profile, no records are created on our servers, and we have no access to any part of your profile information, and even to the fact that you created a profile - it is a local record stored only on your device. That means that if you delete the app, and have no backup, you will permanently lose all the data and the private connections you create with other users.
+You can transfer the profile to another device by creating a backup of the app data and restoring it on the new device, but you cannot use to devices with the copy of the same profile at the same time.
```
spaced4ndy, line 33:
```suggestion
You can transfer the profile to another device by creating a backup of the app data and restoring it on the new device, but you cannot use two devices with the copy of the same profile at the same time.
```

spaced4ndy: or better "multiple" or "more than one"

PRIVACY.md
```diff
@@ -1,134 +1,166 @@
...
+
+SimpleX relay servers do not store information about which queues are linked to your profile on the device, and they do not collect any information that would allow infrastructure owners and providers to establish that these queues are related to your device or your profile - the access to each queue is authorized by two anonymous unique cryptographic keys, different for each queue, and separate for sender and recipient of the messages.
+
+#### Connection links privacy
+
+When you create a connection with another user, the app generate a "link" that can be shared with the user to establish the connection via any connection. This link is safe to share via insecure channels, as long as you can identify the recipient and also trust that this channel did not replace this link (to mitigate the latter risk you can validate the security code via the app).
```
spaced4ndy, line 59:
```suggestion
When you create a connection with another user, the app generates a "link" that can be shared with the user to establish the connection via any connection. This link is safe to share via insecure channels, as long as you can identify the recipient and also trust that this channel did not replace this link (to mitigate the latter risk you can validate the security code via the app).
```

PRIVACY.md
```diff
@@ -1,134 +1,166 @@
...
 
-At the time of updating this document all our client applications allow configuring the servers. Our servers do not store information about which queues are linked to your profile on the device, and they do not collect any information that would allow us to establish that these queues are related to your device or your profile - the access to each queue is authorized by two anonymous unique cryptographic keys, different for each queue, and separate for sender and recipient of the messages.
+While the connection "links" contain SimpleX Chat Ltd domain name `simplex.chat`, this site is never accessed by the app, and is only used for these purposes:
+- to direct the new users to the app download instructions,
+- to show connection QR code that can be scanned via the app,
+- and to "namespace" these links,
+- to open links directly in the app when it is clicked outside of the app.
```
spaced4ndy, line 65:
```suggestion
- to direct the new users to the app download instructions,
- to show connection QR code that can be scanned via the app,
- to "namespace" these links,
- and to open links directly in the app when it is clicked outside of the app.
```

PRIVACY.md
```diff
@@ -1,134 +1,166 @@
...
 
 #### Another information stored on the servers
 
-Additional technical information can be stored on our servers, including randomly generated authentication tokens, keys, push tokens, and other material that is necessary to transmit messages. SimpleX Chat limits this additional technical information to the minimum required to operate the Services.
+Additional technical information can be stored on our servers, including randomly generated authentication tokens, keys, push tokens, and other material that is necessary to transmit messages. SimpleX Chat design limits this additional technical information to the minimum required to operate the software and servers. Servers can only store any data that can link to any particular user or device, including IP addresses, geographic location, device identifiers, or any information related the transport sessions, only for the purposes to prevent server overloading and attacks. This information is not stored for the absolute majority of the users, even who use the servers very actively.
```
spaced4ndy, line 79:
```suggestion
Additional technical information can be stored on our servers, including randomly generated authentication tokens, keys, push tokens, and other material that is necessary to transmit messages. SimpleX Chat design limits this additional technical information to the minimum required to operate the software and servers. Servers can store any data that can link to any particular user or device, including IP addresses, geographic location, device identifiers, or any information related the transport sessions, only for the purposes to prevent server overloading and attacks. This information is not stored for the absolute majority of the users, even who use the servers very actively.
```

PRIVACY.md
```diff
@@ -1,134 +1,166 @@
...
 
-**Minimal age**. You must be at least 13 years old to use our Services. The minimum age to use our Services without parental approval may be higher in your country.
+**Client applications**. Our client application Software (referred to as "app" or "apps") also has no modifications compared with published open-source code, and any developers of the alternative client apps based on our code are required to publish any modifications under the same AGPLv3 license. Client applications do not include any tracking or analytics code, and does not share any information with SimpleX Chat Ltd or any other third parties.
```
spaced4ndy, line 120:
```suggestion
**Client applications**. Our client application Software (referred to as "app" or "apps") also has no modifications compared with published open-source code, and any developers of the alternative client apps based on our code are required to publish any modifications under the same AGPLv3 license. Client applications do not include any tracking or analytics code, and do not share any information with SimpleX Chat Ltd or any other third parties.
```

PRIVACY.md
easthvan: My concern is that technically you can use the profile. the database on multiple devices bit you will face issues on your devices quite soon, some messages will not be delivered etc. So it is not that you can not at all. but it is not going to function as intended andvarious unexpected errors you will face and will lose some of your new messages etc. It is or an intended usage form, not recommended and after key rotations it will fail too. 

I think all this should be made clear that if the user clones its database onto more than one device, they will still open and seemingly work for a very limited time etc. For example in Session chat app, the biggest concern is the fact that there you can clone accounts (IDs by the seed) onto an unlimited number of devices and it still works "normally" and no one will ever notice it either.

PR #3821, opened 2024-02-19
core: filter out user contact on merge, delete

spaced4ndy:


spaced4ndy:
closing, solved in #3822

PR #3823, opened 2024-02-19
core, ui: remove usage of inline files (send only xftp files)

spaced4ndy:


src/Simplex/Chat.hs
```diff
@@ -788,10 +730,10 @@ processChatCommand' vr = \case
                   withStore' $
                     \db -> createSndFTDescrXFTP db user (Just m) conn ft fileDescr
               saveMemberFD _ = pure ()
-        pure (fInv, ciFile, ft)
-      unzipMaybe3 :: Maybe (a, b, c) -> (Maybe a, Maybe b, Maybe c)
-      unzipMaybe3 (Just (a, b, c)) = (Just a, Just b, Just c)
-      unzipMaybe3 _ = (Nothing, Nothing, Nothing)
+        pure (fInv, ciFile)
+      unzipMaybe2 :: Maybe (a, b) -> (Maybe a, Maybe b)
```
epoberezkin, line 734: https://hackage.haskell.org/package/base-4.19.0.0/docs/Data-List-NonEmpty.html#v:unzip

src/Simplex/Chat.hs
```diff
@@ -718,53 +686,27 @@ processChatCommand' vr = \case
           | isVoice mc && not (groupFeatureAllowed SGFVoice gInfo) = notAllowedError GFVoice
           | not (isVoice mc) && isJust file_ && not (groupFeatureAllowed SGFFiles gInfo) = notAllowedError GFFiles
           | otherwise = do
-              (fInv_, ciFile_, ft_) <- unzipMaybe3 <$> setupSndFileTransfer g (length $ filter memberCurrent ms)
+              (fInv_, ciFile_) <- unzipMaybe2 <$> setupSndFileTransfer g (length $ filter memberCurrent ms)
```
epoberezkin, line 689:
```suggestion
              (fInv_, ciFile_) <- L.unzip <$> setupSndFileTransfer g (length $ filter memberCurrent ms)
```

epoberezkin, line 689:
```suggestion
              (fInv_, ciFile_) <- unzipMaybe <$> setupSndFileTransfer g (length $ filter memberCurrent ms)
```

tests/ChatClient.hs
```diff
@@ -287,7 +287,10 @@ readTerminalOutput t termQ = do
 withTmpFiles :: IO () -> IO ()
 withTmpFiles =
   bracket_
-    (createDirectoryIfMissing False "tests/tmp")
+    ( do
+        createDirectoryIfMissing False "tests/tmp"
+        createDirectoryIfMissing False "tests/tmp/tmp" -- for XFTP temporary files
```
epoberezkin, line 292: this change should not be needed, it should create this directory. If it doesn't, better to fix it.

spaced4ndy: it's in tests

spaced4ndy: It's set in SetTempFolder API, and then passed to agent only on chat start. Folder is also created in SetTempFolder. So in tests it would require stopping chat, setting folder, and starting chat in each test. There is such test - testXFTPWithRelativePaths. To not clutter all other tests better to just have this directory prepared and avoid this workaround.

spaced4ndy: Previously it was just configured to tests/tmp, which is a tests temp folder and this sometimes conflicts with test logic, for example when database is deleted and it deletes whole tests temp folder instead of only necessary folders, so some tests already configured this additional nested folder as temporary and created it at the start of test.

PR #3826, opened 2024-02-20
core: check user record when deleting contact and display name

spaced4ndy:


src/Simplex/Chat/Store/Shared.hs
```diff
@@ -401,3 +403,17 @@ createWithRandomBytes' size gVar create = tryCreate 3
...
+
+checkContactIsUser :: DB.Connection -> User -> Contact -> IO Bool
+checkContactIsUser db User {userContactId} Contact {contactId} = do
+  isUser_ <-
+    maybeFirstRow fromOnly $
+      DB.query db "SELECT is_user FROM contacts WHERE contact_id = ?" (Only contactId)
+  pure $ fromMaybe False isUser_ || contactId == userContactId
+
+checkLDNIsUser :: DB.Connection -> User -> ContactName -> IO Bool
+checkLDNIsUser db User {userId} ldn = do
```
epoberezkin, line 415: combine to one function

src/Simplex/Chat/Store/Shared.hs
```diff
@@ -110,6 +110,8 @@ data StoreError
   | SERemoteHostDuplicateCA
   | SERemoteCtrlNotFound {remoteCtrlId :: RemoteCtrlId}
   | SERemoteCtrlDuplicateCA
+  | SEProhibitedDeleteUserContact {contactId :: ContactId}
+  | SEProhibitedDeleteUserName {localDisplayName :: ContactName}
```
epoberezkin, line 114: combine to one error - unclear why two are needed

src/Simplex/Chat/Store/Groups.hs
```diff
@@ -1361,7 +1364,8 @@ updateGroupProfile db User {userId} g@GroupInfo {groupId, localDisplayName, grou
         db
         "UPDATE groups SET local_display_name = ?, updated_at = ? WHERE user_id = ? AND group_id = ?"
         (ldn, currentTs, userId, groupId)
-      DB.execute db "DELETE FROM display_names WHERE local_display_name = ? AND user_id = ?" (localDisplayName, userId)
+      unlessM (checkLDNIsUser db user localDisplayName) $
+        DB.execute db "DELETE FROM display_names WHERE local_display_name = ? AND user_id = ?" (localDisplayName, userId)
```
epoberezkin, line 1368: this can be one function that only deleted ldn if it's not a user

epoberezkin: DELETE FROM display_names
WHERE local_display_name = ? AND  user_id = ?
  AND local_display_name != (SELECT local_display_name FROM users where user_id = ?)

src/Simplex/Chat/Store/Groups.hs
```diff
@@ -2030,7 +2038,8 @@ updateMemberProfile db User {userId} m p'
           db
           "UPDATE group_members SET local_display_name = ?, updated_at = ? WHERE user_id = ? AND group_member_id = ?"
           (ldn, currentTs, userId, groupMemberId)
-        DB.execute db "DELETE FROM display_names WHERE local_display_name = ? AND user_id = ?" (localDisplayName, userId)
+        unlessM (checkLDNIsUser db user localDisplayName) $
+          DB.execute db "DELETE FROM display_names WHERE local_display_name = ? AND user_id = ?" (localDisplayName, userId)
```
epoberezkin, line 2042: same as above

src/Simplex/Chat/Store/Shared.hs
epoberezkin: also this query does not account for the possibility of another corruption, where is_user is inconsistent with contact_id in user record

epoberezkin: SELECT 1 FROM users
  WHERE user_id = ?
    AND (local_display_name = ? || contact_id = ?)
LIMIT 1

PR #3834, opened 2024-02-22
chat: store app settings in db

dpwiz:


src/Simplex/Chat/Migrations/M20240222_app_settings.hs
```diff
@@ -0,0 +1,20 @@
...
+module Simplex.Chat.Migrations.M20240222_app_settings where
+
+import Database.SQLite.Simple (Query)
+import Database.SQLite.Simple.QQ (sql)
+
+m20240222_app_settings :: Query
+m20240222_app_settings =
+  [sql|
+CREATE TABLE app_settings (
+  app_settings BLOB NOT NULL
```
epoberezkin, line 12: TEXT?

tests/ChatTests/Direct.hs
```diff
@@ -1113,6 +1113,19 @@ testMaintenanceModeWithSettings tmp = do
       alice ##> "/_stop"
       alice <## "chat stopped"
       let settings = T.unpack . safeDecodeUtf8 . LB.toStrict $ J.encode defaultAppSettings
+      -- app-provided defaults
+      alice ##> ("/_get app settings " <> settings)
+      alice <## ("app settings: " <> settings)
+      -- parser defaults fallback
+      alice ##> "/_get app settings"
+      alice <## ("app settings: " <> settings)
```
spaced4ndy, line 1121: this doesn't test that providing app settings doesn't return default settings, first test should pass modified default settings to api

PR #3847, opened 2024-02-28
docs: PQ integration rfc

spaced4ndy:


docs/rfcs/2024-02-28-pq-integration.md
```diff
@@ -0,0 +1,60 @@
...
+- Send before introductions
+  - new protocol message
+    - XGrpIntro :: GrpIntro -> ChatMsgEvent 'Json -- (GrpIntro is a box type with Int, for possible extension)
+  - or put into XGrpInfo
+    - XGrpInfo :: GroupProfile -> GroupStats -> ChatMsgEvent 'Json -- GroupData?
+    - can update profile between invitation if it happened before joining
+    - can later add logic to "verify" stats?
+    - may be over-complicated until since there "supposed" use cases are out-of-scope / not planned / not known
+
+- What should be default if it's not known? (e.g. admin has older version)
```
epoberezkin, line 27: off is ok, as old version means no PQ support. Also, adding to invitations seems ok, as most groups are either small or large.

docs/rfcs/2024-02-28-pq-integration.md
```diff
@@ -0,0 +1,60 @@
...
+  - On -> then off when member count reaches 20?
+
+### Communicate intent and current state of each conversation
+
+- Current state items
+  - RCEPQEnabled (see #3845) both for direct conversation and per member (regular event items, merged in UI)
+  - created when PQ changes for contact/member (e.g. received from agent on MsgMeta / SENT)
+  - experimental toggle is planned: it doesn't affect contacts/members with already enabled PQ
```
epoberezkin, line 35: maybe enable PQ experimental for contacts only? On another hand we do need to test how it works in groups...

epoberezkin: Could be two separate toggles maybe.

docs/rfcs/2024-02-28-pq-integration.md
```diff
@@ -0,0 +1,60 @@
...
+  - should be well pronounced in UI, not merged
+  - should always say that conversation is e2e encrypted
+  - in direct chats:
+    - reflect actual state of PQ at the time of creation
+    - created during connection handshake when receiving first info about PQ in MsgMeta / some other event (TBC agent api)
+    - will not update if state changes (e.g. upgrades), as toggle is planned to be removed, PQ can't be downgraded, all will support soon
+    - flag in contacts table "e2e_info_created" to only create it once?
+    - should create for legacy contacts or not?
+  - in groups:
+    - reflect intent (should say "PQ will be used for members who support") based on number of members (see above) + toggle
```
epoberezkin, line 49: I think it may reflect the state with admin, so it's not exactly the intent.

docs/rfcs/2024-02-28-pq-integration.md
```diff
@@ -0,0 +1,60 @@
...
+    - created at the same time as feature items? race with history may be possible, but we don't observe it? need to double check or ignore
+    - if based on XGrpInv GroupInvitation (first option above), can create item even before joining
```
epoberezkin, line 51: not a good idea, probably better to wait till admin connects

docs/rfcs/2024-02-28-pq-integration.md
```diff
@@ -0,0 +1,60 @@
...
+    - also will not update (as conversation progresses and it will scroll far up anyway) even if group size changes and it's disabled
+    - flag in groups table "e2e_info_created" to only create it once? and state is only reflected by RCEPQEnabled items?
```
epoberezkin, line 53: why not just link its creation to protocol logistics? not sure why flag is needed.

spaced4ndy: could be used if rich item is to be shown for old contacts / groups (based on first Msg that enables in contact, for example, or after "switching intent on" in groups), in this case it would happen not as part of connection / join

spaced4ndy: though in groups it may be better to create it each time "intent" changes (next line) - TBC

docs/rfcs/2024-02-28-pq-integration.md
spaced4ndy: the admin is only one of many members though, so I think it should say that it's per member depending on their version. At least initially until most clients support it

docs/rfcs/2024-02-28-pq-integration.md
spaced4ndy: if group invitation has size it would already mean that admin is of the version that supports it, so having size < limit can be used to create item. We can also allow to open group conversation while it's "connecting", so it can be seen

spaced4ndy: the downside is that if admin is available and join is fast it will be buried behind history, which is an often case, so it may be worse

PR #3870, opened 2024-03-06
ios: pq support

spaced4ndy:


apps/ios/Shared/Model/SimpleXAPI.swift
```diff
@@ -1818,6 +1831,12 @@ func processReceivedMsg(_ res: ChatResponse) async {
                 }
             }
         }
+    case let .contactPQEnabled(user, contact, _):
+        if active(user) {
+            await MainActor.run {
+                m.updateContact(contact) // or updateContactConnectionStats?
```
spaced4ndy, line 1837: this is not very likely to happen on chat info view, so probably doesn't matter. can even ignore this event probably..

PR #3889, opened 2024-03-10
core: organize withAckMessage

dpwiz:


src/Simplex/Chat.hs
```diff
@@ -3409,7 +3411,7 @@ processAgentMsgRcvFile _corrId aFileId msg =
 
 processAgentMessageConn :: forall m. ChatMonad m => VersionRange -> User -> ACorrId -> ConnId -> ACommand 'Agent 'AEConn -> m ()
 processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage = do
-  entity <- withStore (\db -> getConnectionEntity db vr user $ AgentConnId agentConnId) >>= updateConnStatus
+  entity <- (withStore (\db -> getConnectionEntity db vr user $ AgentConnId agentConnId) >>= updateConnStatus) `catchChatError` \e -> throwError $ ChatErrorAgent (CRITICAL True $ show e) Nothing
```
epoberezkin, line 3414: we only want to catch database Busy/Locked errors here, as we throw exception if entity is absent

src/Simplex/Chat.hs
```diff
@@ -3219,9 +3219,11 @@ processAgentMessage _ connId DEL_CONN =
   toView $ CRAgentConnDeleted (AgentConnId connId)
 processAgentMessage corrId connId msg = do
   vr <- chatVersionRange
-  withStore' (`getUserByAConnId` AgentConnId connId) >>= \case
+  critical (withStore' (`getUserByAConnId` AgentConnId connId)) >>= \case
```
epoberezkin, line 3222: we only want to catch database Busy/Locked errors here

src/Simplex/Chat.hs
```diff
@@ -3482,9 +3496,13 @@ processAgentMsgRcvFile _corrId aFileId msg =
               agentXFTPDeleteRcvFile aFileId fileId
               toView $ CRRcvFileError user ci e ft
 
-processAgentMessageConn :: forall m. ChatMonad m => (PQSupport -> VersionRangeChat) -> User -> ACorrId -> ConnId -> ACommand 'Agent 'AEConn -> m ()
+processAgentMessageConn :: forall m . ChatMonad m => (PQSupport -> VersionRangeChat) -> User -> ACorrId -> ConnId -> ACommand 'Agent 'AEConn -> m ()
 processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage = do
-  entity <- withStore (\db -> getConnectionEntity db vr user $ AgentConnId agentConnId) >>= updateConnStatus
+  -- Missing connection/entity errors here will be sent to the view but not shown as CRITICAL alert,
+  -- as in this case no need to ACK message - we can't process messages for this connection anyway.
```
spaced4ndy, line 3502: also it's a "normal" situation for agent to send connection deletion events DEL_CONN, DEL_RCVQ after connection record has been deleted in chat

src/Simplex/Chat.hs
```diff
@@ -4241,6 +4256,7 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
             _ -> pure ()
         CON _ -> startReceivingFile user fileId
         MSG meta _ msgBody -> do
+          -- XXX: not all branches do ACK
```
spaced4ndy, line 4259: it's legacy files anyway, no longer used - to be deprecated in future version
```suggestion
```

src/Simplex/Chat.hs
```diff
@@ -4384,19 +4400,24 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
     withAckMessage' cId conn msgMeta action = do
-      cmdId <- createAckCmd conn
-      withAckMessage cId cmdId msgMeta $ action $> False
+      withAckMessage cId conn msgMeta False $ \_cmdId -> action $> False
 
-    withAckMessage :: ConnId -> CommandId -> MsgMeta -> m Bool -> m ()
-    withAckMessage cId cmdId msgMeta action = do
+    withAckMessage :: ConnId -> Connection -> MsgMeta -> Bool -> (CommandId -> m Bool) -> m ()
+    withAckMessage cId conn msgMeta showCritical action = do
+      cmdId <- createAckCmd conn `catchChatError` \e -> throwError $ ChatErrorAgent (CRITICAL True $ show e) Nothing
```
spaced4ndy, line 4407: we could reduce load on db by not saving ack cmds

spaced4ndy: and possibly make ACK have special reply instead of OK (ACK_OK?) - to know not to search db for it on agent reply

PR #3891, opened 2024-03-11
multiplatform: e2e information chat items

spaced4ndy:


apps/ios/Shared/Views/Chat/ChatItemView.swift
```diff
@@ -177,7 +177,7 @@ struct ChatItemContentView<Content: View>: View {
 
     private func directE2EEInfoText(_ info: E2EEInfo) -> Text {
         info.pqEnabled
-        ? Text("Messages, files and calls are protected by **quantum resistant e2e encryption**. It has perfect forward secrecy, repudiation and break-in recovery.")
+        ? Text("Messages, files and calls are protected by **quantum resistant e2e encryption** with perfect forward secrecy, repudiation and break-in recovery.")
```
spaced4ndy, line 180: "It" - what? "Encryption has" is strange, "messages etc." doesn't connect

PR #3904, opened 2024-03-13
core: don't save async commands records for ack

spaced4ndy:


src/Simplex/Chat.hs
```diff
@@ -3589,15 +3590,14 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
           -- TODO only acknowledge without saving message?
           -- probably this branch is never executed, so there should be no reason
           -- to save message if contact hasn't been created yet - chat item isn't created anyway
-          withAckMessage agentConnId conn meta False $ \cmdId -> do
-            (_conn', _) <- saveDirectRcvMSG conn meta cmdId msgBody
-            pure False
+          withAckMessage' agentConnId meta $ do
+            (_conn', _) <- saveDirectRcvMSG conn meta msgBody
+            pure ()
```
dpwiz, line 3595: `void` ?

spaced4ndy: yeah.. I just wanted to keep conn' to show it's changed here as in other places. Probably this whole logic can be removed (only ack without saving message - see comment above)

spaced4ndy: changed to void

PR #3905, opened 2024-03-13
core: temp-folder option

spaced4ndy:


src/Simplex/Chat/Options.hs
```diff
@@ -258,6 +259,13 @@ chatOptsP appDir defaultDbFileName = do
             <> metavar "FOLDER"
             <> help "Folder to use for sent and received files"
         )
+  optTempDirectory <-
+    optional $
+      strOption
+        ( long "tmp-dir"
+            <> metavar "TMP_DIR"
```
dpwiz, line 266: metavar should be something like "PATH", describing the expected content: `--tmp-dir PATH`

spaced4ndy: changed to FOLDER as one for files folder

PR #3925, opened 2024-03-19
ui: add missing error types

spaced4ndy:


apps/ios/SimpleXChat/APITypes.swift
```diff
@@ -1878,13 +1878,17 @@ public enum XFTPErrorType: Decodable {
...
     case INTERNAL
 }
 
 public enum RCErrorType: Decodable {
     case `internal`(internalErr: String)
     case identity
     case noLocalAddress
+    case newController
+    case notDiscovered
     case tlsStartFailed
```
spaced4ndy, line 1892: `tLSStartFailed` ?

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/SimpleXAPI.kt
```diff
@@ -5211,6 +5217,8 @@ sealed class RCErrorType {
   @Serializable @SerialName("internal") data class INTERNAL(val internalErr: String): RCErrorType()
   @Serializable @SerialName("identity") object IDENTITY: RCErrorType()
   @Serializable @SerialName("noLocalAddress") object NO_LOCAL_ADDRESS: RCErrorType()
+  @Serializable @SerialName("newController") object NEW_CONTROLLER: RCErrorType()
+  @Serializable @SerialName("notDiscovered") object NOT_DISCOVERED: RCErrorType()
   @Serializable @SerialName("tlsStartFailed") object TLS_START_FAILED: RCErrorType()
```
spaced4ndy, line 5222: `tLSStartFailed` ?

apps/ios/SimpleXChat/APITypes.swift
spaced4ndy: unless it's renamed in backend from RCETLSStartFailed to RCETlsStartFailed (which is more work)

PR #3962, opened 2024-03-27
core: entity locks

spaced4ndy:


tests/ChatTests/Profiles.hs
```diff
@@ -179,10 +179,12 @@ testMultiWordProfileNames =
...
-      alice <## "member #'Our Team' 'Cath Johnson' does not have direct connection, creating"
-      alice <## "contact for member #'Our Team' 'Cath Johnson' is created"
-      alice <## "sent invitation to connect directly to member #'Our Team' 'Cath Johnson'"
-      alice <# "@'Cath Johnson' hello"
+      alice
+        <### [ "member #'Our Team' 'Cath Johnson' does not have direct connection, creating",
+               "contact for member #'Our Team' 'Cath Johnson' is created",
+               "sent invitation to connect directly to member #'Our Team' 'Cath Johnson'",
+               WithTime "@'Cath Johnson' hello"
+             ]
```
spaced4ndy, line 187: this fix is unrelated to locks - in similar tests we also expect in any order

tests/ChatTests/Files.hs
```diff
@@ -787,7 +787,8 @@ testXFTPCancelRcvRepeat =
       bob ##> "/fr 1 ./tests/tmp"
       bob
         <### [ "saving file 1 from alice to ./tests/tmp/testfile_1",
-               "started receiving file 1 (testfile) from alice"
+               "started receiving file 1 (testfile) from alice",
+               StartsWith "chat db error: SERcvFileNotFoundXFTP"
```
spaced4ndy, line 791: this test always had error but a different one - CENoRcvFileUser, it was muted by log level. Error happens because agent reports next downloaded chunk when cancel happened in progress

src/Simplex/Chat/Store/Connections.hs
```diff
@@ -37,6 +39,49 @@ import qualified Simplex.Messaging.Agent.Store.SQLite.DB as DB
 import Simplex.Messaging.Crypto.Ratchet (PQSupport)
 import Simplex.Messaging.Util (eitherToMaybe)
 
+getConnectionEntityId :: DB.Connection -> AgentConnId -> ExceptT StoreError IO ConnectionEntityId
```
spaced4ndy, line 42: couldn't split getConnectionEntity, because its first step is reading connection by id, and it also could be updated so it should be read after taking lock. also it's used in other places as well

src/Simplex/Chat/Protocol.hs
```diff
@@ -112,6 +113,14 @@ agentToChatVersion v
   | v < pqdrSMPAgentVersion = initialChatVersion
   | otherwise = pqEncryptionCompressionVersion
 
+data ConnectionEntityId
```
spaced4ndy, line 116: probably would make more sense to have this together with ConnectionEntity somewhere in Types, but since ConnectionEntity is currently here, didn't move

dpwiz (dismissed):

PR #3980, opened 2024-04-01
rfc: protocol extensions for channels/super-peers

epoberezkin:


docs/rfcs/2024-04-01-super-peers-2.md
```diff
@@ -0,0 +1,262 @@
...
+1. Separate non-participating members that can only be added as a super-peers and do not have their own roles in the group, can't send their own messages or administer the group.
+
+2. Super-peer being a function of any member, irrespective of their role.
+
+While approach 1 can be simpler, it has its downsides:
+- more complex migration of the existing groups - e.g., directory service cannot become a super-peer while remaining group admin (or moderator, if we add a new role).
+- less usable clients - users can have desktop clients with good internet connectivity and sufficient computing resources to effectively host several medium size groups, even with SQLite database.
+- it makes super-peers more like servers, being both unusable as clients and also requiring larger concurrency, and therefore increasing centralization.
+
+Therefore, the approach 2 looks more attractive, when super-peers must have some role in the group, and the super peers that cannot send messages will have observer role.
```
spaced4ndy, line 18: it seems strange to allow non-admins be super-peers, as there supposedly should be higher level of trust to them since they're relaying all messages

docs/rfcs/2024-04-01-super-peers-2.md
```diff
@@ -0,0 +1,262 @@
...
+  prob = d/n ; delivery probability based on target delivery redundancy
+  point = (h + r)/2 ; as member ID and message hash are uniformly random in 0..1 range, `point` will also be uniformly random in 0..1 range
+  start = (1 - prob) * i / (n - 1) ; `start` for the peer with `i == 0` will be `0` and with `i == n - 1` will be `1 - prob`
+  end = start + prob ; `end` for peer 0 will be `prob` and for peer `n - 1` will be 1
+  ; for all peers the range `start..end` will have width `prob`
+  if (point >= start && point < end) deliver
+  else skip;
+```
+
+This algorithm can be proven to result in target average delivery redundancy.
```
spaced4ndy, line 116: > E.g., in groups with 2 super-peers it would be desirable to deliver each message 1.33 (4/3) times on average, while for groups with 3 super peers it can be 2 or 1.667 (5/3) times.

so in practice the messages will have different redundancy? Why not set redundancy to integers, and require super peers to aim for it for each message, seems like it would make more sense than arbitrarily passing some messages without redundancy. Even in same scope of messages some messages can be more important than others, it doesn't seem correct to leave these decisions to chance / rely on requests for re-delivery (?). Even if they're implemented, errors (failed deliveries) are likely in network and it might incur comparable traffic to simply having redundancy set at least to 2 and higher complexity of protocol.

Super-peers could calculate the "index" of a given message, with peers to deliver it determined based on some rule, for example "index" peer and next in circle.

docs/rfcs/2024-04-01-super-peers-2.md
```diff
@@ -0,0 +1,262 @@
...
+- establishing or losing the connection with other members (e.g., when AUTH is received).
+- informing about their decision to stop being super peers in advance.
+- most likely using delivery receipts in communications between super-peers that would include message hashes in RCVD info.
+
+## Authorising administrative changes
+
+As members no longer send messages directly to other members, in addition to the risk of the initial MITM by admin (which is now mitigated by having multiple super-peers) there is a risk of super-peers being compromised at a later stage, and in case of administrative changes (member or group level changes) we have these options:
+
+1. have all members postpone these changes until they are communicated by a sufficient number of super-peers (that would require configuring consensus on the group level).
+2. have admins and owners sign administrative changes with the public key included in their profile during member introduction.
```
spaced4ndy, line 128: which is also passed by super-peers? if so, also adds to the point that being a super-peer requires a certain level of trust

docs/rfcs/2024-04-01-super-peers-2.md
```diff
@@ -0,0 +1,262 @@
...
+
+The downside of approach 2 is the lack of repudiation, in fact, there is a non-repudiation quality of such administrative changes as role changes, member additions and deletions, and group changes.
+
+Overall, while repudiation of sent messages appears as important, it seems much less important for administrative changes, and signing member and group changes while relying on Merkle DAG for message history integrity appears to be an optimal tradeoff.
+
+To support this functionality the `admin` and `owner` members have to add public keys to their profiles and communicate profile updates to all members (before or after group is switched to super-peers, but probably before is better).
```
spaced4ndy, line 136: profile update has also to be communicated to new members when group is switched, does this imply at that point super-peers would be passing these updates, or owners and admins will stay connected to all? (I guess first?)

docs/rfcs/2024-04-01-super-peers-2.md
```diff
@@ -0,0 +1,262 @@
...
+
+The middle ground here is moderation, and trade-off here is more nuanced. Practically, whether message is fully removed or marked as removed is the group policy decision, and it can be made based on whether it's more important to preserve content or to remove undesirable content without trace, and also given that super-peers are likely to be give automatic moderation capabilities anyway, preserving deniability for moderation events and relying on Merkle DAG to identify integrity violation seems a better alternative than signing moderation messages. Although it can also be a part of group policy whether to require signing moderation events.
```
spaced4ndy, line 138: we could have a separate preference/policy for moderation - "Moderation full deletion". It seems justified to want to have moderated messages fully removed, while at the same time keeping "mark for deletion" functionality for deletions of own messages.

docs/rfcs/2024-04-01-super-peers-2.md
```diff
@@ -0,0 +1,262 @@
...
+
+1. have all members postpone these changes until they are communicated by a sufficient number of super-peers (that would require configuring consensus on the group level).
+2. have admins and owners sign administrative changes with the public key included in their profile during member introduction.
+
+The downside of approach 1 is that administrative actions will be delayed and have to be executed not at the time the message is delivered, but at the time message is confirmed by other super-peers. That is separate and in addition to member consensus for some changes (see below). That also means that groups with one super-peer will have no defence mechanism against super-peer being compromised. That also means that groups with two super-peers will either also have no such defence (in case required consensus level is 1) or will require that both super-peers are available, and no administrative actions will be executed unless both super-peers are available.
```
spaced4ndy, line 130: > and no administrative actions will be executed unless both super-peers are available.

maybe it's not a downside and would complicate compromising administrative actions without having to implement signatures? (rfc already introduces some subprotocol ideas for that, and it's unclear how owner/admin profile updates are passed). Maybe minimum number of super-peers should be 3 (third added after joining link), with 2 required for consensus?

docs/rfcs/2024-04-01-super-peers-2.md
```diff
@@ -0,0 +1,262 @@
...
+
+## Role, rank and moderation permission changes of members and group profile changes
+
+There are two problems with the current approach to changing permission, when a single member makes this decision:
+- member whose role changes may disagree. E.g., the member may not be willing to have owner or admin role in some groups. Even more so, the member may be unable to perform super-peer functions (have rank 1).
+- other owners or admins may disagree with the change or it may have been made by mistake.
+
+With the move to super-peers it is additionally complicated by the fact that super-peers will be forwarding these messages, and they could be compromised, thus disrupting group functioning - this risk currently exists with the existing group directory that plays admin role and in case it is compromised it can remove all members from the group.
+
+Part of this problem is addressed in the previous point by requiring to sign administrative changes.
```
spaced4ndy, line 172: It seems futile to hope for reliability in group where super-peers are compromised, especially with 2 super-peers and unclear way to reach consensus. I think rfc leaned too early into idea of any members being able to be super-peers. Perhaps it is not required for regular clients to be able to be super peers (or at least via the same member record that can compose messages). Instead owners could add an "invisible" member record to group to act as super-peer (in case they use a highly available client, which usually isn't a case even if it's desktop app, since people turn it off/put to sleep, so it doesn't seem that valuable for most groups), or request "hosted" super-peers (similar to simplex-directory) up to certain quota (min 3 max 5? or self-hosted to differentiate from chat client), or just default to 3 for MVP.

docs/rfcs/2024-04-01-super-peers-2.md
```diff
@@ -0,0 +1,262 @@
...
+The messages send from multiple owners can have the benefit of avoiding targeted attacks on the member who originated it - it will only be known to other members sending the message, but not to other members, creating mutual responsibility for some changes and/or announcements.
+
+## Choosing MVP scope
+
+With all these ideas for improvements, the challenge is to choose some valuable MVP shippable in the shortest time.
+
+From the UX point of view, the most lacking parts are:
+
+- broadcasting messages via super-peers, including:
+  - protocol to accept and approve role changes is necessary, as it should not be possible to unilaterally appoint some member to be a super-peer.
```
spaced4ndy, line 250: I don't think accepting role change is MVP

docs/rfcs/2024-04-01-super-peers-2.md
epoberezkin: Yes, but it protects from future compromise. Plus super-peer redundancy validates the keys - they can be controlled by different parties.

PR #4009, opened 2024-04-10
core: inactive connections/members

spaced4ndy:


src/Simplex/Chat.hs
```diff
@@ -4331,11 +4336,17 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
       OK ->
         -- [async agent commands] continuation on receiving OK
         when (corrId /= "") $ withCompletedCommand conn agentMsg $ \_cmdData -> pure ()
+      QCONT -> do
+        processConnQCONT connEntity conn
+        lastSharedMsgId <- withStore' $ \db -> getLastItemSharedMsgId db m
+        forM_ lastSharedMsgId $ \lastMsgId -> do
+          let conn' = conn {inactive = False} :: Connection
+          void $ sendDirectMemberMessage conn' (XGrpMsgSkipped lastMsgId) groupId
```
spaced4ndy, line 4344: `processConnQCONT` could return updated connection `conn'`

src/Simplex/Chat.hs
```diff
@@ -4331,11 +4336,17 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
       OK ->
         -- [async agent commands] continuation on receiving OK
         when (corrId /= "") $ withCompletedCommand conn agentMsg $ \_cmdData -> pure ()
+      QCONT -> do
+        processConnQCONT connEntity conn
+        lastSharedMsgId <- withStore' $ \db -> getLastItemSharedMsgId db m
```
spaced4ndy, line 4341: - for super peer it may be better to instead search via `msg_deliveries` table, to account for forwarded group messages (see `forwardMsg_`)
  - may have to additionally store original shared msg id (currently shared_msg_id is for the forwarded message itself)
- or populate `group_snd_item_statuses`?
- store additional data

src/Simplex/Chat/Messages/CIContent.hs
```diff
@@ -311,6 +312,7 @@ rcvGroupEventToText = \case
   RGEInvitedViaGroupLink -> "invited via your group link"
   RGEMemberCreatedContact -> "started direct connection with you"
   RGEMemberProfileUpdated {} -> "updated profile"
+  RGESkippedMessages -> "notified about skipped messages"
```
spaced4ndy, line 315: draft - probably should have different text, e.g. "continued message delivery [after your inactivity]". and/or more info?

spaced4ndy: maybe no item is required

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -2581,3 +2582,22 @@ getGroupHistoryItems db user@User {userId} GroupInfo {groupId} count = do
...
+    DB.query
+      db
+      [sql|
+        SELECT shared_msg_id
+        FROM chat_items
+        WHERE chat_item_id = (
+          SELECT chat_item_id
+          FROM group_snd_item_statuses
+          WHERE group_member_id = ?
+            AND (group_snd_item_status LIKE 'snd_sent%' OR group_snd_item_status LIKE 'snd_rcvd%')
```
spaced4ndy, line 2598: could store "tags" to avoid LIKE

src/Simplex/Chat.hs
```diff
@@ -5917,6 +5942,11 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
             XGrpInfo p' -> xGrpInfo gInfo author p' rcvMsg msgTs
             _ -> messageError $ "x.grp.msg.forward: unsupported forwarded event " <> T.pack (show $ toCMEventTag event)
 
+    xGrpMsgSkipped :: GroupInfo -> GroupMember -> SharedMsgId -> RcvMessage -> UTCTime -> CM ()
+    xGrpMsgSkipped _gInfo _m _sharedMsgId _msg _brokerTs = do
+      -- TODO can request delivery starting with _sharedMsgId here
+      pure ()
```
spaced4ndy, line 5948: originally had a new type of group event chat item created here, but removed it as there's already integrity error item created so it seems unnecessary. Maybe should add it back, with it saying "member continued delivery for you"

PR #4024, opened 2024-04-12
ios: forward ui

spaced4ndy:


apps/ios/Shared/Views/Chat/ChatItemInfoView.swift
```diff
@@ -240,6 +252,37 @@ struct ChatItemInfoView: View {
         }
     }
 
+    @ViewBuilder private func forwardedFromTab(_ forwardedFromItem: AChatItem) -> some View {
```
spaced4ndy, line 255: Rendering full item here via ChatItemView is problematic due to various item interactions. For example, voice message renders as playable but can't be played (probably requires adding player to view).

Accounting for all possible interactions seems pointless as the item copy is readily available - the item for which info was opened.

apps/ios/Shared/Views/Chat/ChatItem/FramedItemView.swift
```diff
@@ -65,8 +66,13 @@ struct FramedItemView: View {
...
                             }
                         }
-                } else if chatItem.meta.itemForwarded != nil {
-                    framedItemHeader(icon: "arrowshape.turn.up.forward", caption: Text("forwarded").italic())
+                } else if let itemForwarded = chatItem.meta.itemForwarded {
+                    if chat.chatInfo.chatType == .local,
+                       let fromChatName = itemForwarded.chatName {
+                        framedItemHeader(icon: "arrowshape.turn.up.forward", caption: Text("saved from \(fromChatName)").italic(), padBottom: true)
+                    } else {
+                        framedItemHeader(icon: "arrowshape.turn.up.forward", caption: Text("forwarded").italic(), padBottom: true)
```
epoberezkin, line 74: the condition can be just on text:

chat.chatInfo.chatType == .local ? "saved from \(fromChatName)" : "forwarded"

epoberezkin: and if it's local and there is no name it should be "saved" - probably better to get it to a function and avoid "if" here.

apps/ios/Shared/Views/Chat/ChatItem/FramedItemView.swift
```diff
@@ -40,19 +40,20 @@ struct FramedItemView: View {
     var body: some View {
         let v = ZStack(alignment: .bottomTrailing) {
             VStack(alignment: .leading, spacing: 0) {
+                let headerPadBottom = chatItem.quotedItem == nil && chatItem.meta.itemForwarded == nil
```
epoberezkin, line 43: this is a bit of a hack, you are mixing two concerns and have parallel ifs here - better make another function for cases where padding is needed (in addition to framedItemHeader)

epoberezkin: although maybe I don't get the problem...

apps/ios/Shared/Views/Chat/ChatItem/FramedItemView.swift
```diff
@@ -180,7 +186,7 @@ struct FramedItemView: View {
         .foregroundColor(.secondary)
         .padding(.horizontal, 12)
         .padding(.top, 6)
-        .padding(.bottom, chatItem.quotedItem == nil ? 6 : 0) // TODO think how to regroup
+        .padding(.bottom, padBottom ? 6 : 0)
```
epoberezkin, line 189: ah, so it was simply determined inside... why pass from outside?

apps/ios/Shared/Views/Chat/ChatItemForwardingView.swift
```diff
@@ -0,0 +1,145 @@
...
+    @ViewBuilder private func forwardListNavLinkView(_ chat: Chat) -> some View {
+        Button {
+            if chat.id != fromChatInfo.id {
+                Task {
+                    await MainActor.run {
+                        chatModel.forwardToChatId = chat.id
+                        chatModel.forward = ComposeState.init(forwardingItem: ci, fromChatInfo: fromChatInfo)
+                        dismiss()
+                    }
+                    DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
```
epoberezkin, line 104: the problem can be that it is executed after Task termination - why it is asyncAfter and not Task.delay with await MainActor.run?

apps/ios/Shared/Views/Chat/ChatItem/FramedItemView.swift
spaced4ndy: to avoid extra padding between marked deleted and forwarded headers

apps/ios/Shared/Views/Chat/ChatItem/FramedItemView.swift
spaced4ndy: first branch also requires checking fromChatName is not null

apps/ios/Shared/Views/Chat/ChatItem/FramedItemView.swift
epoberezkin: why not amend the condition locally?

spaced4ndy: it would affect forwarded items themselves them - same view is used for them

apps/ios/Shared/Views/Chat/ChatItem/FramedItemView.swift
spaced4ndy: refactored into function

apps/ios/Shared/Views/Chat/ChatItemInfoView.swift
```diff
@@ -275,6 +287,47 @@ struct ChatItemInfoView: View {
         : Color(uiColor: .tertiarySystemGroupedBackground)
     }
 
+    @ViewBuilder private func forwardedFromTab(_ forwardedFromItem: AChatItem) -> some View {
```
spaced4ndy, line 290: Rendering full item here via ChatItemView is problematic due to various item interactions. For example, voice message renders as playable but can't be played (probably requires adding player to view).

Accounting for all possible interactions seems pointless as the item copy is readily available - the item for which info was opened.

apps/ios/Shared/Views/Chat/ChatItemInfoView.swift
```diff
@@ -275,6 +295,75 @@ struct ChatItemInfoView: View {
...
+            .padding()
+        }
+        .frame(maxHeight: .infinity, alignment: .top)
+    }
+
+    private func forwardedFromView(_ forwardedFromItem: AChatItem) -> some View {
+        VStack(alignment: .leading, spacing: 8) {
+            Button {
+                Task {
+                    await MainActor.run {
```
avently, line 317: Don't think it's needed since clicks are processed in main thread anyway

PR #4026, opened 2024-04-13
docs: Fix missing images for FR, CZ and PL translation

Arturro43:
Translated docs were calling for non existent directory. Fixed by copying "images" directory to each of the translations directory. During testing, markdown wasn't able to point to correct (main "images") directory. This fix is not perfect, but at least it works...


Arturro43:
Found better way, closing.

PR #4036, opened 2024-04-16
docs: IP address protection rfc

spaced4ndy:


docs/rfcs/2024-04-16-ip-address-protection.md
```diff
@@ -0,0 +1,81 @@
...
+  3. Some file chunks are on unknown servers, tor is enabled
+    - Do not show alert, accept automatically (images, voice messages) or without approval (other files)
+    - `onlyViaTor` is True
+- On file download:
+  - `onlyViaTor` equal False is ignored
+  - If `onlyViaTor` is True, and tor is not enabled, throw new "permanent" error
+    - Permanent for simplicity as this is an edge case
+    - File record to be removed from agent
+    - Could analyze whether it's remaining (not yet downloaded) chunks are on unknown servers, and only abort in this case (and instead proceed if remaining chunks are on known servers)
+      - This requires changing download to load data for all chunks instead of loading only current chunk
```
epoberezkin, line 71: ?

spaced4ndy: xftp download worker works with a single chunk - we could additionally load other chunks to check whether remaining chunks are on unknown servers

spaced4ndy: actually similar is already done for checking completeness, but after download, this would be checked before download

PR #4039, opened 2024-04-16
android, desktop: forward ui

avently:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatItemInfoView.kt
```diff
@@ -222,6 +289,22 @@ fun ChatItemInfoView(chatModel: ChatModel, ci: ChatItem, ciInfo: ChatItemInfo, d
     }
   }
 
+  @Composable
+  fun ForwardedFromTab(forwardedFromItem: AChatItem) {
+    // LALAL SCROLLBAR DOESN'T WORK
```
spaced4ndy, line 294: ? :)

avently: It's something that doesn't work yet and no time to look at it closer. Every tab on that screen has invisible scrollbar inside, so it's something to look at later

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ComposeView.kt
```diff
@@ -563,7 +590,12 @@ fun ComposeView(
         sent = send(chat, MsgContent.MCText(msgText), quotedItemId, null, live, ttl)
       }
     }
+    val wasForwarding = cs.forwarding
     clearState(live)
+    val draft = chatModel.draft.value
+    if (wasForwarding && chatModel.draftChatId.value == chat.chatInfo.id && draft != null) {
+      composeState.value = draft
+    }
```
spaced4ndy, line 598: a couple of interactions seems incorrect:
- when forwarding to a chat where draft is available, draft is offered as "forward comment" (already in compose), but if it had an attachment it won't be sent. instead draft shouldn't be included initially, but after forward is sent compose is populated with draft
- second is not related to draft - when compose is not empty, forward in the same chat overwrites it, making text empty and removing preview. instead text and link preview should be kept (other previews should be removed)

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatItemInfoView.kt
```diff
@@ -151,6 +153,71 @@ fun ChatItemInfoView(chatModel: ChatModel, ci: ChatItem, ciInfo: ChatItemInfo, d
...
+      ) {
+        Column {
+          if (forwardedFromItem.chatItem.chatDir.sent) {
+            ChatPreviewTitleText(text = stringResource(MR.strings.sender_you_pronoun), fontStyle = FontStyle.Italic)
+            Spacer(Modifier.height(DEFAULT_PADDING_HALF))
+            Text(forwardedFromItem.chatInfo.chatViewName, maxLines = 1, style = MaterialTheme.typography.body1.copy(color = if (isInDarkTheme()) MessagePreviewDark else MessagePreviewLight, lineHeight = 22.sp))
+          } else if (forwardedFromItem.chatItem.chatDir is CIDirection.GroupRcv) {
+            ChatPreviewTitleText(text = forwardedFromItem.chatItem.chatDir.groupMember.chatViewName)
+            Spacer(Modifier.height(DEFAULT_PADDING_HALF))
+            Text(forwardedFromItem.chatInfo.chatViewName, maxLines = 1, style = MaterialTheme.typography.body1.copy(color = if (isInDarkTheme()) MessagePreviewDark else MessagePreviewLight, lineHeight = 22.sp))
```
spaced4ndy, line 190: here also secondary

PR #4043, opened 2024-04-18
core: check known relays before file reception, support user approval of unknown relays

spaced4ndy:


src/Simplex/Chat/Controller.hs
```diff
@@ -1110,6 +1110,7 @@ data ChatErrorType
   | CEFileImageType {filePath :: FilePath}
   | CEFileImageSize {filePath :: FilePath}
   | CEFileNotReceived {fileId :: FileTransferId}
+  | CEFileAbortedNotApproved {fileId :: FileTransferId, unknownServers :: [XFTPServer]}
```
epoberezkin, line 1113:
```suggestion
  | CEFileNotApproved {fileId :: FileTransferId, unknownServers :: [XFTPServer]}
```

src/Simplex/Chat/Messages.hs
```diff
@@ -529,6 +529,7 @@ data CIFileStatus (d :: MsgDirection) where
   CIFSRcvInvitation :: CIFileStatus 'MDRcv
   CIFSRcvAccepted :: CIFileStatus 'MDRcv
   CIFSRcvTransfer :: {rcvProgress :: Int64, rcvTotal :: Int64} -> CIFileStatus 'MDRcv
+  CIFSRcvAborted :: CIFileStatus 'MDRcv
```
epoberezkin, line 532:
```suggestion
  CIFSRcvNotApproved :: CIFileStatus 'MDRcv
```

epoberezkin: Aborted is unclear here, better to use the same term consistently

src/Simplex/Chat/Controller.hs
epoberezkin: Aborted is unnecessary here

src/Simplex/Chat/Messages.hs
spaced4ndy: this does happen when it's aborted though, could be reused for other errors possibly

PR #4051, opened 2024-04-19
ios: remove experimental PQ toggle

spaced4ndy:


apps/ios/Shared/Views/Chat/ChatInfoView.swift
```diff
@@ -168,19 +165,11 @@ struct ChatInfoView: View {
                 }
                 .disabled(!contact.ready || !contact.active)
 
-                if pqExperimentalEnabled,
-                   let conn = contact.activeConn {
+                if let conn = contact.activeConn {
                     Section {
                         infoRow(Text(String("E2E encryption")), conn.connPQEnabled ? "Quantum resistant" : "Standard")
```
spaced4ndy, line 170: do we want to keep it?

PR #4060, opened 2024-04-22
android: remove experimental PQ toggle

spaced4ndy:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatInfoView.kt
```diff
@@ -360,13 +345,9 @@ fun ChatInfoLayout(
     }
 
     val conn = contact.activeConn
-    if (pqExperimentalEnabled && conn != null) {
-      SectionView("Quantum resistant E2E encryption") {
+    if (conn != null) {
+      SectionView {
         InfoRow("E2E encryption", if (conn.connPQEnabled) "Quantum resistant" else "Standard")
```
spaced4ndy, line 350: keeping same as in iOS

PR #4075, opened 2024-04-23
ui: translations

epoberezkin:


summoner001:
I restored the messages that were lost in the incident in weblate.

The gettext po file backup worked fine, except for the 7 texts that were added yesterday, but I transferred them from android translation.

Ready to use again.

summoner001:
I finished the review.

But for some reason it doesn't send in that I'm done with review.

Anyway.

Yes, these should be on the weblate. I think it's good.

PR #4077, opened 2024-04-24
 notification led for new message

### LINKED ISSUE #4029: Notification LED not light up and blink

### Platform

Android

### OS version

Android 9

### App version

5.6.1 (193)

### Current Behavior

When the phone display is off and I receive a message, I hear the notification tone, but the notification LED on the phone does not blinking.

It would be nice to be able to adjust the colour within the application...

### Expected Behavior

When I get a message with the display off, don't just play a sound, use the notification LED on the phone. 

### Steps To Reproduce

I don't know if a device specific problem (my device is a Cubot Quest) but if you have on the phone a notification LED see it when you get a message the LED not blinking.

### Relevant log output

_No response_

psuzn:
Fixes #4029

### DIFF
--- a/apps/multiplatform/android/src/main/java/chat/simplex/app/model/NtfManager.android.kt
+++ b/apps/multiplatform/android/src/main/java/chat/simplex/app/model/NtfManager.android.kt
@@ -106,6 +106,7 @@ object NtfManager {
       .setSmallIcon(R.drawable.ntf_icon)
       .setLargeIcon(largeIcon)
       .setColor(0x88FFFF)
+      .setLights(0x88FFFF,500,2000)
       .setAutoCancel(true)
       .setVibrate(if (actions.isEmpty()) null else longArrayOf(0, 250, 250, 250))
       .setContentIntent(chatPendingIntent(OpenChatAction, user.userId, chatId))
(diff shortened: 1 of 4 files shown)


psuzn:

epoberezkin:
Thank you!

It is cool, but it will raise objections from some users as something that can compromise privacy - so there should be an option in Notification settings, something like "Show light" (and we can set default to "on" for the new users and "off" for the existing users).

@avently

summoner001:
@epoberezkin 

Couldn't it be possible to set the LED color per profile like in k9 email?

summoner001:
@epoberezkin 

Couldn't it be possible to set the LED color per profile like in k9 email?

epoberezkin:
we don't set most privacy settings per profile, so no need to complicate it - the MVP to release it would be a global in-app setting that would apply to all profiles, because then it won't require any changes in the core - it just requires updating native preferences and adding a simple toggle in the UI.

psuzn:
| Screenshot 2024-04-29 at 19 13 03 | Screenshot 2024-04-29 at 19 25 57|
|-----------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------|

epoberezkin:
Great!

I think we should change it to toggle and ship it. The UX convention is to use toggle for options with on/off states.

epoberezkin:
see comment in #4229

PR #4110, opened 2024-04-30
core: wallpapers api

epoberezkin:


src/Simplex/Chat.hs
```diff
@@ -7086,6 +7122,8 @@ chatCommandP =
       "/_set alias @" *> (APISetContactAlias <$> A.decimal <*> (A.space *> textP <|> pure "")),
       "/_set alias :" *> (APISetConnectionAlias <$> A.decimal <*> (A.space *> textP <|> pure "")),
       "/_set prefs @" *> (APISetContactPrefs <$> A.decimal <* A.space <*> jsonP),
+      "/_set theme user " *> (APISetUserUITheme <$> A.decimal <*> optional (A.space *> jsonP)),
+      "/_set theme " *> (APISetChatUITheme <$> chatRefP <*> optional (A.space *> jsonP)),
```
dpwiz, line 7126: `_set chat theme` / `_set user theme` for order-independent parsing and natural reading?

src/Simplex/Chat/Types/UITheme.hs
```diff
@@ -0,0 +1,124 @@
...
+  { base :: ColorScheme,
+    wallpaper :: Maybe ChatWallpaper,
+    colors :: UIColors
+  }
+  deriving (Eq, Show)
+
+defaultUITheme :: UITheme
+defaultUITheme = UITheme CSLight Nothing defaultUIColors
+
+data ColorScheme = CSLight | CSDark | CSSimplex
```
avently, line 31: There is another option here which is `System`. This means that if user chose System theme, it can be light/dark based on scheme enabled system-wide.
And there is another option that chooses between two dark themes (Dark vs SimpleX):

https://github.com/simplex-chat/simplex-chat/blob/stable/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/SimpleXAPI.kt#L168

This one probably is not that important

src/Simplex/Chat/AppSettings.hs
```diff
@@ -69,7 +72,9 @@ defaultAppSettings =
       confirmDBUpgrades = Just False,
       androidCallOnLockScreen = Just LSCShow,
       iosCallKitEnabled = Just True,
-      iosCallKitCallsInRecents = Just False
+      iosCallKitCallsInRecents = Just False,
+      uiColorScheme = Just CSLight,
```
avently, line 76: Default is System, not Light

src/Simplex/Chat/Types/UITheme.hs
```diff
@@ -0,0 +1,124 @@
...
+
+data UITheme = UITheme
+  { base :: ColorScheme,
+    wallpaper :: Maybe ChatWallpaper,
+    colors :: UIColors
+  }
+  deriving (Eq, Show)
+
+defaultUITheme :: UITheme
+defaultUITheme = UITheme CSLight Nothing defaultUIColors
```
avently, line 29: Same, default it System

src/Simplex/Chat/AppSettings.hs
```diff
@@ -43,7 +44,9 @@ data AppSettings = AppSettings
     confirmDBUpgrades :: Maybe Bool,
     androidCallOnLockScreen :: Maybe LockScreenCalls,
     iosCallKitEnabled :: Maybe Bool,
-    iosCallKitCallsInRecents :: Maybe Bool
+    iosCallKitCallsInRecents :: Maybe Bool,
+    uiColorScheme :: Maybe ColorScheme,
+    uiThemes :: Maybe [UITheme]
```
avently, line 49: Maybe you can include profile image corner radius in the app settings too since it's supported on every platform

PR #4111, opened 2024-04-30
core: api to differentiate contacts and conversations

spaced4ndy:


src/Simplex/Chat.hs
```diff
@@ -1072,6 +1083,14 @@ processChatCommand' vr = \case
               pure (e_, map aConnId conns)
     CTLocal -> pure $ chatCmdError (Just user) "not supported"
     CTContactRequest -> pure $ chatCmdError (Just user) "not supported"
+  APIMarkConversationDeleted cRef@(ChatRef cType _chatId) deleted -> withUser $ \user -> case cType of
```
spaced4ndy, line 1086: when deleting conversation, if contact is not kept, UI would call APIDeleteChat instead

src/Simplex/Chat/Controller.hs
```diff
@@ -293,7 +293,8 @@ data ChatCommand
   | UserRead
   | APIChatRead ChatRef (Maybe (ChatItemId, ChatItemId))
   | APIChatUnread ChatRef Bool
-  | APIDeleteChat ChatRef Bool -- `notify` flag is only applied to direct chats
+  | APIDeleteChat ChatRef Bool Bool -- `keep_conversation` and `notify` flags are only applied to direct chats
```
epoberezkin, line 296:
```suggestion
  | APIDeleteChat {contactId :: ContactId, notify :: Bool, keepChat :: Bool}
 -- `keep_conversation` and `notify` flags are only applied to direct chats
```

maybe better split...

src/Simplex/Chat/Controller.hs
```diff
@@ -293,7 +293,8 @@ data ChatCommand
...
+  | APIMarkConversationDeleted ChatRef Bool
```
epoberezkin, line 297:
```suggestion
  | APIMarkChatDeleted ChatRef Bool
```

src/Simplex/Chat/Controller.hs
```diff
@@ -615,6 +616,7 @@ data ChatResponse
   | CRGroupMemberUpdated {user :: User, groupInfo :: GroupInfo, fromMember :: GroupMember, toMember :: GroupMember}
   | CRContactsMerged {user :: User, intoContact :: Contact, mergedContact :: Contact, updatedContact :: Contact}
   | CRContactDeleted {user :: User, contact :: Contact}
+  | CRContactConversationDeleted {user :: User, contact :: Contact, deleted :: Bool}
```
epoberezkin, line 619:
```suggestion
  | CRContactChatDeleted {user :: User, contact :: Contact, deleted :: Bool}
```

src/Simplex/Chat/Migrations/M20240430_conversation_deleted.hs
```diff
@@ -0,0 +1,18 @@
...
+
+module Simplex.Chat.Migrations.M20240430_conversation_deleted where
+
+import Database.SQLite.Simple (Query)
+import Database.SQLite.Simple.QQ (sql)
+
+m20240430_conversation_deleted :: Query
+m20240430_conversation_deleted =
+  [sql|
+ALTER TABLE contacts ADD COLUMN conversation_deleted INTEGER NOT NULL DEFAULT 0;
```
epoberezkin, line 11:
```suggestion
ALTER TABLE contacts ADD COLUMN chat_deleted INTEGER NOT NULL DEFAULT 0;
```

src/Simplex/Chat/Migrations/M20240430_conversation_deleted.hs
```diff
@@ -0,0 +1,18 @@
+{-# LANGUAGE QuasiQuotes #-}
+
+module Simplex.Chat.Migrations.M20240430_conversation_deleted where
+
+import Database.SQLite.Simple (Query)
+import Database.SQLite.Simple.QQ (sql)
+
+m20240430_conversation_deleted :: Query
```
epoberezkin, line 8:
```suggestion
m20240430_chat_deleted :: Query
```

I don't think we need a new term here

src/Simplex/Chat.hs
```diff
@@ -1009,22 +1009,38 @@ processChatCommand' vr = \case
...
+            cancelFilesInProgress user filesInfo
+            sendDelDeleteConns ct notify
+            -- TODO db: delete connection
+            -- TODO db: updateContactStatus to CSDeleted
+            pure $ CRContactDeleted user ct
+          CDMMessages -> do
+            void $ processChatCommand $ APIClearChat cRef
+            -- TODO ct <- db: set chat_deleted to deleted
+            -- (can be unset on any new chat item (see updateChatTs), and on APIMarkConversationDeleted False)
+            pure $ CRContactDeleted user ct
```
spaced4ndy, line 1037: UI to differentiate based on call params - update in model instead of deleting

src/Simplex/Chat.hs
```diff
@@ -1009,22 +1009,38 @@ processChatCommand' vr = \case
...
+            cancelFilesInProgress user filesInfo
+            sendDelDeleteConns ct notify
+            -- TODO db: delete connection
+            -- TODO db: updateContactStatus to CSDeleted
+            pure $ CRContactDeleted user ct
+          CDMMessages -> do
+            void $ processChatCommand $ APIClearChat cRef
+            -- TODO ct <- db: set chat_deleted to deleted
+            -- (can be unset on any new chat item, see updateChatTs)
+            pure $ CRContactDeleted user ct
```
spaced4ndy, line 1037: for CDMEntity and CDMMessages UI to differentiate based on call params - update in model instead of deleting

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -315,7 +315,7 @@ updateChatTs db User {userId} chatDirection chatTs = case toChatInfo chatDirecti
   DirectChat Contact {contactId} ->
     DB.execute
       db
-      "UPDATE contacts SET chat_ts = ? WHERE user_id = ? AND contact_id = ?"
+      "UPDATE contacts SET chat_ts = ?, chat_deleted = 0 WHERE user_id = ? AND contact_id = ?"
```
spaced4ndy, line 318: on new chat items UI would "manually" update contacts with chatDeleted set to False. May be a bit hacky but it seems a normal assumption to do from UI view point and less hassle than correctly propagate contact update from db operations to all responses in backend. This is somewhat similar in logic to chat "popping" to top on new chat items being only UI logic

PR #4184, opened 2024-05-15
docs: fix example systemd service unit in xftp-server documentation page

sergeyshadok:
Systemd itself uses %<something> tokens so %% needs to be used for specifiers

(https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html)


sergeyshadok:

PR #4224, opened 2024-05-25
core: optimize subscribeUser

dpwiz:


src/Simplex/Chat.hs
```diff
@@ -3364,173 +3364,135 @@ agentSubscriber = do
...
-  -- TODO possibly, we could either disable these events or replace with less noisy for API
-  contactLinkSubsToView rs ucs
-  groupSubsToView rs gs ms ce
-  sndFileSubsToView rs sfts
-  rcvFileSubsToView rs rfts
-  pendingConnSubsToView rs pcs
+  env <- ask
+  (conns, ctConns, ucs, gs, mConns, sfts, rfts, pcConns) <-
+    withStore' $ \db -> flip runReaderT env $ do
+      -- run in one transaction so onlyNeeded will be atomic
```
epoberezkin, line 3373: it is unnecessary complication, as once the app goes into foreground NSE will no longer be executed, so onlyNeeded cannot change.

Even if it were not the case, making it atomic would not solve the problem, as if NSE could run _after_ the app re-opened, we would have a bigger issue - the new connections would not resubscribe.

So combining them into one transaction is completely unnecessary.

CherrieTheShifter:
Yes this. Optimize and fix .ore of the app

PR #4229, opened 2024-05-26
android: notification led

epoberezkin:


avently:
Looks like in order to make this feature working on API >= 26 (our min suppported API is 26) we need to enable lights on notification channel and after that it can't be turned off. So, we can't enable it now without recreating all notification channels that needs to show notificaiton with lights. But the good news for those who need this feature they can just enable it themeselves on choosen notification channel by going to app's notification settings and enabling lights manually, for example, on message notification channel. It will work without any code modification. So I think adding a button to open settings with a meaningful footer text is enough. Other changes aren't needed

avently:
@psuzn can you confirm that enabling lights in notification settings without having code modifications from this PR works for you for turning lights on?

PR #4245, opened 2024-05-28
android: show alert of exception while initializing WebView

avently:


apps/multiplatform/common/src/commonMain/resources/MR/base/strings.xml
```diff
@@ -874,6 +874,7 @@
   <string name="audio_device_speaker">Speaker</string>
   <string name="audio_device_wired_headphones">Headphones</string>
   <string name="audio_device_bluetooth">Bluetooth</string>
+  <string name="error_initializing_web_view">Error while initializing WebView. Make sure it\'s the newest version. Contact developers for assistance: \n%s</string>
```
epoberezkin, line 877:
```suggestion
  <string name="error_initializing_web_view">Error initializing WebView. Update your system to the new version. Please contact developers.\nError: %s</string>
```

PR #4256, opened 2024-05-30
cli: options for private message routing and fallback mode

epoberezkin:


src/Simplex/Chat/Options.hs
```diff
@@ -123,18 +129,36 @@ coreChatOptsP appDir defaultDbFileName = do
...
         ( long "socks-proxy"
             <> metavar "SOCKS5"
             <> help "Use SOCKS5 proxy at `ipv4:port` or `:port`"
             <> value Nothing
         )
+  smpProxyMode_ <-
+    optional $
+      option
+        textJsonParse
+        ( short 'y'
```
spaced4ndy, line 142: Maybe remove short options? This is usually written once then repeated. y is often used as "automatically answer yes"

PR #4294, opened 2024-06-05
android, desktop: handle situation when not all databases created before shut down

avently:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/helpers/DatabaseUtils.kt
```diff
@@ -39,22 +39,28 @@ object DatabaseUtils {
...
 
-  private fun hasDatabase(rootDir: String): Boolean =
-    File(rootDir + File.separator + chatDatabaseFileName).exists() && File(rootDir + File.separator + agentDatabaseFileName).exists()
+  private fun hasAtLeastOneDatabase(rootDir: String): Boolean =
+    File(rootDir + File.separator + chatDatabaseFileName).exists() || File(rootDir + File.separator + agentDatabaseFileName).exists()
+
+  fun hasOnlyOneDatabase(rootDir: String): Boolean {
+    val chat = File(rootDir + File.separator + chatDatabaseFileName).exists()
+    val agent = File(rootDir + File.separator + agentDatabaseFileName).exists()
+    return (chat || agent) && (!chat || !agent)
```
epoberezkin, line 48:
```suggestion
    return !(chat && agent) && (chat || agent)
```

PR #4304, opened 2024-06-07
ios: wallpapers

avently:


Yourgirl16:
Please Stanislav Dmitrenko how can you so evil and told me you already fvcking another girl in Vietnam while you made me miserable abused me and left me in pieces while all i do is care for you

apps/ios/SimpleXChat/AppGroup.swift
```diff
@@ -276,6 +276,68 @@ public class IntDefault: Default<Int> {
     }
 }
 
+public class StringDefault: Default<String> {
```
epoberezkin, line 279: don't think we need it - and not in app group anyway - it's a usual string default.

apps/ios/SimpleXChat/AppGroup.swift
```diff
@@ -276,6 +276,68 @@ public class IntDefault: Default<Int> {
...
+    public func get() -> String {
+        self.defaults.string(forKey: self.key)!
+    }
+}
+
+public class GenericDefault<T> {
```
epoberezkin, line 285: some hammer factory factory factory...
If it's used just for one type, better define it in place, without any abstractions.

PR #4319, opened 2024-06-12
core: servers summary api

spaced4ndy:


src/Simplex/Chat/Stats.hs
```diff
@@ -0,0 +1,86 @@
...
+    known :: Maybe Bool,
+    -- state:
+    -- Just if currently used in session, otherwise Nothing;
+    -- onlyProxiedSMPServers would always have state and statsData as Nothing
+    state :: Maybe SMPServerState,
+    -- statsData:
+    -- even if state is Nothing, statsData can be Just - server could be used earlier in session
+    -- or in previous sessions and stats for it were restored
+    statsData :: Maybe AgentSMPServerStatsData
+  }
```
spaced4ndy, line 43: initially I tried more strict types that wouldn't allow impossible states, e.g. `CurrentlyUsedSMPServerSummary` had `state :: SMPServerState` (not Maybe), `PreviouslyUsedSMPServerSummary` didn't have `state`, `onlyProxiedSMPServers` was simply list `[SMPServer]`.

But after also accounting for `known` - it required creating yet another type for onlyProxiedSMPServers, because server can be disabled/known and still have no stats and state and so be "only proxied" - I got replaced these more precise types with a single `SMPServerSummary` with Maybe fields

src/Simplex/Chat/Stats.hs
```diff
@@ -0,0 +1,282 @@
...
+import Simplex.Messaging.Agent.Client
+import Simplex.Messaging.Agent.Protocol (UserId)
+import Simplex.Messaging.Agent.Stats
+import Simplex.Messaging.Parsers (defaultJSON)
+import Simplex.Messaging.Protocol
+
+data PresentedServersSummary = PresentedServersSummary
+  { statsStartedAt :: UTCTime,
+    userServersSummary :: ServersSummary,
+    allServersSummary :: ServersSummary
```
epoberezkin, line 23:
```suggestion
    currentUserServers :: ServersSummary,
    allUsersServers :: ServersSummary
```

src/Simplex/Chat/Stats.hs
```diff
@@ -0,0 +1,282 @@
...
+                  then M.insertWith combineSummaries server summaryForAll allAcc
+                  else allAcc
+           in (newUAcc, newAllAcc)
+        combineSummaries
+          (SMPServerSummary _ _ sessions1 subs1 stats1)
+          (SMPServerSummary smpServer known sessions2 subs2 stats2) =
+            SMPServerSummary
+              { smpServer,
+                known,
+                sessions = addMaybes addServerSessions sessions1 sessions2,
```
epoberezkin, line 173:
```suggestion
                sessions = addServerSessions <$> sessions1 <*> sessions2,
```

PR #4321, opened 2024-06-13
ios: servers summary

spaced4ndy:


apps/ios/Shared/Views/ChatList/ServersSummaryView.swift
```diff
@@ -0,0 +1,571 @@
...
+        }
+    }
+
+    var wifiColor: (Color, Double, Double) {
+        if activeSubs > 0 {
+            let wifiVariableValue = ( // wifi has 3 sections
+                activeSubsPercentage >= 1 ? 1
+                : (activeSubsPercentage >= 0.5 && activeSubsPercentage < 1) ? 0.6
+                : (activeSubsPercentage > 0 && activeSubsPercentage < 0.5) ? 0.3
+                : 0
```
spaced4ndy, line 329: in case of ChatListView indicator, maybe we should apply activeSubsPercentage directly, in that case "wifi" would be fully colored starting with 68% and "cellularbars" - with 76%

PR #4324, opened 2024-06-13
ios: implemented disappearing search bar on chat list

diogofcunha:
## Demo

https://github.com/simplex-chat/simplex-chat/assets/10208017/8350e7a3-9a68-4666-ae12-eb8670ff038c

## Implementation of scroll tracking
After extensive testing, the most reliable method for tracking scrolling is to monitor the topmost rendered row and base the scroll direction on that when we are dealing with `List`. Tracking seems more reliable on `ScrollView` but for some reason it does prevent tapables to work correctly inside the scroll view, there are several reported bugs of these kind of use cases out there, specially related to swiftUI. 

This approach has proven to be the most stable. Tracking the general scrolling maxY and minY is unreliable as on list it does jump up and down randomly and causes flickering in the visibility of the bar and prone to hidden bugs. Unlike on the web, determining the first rendered row is quite straightforward and the same approach can now be taken for other screens.


apps/ios/Shared/Views/ChatList/ChatListView.swift
```diff
@@ -55,10 +61,56 @@ struct ChatListView: View {
...
+                    tbcontent
+                        .scaleEffect(x: 1, y: oneHandUI ? -1 : 1, anchor: .center)
+                        .frame(maxWidth: .infinity)
+                        .padding(.horizontal, 25)
+                        .padding(.bottom, 5)
+                        .toolbar(.hidden, for: .bottomBar)
+                }
+                
+                if !chatModel.chats.isEmpty && searchVisible {
+                    VStack {
```
epoberezkin, line 113: not needed

apps/ios/Shared/Views/ChatList/ChatListView.swift
```diff
@@ -9,17 +9,19 @@
...
     @FocusState private var searchFocussed
     @State private var searchText = ""
     @State private var searchShowingSimplexLink = false
     @State private var searchChatFilteredBySimplexLink: String? = nil
-
+    @State var topVisibleRowIndex: Int? = nil;
+    @State private var searchVisible: Bool = true;
     @AppStorage(DEFAULT_SHOW_UNREAD_AND_FAVORITES) private var showUnreadAndFavorites = false
     @AppStorage(DEFAULT_ONE_HAND_UI) private var oneHandUI = false
+    let toolbarContent: ToolbarContent?
```
epoberezkin, line 24: why is it passed as parameter

PR #4348, opened 2024-06-23
iOS Timeline: 1. Hashable Conformance

levitatingpineapple:
Part of `Timeline` feature:

1. [Hashable Conformance](https://github.com/simplex-chat/simplex-chat/pull/4348)
2. [Views](https://github.com/simplex-chat/simplex-chat/pull/4350)
3. [Integration](https://github.com/simplex-chat/simplex-chat/pull/4351)

## Contents

[UICollectionViewDiffableDataSource](https://developer.apple.com/documentation/uikit/uicollectionviewdiffabledatasource) uses hash values, 
to drive cell state updates and provide coherent animations. 
This PR adds (mostly) synthesised conformance to chat and API types.

PR #4350, opened 2024-06-23
iOS Timeline: 2. Views

levitatingpineapple:
Part of `Timeline` feature:

1. [Hashable Conformance](https://github.com/simplex-chat/simplex-chat/pull/4348)
2. [Views](https://github.com/simplex-chat/simplex-chat/pull/4350)
3. [Integration](https://github.com/simplex-chat/simplex-chat/pull/4351)

## Motivation

The current timeline implementation has high CPU and battery usage during scrolling,
which can result in dropped frames.
This is especially evident:

- In [Low Power Mode](https://support.apple.com/en-us/101604)
- On an Older Device (The app still supports [1st gen iPhone SE](https://en.wikipedia.org/wiki/IPhone_SE_(1st_generation)))
- When ChatItems include media or files

There are two main causes:

### 1. State Management

The current cell dequeuing implementation causes the whole view hierarchy to be reevaluated, for every loaded cell. 
While SwiftUI will reuse the underlying views, it still needs to recompute `body` of all of the views for diffing. 
This can be observed, by setting a random colour as cell background. 
Notice, how (on the left side) every cell changes colour, when a new one comes into the view:

https://github.com/simplex-chat/simplex-chat/assets/28978251/41f3434b-8662-4e89-8275-d0dcbb61f0ee

### 2. Chat Item Merging

The app will merge similar `ChatItem`s. 
Current implementation will still render a zero width cell with an [empty ZStack](https://github.com/simplex-chat/simplex-chat/blob/85dc467f74a46b88112bcb20a2b909eb82846923/apps/ios/Shared/Views/Chat/ChatView.swift#L560). 
This means - when there are many merged items, 
the app will have to process all of the zero height cells at once causing a lag spike.

### Solving State Management

The first issue is addressed, by replacing custom cell dequeuing with standard [UITableViewController](https://developer.apple.com/documentation/uikit/uitableviewcontroller) paired with [UIHostingConfiguration](https://developer.apple.com/documentation/SwiftUI/UIHostingConfiguration). This approach facilitates good performance and additional flexibility, 
while still reusing existing SwiftUI views for rendering `ChatItem`s.

Unfortunately, this approach is not compatible with existing `UIKit` backed context menu,
which increases scope of this PR. 
However, I believe - the additional stability and ease of iteration should offset
the cost of reimplementing the context menu in SwiftUI.

### Solving Item Merging

The second issue can be solved by adding an additional layer of `Timeline.Item` abstraction,
which handles the relationship between cells and chat items. 
`Timeline.Item` represents a single timeline cell, which contains at least one `ChatItem` 
with any additional chat items, which have been merged together.

Additionally this approach allows us to have required data available in a single structure,
and avoid index calculations, which can be error prone.

## Regressions

While the solution is working great, there are few features that might be better solved separately:

- Media Previews are too large
In SwiftUI views are expected to determine their own size, given parent views [ProposeViewSize](https://developer.apple.com/documentation/swiftui/proposedviewsize), 
so the current approach with passing `maxWidth` trough the view hierarchy won't work.
While this is a fairly straight forward task (since we don't need to use UIKit context menu anymore),
I think it would be better solved in a separate PR, as to not increase the scope even further.

- Simpler "scroll to unread" button
The new implementation does not distinguish between unread items above or below visible cells, 
and simply scrolls to the oldest unread item.
There is some additional complexity in making this work, 
so it could be done together with implementing non-contiguous timeline 
(a.k.a. jumping chat items, that have not been yet loaded)

Let me know.

## Result

Heres a demo of two iPhone 12mini's side-by-side:

https://github.com/simplex-chat/simplex-chat/assets/28978251/82cfe3e0-d3da-4dd6-9962-774b8c788f68

### DIFF
--- a/apps/ios/Shared/Model/ChatModel.swift
+++ b/apps/ios/Shared/Model/ChatModel.swift
@@ -728,6 +728,15 @@ final class ChatModel: ObservableObject {
             logger.error("apiChatItemReaction error: \(responseError(error))")
         }
     }
+
+    func loadGroupMembers(groupInfo: GroupInfo) async {
+        let groupMembers = await apiListMembers(groupInfo.groupId)
+        await MainActor.run {
+            if chatId == groupInfo.id {
+                self.groupMembers = groupMembers.map { GMember.init($0) }
+            }
+        }
+    }
 }
 
 struct ShowingInvitation {
--- a/apps/ios/Shared/Model/ChatModel.swift
+++ b/apps/ios/Shared/Model/ChatModel.swift
@@ -711,6 +711,23 @@ final class ChatModel: ObservableObject {
             .unknown
         }
     }
+
+    func set(reaction: MsgReaction, for chatItem: ChatItem, chatInfo: ChatInfo, add: Bool) async {
+        do {
+            let chatItem = try await apiChatItemReaction(
+                type: chatInfo.chatType,
+                id: chatInfo.apiId,
+                itemId: chatItem.id,
+                add: add,
+                reaction: reaction
+            )
+            await MainActor.run {
+                updateChatItem(chatInfo, chatItem)
+            }
+        } catch let error {
+            logger.error("apiChatItemReaction error: \(responseError(error))")
+        }
+    }
 }
 
 struct ShowingInvitation {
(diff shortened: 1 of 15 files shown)


levitatingpineapple:
recheck

PR #4391, opened 2024-07-03
ios: UI to translate chat messages

levitatingpineapple:
# Contents

Adds translate sheet to chat item menu.

# Privacy Considerations

While iOS has on-device translation, it is not enabled by default.
On initial use the OS will present a popup, informing user that text for translation will be sent to apple.
We might consider additional prompt, informing that on-device translation can be enabled in:
`Settings -> Translate -> On-Device Mode`

# Screenshots

https://github.com/simplex-chat/simplex-chat/assets/28978251/fc98bfed-edf7-44f5-8ff6-cf4ecbb77381

### DIFF
--- a/apps/ios/Shared/Views/Chat/ChatView.swift
+++ b/apps/ios/Shared/Views/Chat/ChatView.swift
@@ -839,7 +839,7 @@ struct ChatView: View {
                     shareButton(ci)
                     copyButton(ci)
                 }
-                if mc.isText {
+                if !mc.text.isEmpty {
                     translateButton(text: mc.text)
                 }
                 if let fileSource = fileSource, fileExists {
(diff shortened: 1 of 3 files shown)


epoberezkin:
can we restrict it to on-device?

levitatingpineapple:
> can we restrict it to on-device?

Thats out of our control. 
Best we can do is inform.

epoberezkin:
We should show the alert asking to open the correct settings and configure translations, recommending to enable on-device - is it possible? The alert would have Open settings and Cancel button.

And then not proceed with the translation until the user opens the settings via this alert.

levitatingpineapple:
https://github.com/simplex-chat/simplex-chat/pull/4906

PR #4398, opened 2024-07-04
android, desktop: servers summary

spaced4ndy:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/SimpleXAPI.kt
```diff
@@ -411,6 +413,23 @@ object ChatController {
...
+
+    val r = sendCmd(rh, CC.GetAgentServersSummary(userId))
+
+    if (r is CR.AgentServersSummary) {
+      return r.serversSummary;
+    } else {
+      Log.e(TAG, "getAgentServersSummary bad response: ${r.responseType} ${r.details}")
+      throw Exception("failed to get agent servers summary ${r.responseType} ${r.details}")
+    }
+  }
```
spaced4ndy, line 427: in kotlin we usually return null when response is unexpected, instead of throwing error

```suggestion
  suspend fun getAgentServersSummary(rh: Long?): PresentedServersSummary? {
    val userId = currentUserId("getAgentServersSummary")

    val r = sendCmd(rh, CC.GetAgentServersSummary(userId))

    if (r is CR.AgentServersSummary) return r.serversSummary
    Log.e(TAG, "getAgentServersSummary bad response: ${r.responseType} ${r.details}")
    return null
  }
```

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/SimpleXAPI.kt
```diff
@@ -411,6 +413,23 @@ object ChatController {
...
+
+  suspend fun resetAgentServersStats() {
+    sendCommandOkResp(null, CC.ResetAgentServersStats())
```
spaced4ndy, line 430: I think this actually requires remote host parameter (`rh`) to differentiate resetting stats for host/controller. i.e. if you reset when using mobile from desktop it would be expected to reset for mobile

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/SimpleXAPI.kt
```diff
@@ -920,6 +939,19 @@ object ChatController {
...
+    val userId = currentUserId("reconnectServer")
+
+    val r = sendCmd(rh, CC.ReconnectServer(userId, server))
+    return when (r) {
+      is CR.CmdOk -> true
+      else -> {
+        Log.e(TAG, "reconnectServer bad response: ${r.responseType} ${r.details}")
+        false
+      }
+    }
```
spaced4ndy, line 952: could probably use `sendCommandOkResp`

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/SimpleXAPI.kt
```diff
@@ -3203,7 +3247,8 @@ data class NetCfg(
   val tcpKeepAlive: KeepAliveOpts?,
   val smpPingInterval: Long, // microseconds
   val smpPingCount: Int,
-  val logTLSErrors: Boolean = false
+  val logTLSErrors: Boolean = false,
+  val showSubscriptionPercentage: Boolean = false
```
spaced4ndy, line 3251: this isn't a part of net config - net config only stores settings for network transport. `showSubscriptionPercentage` is a purely UI setting to show/hide percent.

(NetCfg wouldn't correctly decode/encode as backend doesn't have this field)

```suggestion
```

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/SimpleXAPI.kt
```diff
@@ -3428,6 +3473,160 @@ data class TimedMessagesPreference(
...
+@Serializable
+data class SMPServerSummary(
+  val smpServer: String,
+  val known: Boolean? = null,
+  val sessions: ServerSessions? = null,
+  val subs: SMPServerSubs? = null,
+  val stats: AgentSMPServerStatsData? = null
+) {
+  val id: String
+    get() = smpServer
```
spaced4ndy, line 3509: it may be not needed in kotlin (I don't remember w/t looking at code), it was only required in swift for ForEach I believe.. can keep it for now and remove if not used later

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/SimpleXAPI.kt
```diff
@@ -3428,6 +3473,160 @@ data class TimedMessagesPreference(
...
+  val xftpServer: String,
+  val known: Boolean? = null,
+  val sessions: ServerSessions? = null,
+  val stats: AgentXFTPServerStatsData? = null,
+  val rcvInProgress: Boolean,
+  val sndInProgress: Boolean,
+  val delInProgress: Boolean
+) {
+  val id: String
+    get() = xftpServer
```
spaced4ndy, line 3611: same

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ServersSummaryView.kt
```diff
@@ -0,0 +1,992 @@
...
+    sortedServers.map { svr -> SMPServerView(srvSumm = svr, statsStartedAt = statsStartedAt, rh = rh) }
+  }
+  if (footer != null) {
+    SectionTextFooter(
+      footer
+    )
+  }
+}
+
+fun prettySize(sizeInKB: Long): String {
```
spaced4ndy, line 232: could refactor this with `formatBytes` from Util, but that one prints with zeros after dot (e.g. "123.00 MB"), so I kept both. maybe we should reuse this logic for `formatBytes` instead

avently:
Added a commit that allows to close servers view by clicking outside of servers view on desktop
https://github.com/simplex-chat/simplex-chat/pull/4398/commits/68277f4e649e64f483d31c9498a296aaf8d85830

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/SimpleXAPI.kt
```diff
@@ -411,6 +413,20 @@ object ChatController {
...
+
+    val r = sendCmd(rh, CC.GetAgentServersSummary(userId), log = false)
+
+    if (r is CR.AgentServersSummary) return r.serversSummary
+    Log.e(TAG, "getAgentServersSummary bad response: ${r.responseType} ${r.details}")
+    return null
+  }
+
+  suspend fun resetAgentServersStats(rh: Long?): Boolean {
+    return sendCommandOkResp(rh, CC.ResetAgentServersStats())
```
avently, line 427: Kotlin allows to write one line functions this way:

```kotlin
  suspend fun resetAgentServersStats(rh: Long?): Boolean = sendCommandOkResp(rh, CC.ResetAgentServersStats())
```

So no need to print "return" explicitly

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/SimpleXAPI.kt
```diff
@@ -920,6 +940,16 @@ object ChatController {
...
   }
 
+  suspend fun reconnectServer(rh: Long?, server: String): Boolean {
+    val userId = currentUserId("reconnectServer")
+
+    return sendCommandOkResp(rh, CC.ReconnectServer(userId, server))
+  }
+
+  suspend fun reconnectAllServers(rh: Long?): Boolean {
+    return sendCommandOkResp(rh, CC.ReconnectAllServers())
```
avently, line 950: can be simplified by making it one line function without "return"

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/helpers/Section.kt
```diff
@@ -287,6 +287,63 @@ fun InfoRow(title: String, value: String, icon: Painter? = null, iconTint: Color
   }
 }
 
+fun numOrDash(n: Number): String {
+  return if (n.toLong() == 0L) "-" else n.toString()
```
avently, line 291: can be one line

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ServersSummaryView.kt
```diff
@@ -0,0 +1,990 @@
...
+
+fun subscriptionStatusColorAndPercentage(
+  online: Boolean,
+  socksProxy: String?,
+  subs: SMPServerSubs,
+  sess: ServerSessions
+): SubscriptionStatus {
+
+  fun roundedToQuarter(n: Float): Float {
+    return when {
```
avently, line 81: can be
```kotlin
fun () = when {
//...
}
```

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ServersSummaryView.kt
```diff
@@ -0,0 +1,990 @@
...
+@Composable
+fun ModalData.SMPServerSummaryView(
+  rh: RemoteHostInfo?,
+  close: () -> Unit,
+  summary: SMPServerSummary,
+  statsStartedAt: Instant
+) {
+  ModalView(
+    close = {
+      close()
```
avently, line 611: close = close

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ServersSummaryView.kt
```diff
@@ -0,0 +1,990 @@
...
+      ) {
+        selectedUserCategory.value = PresentedUserCategory.CURRENT_USER
+      } else {
+        showUserSelection = true
+      }
+    }
+
+    DisposableEffect(Unit) {
+      onDispose {
+        scope.cancel()
```
avently, line 738: again, why do you need to cancel it?

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ChatListView.kt
```diff
@@ -243,6 +262,54 @@ private fun ChatListToolbar(searchInList: State<TextFieldValue>, drawerState: Dr
...
+
+  fun stopTimer() {
+    timer?.cancel()
+    timer = null
+  }
+
+  DisposableEffect(Unit) {
+    onDispose {
+      stopTimer()
+      scope.cancel()
```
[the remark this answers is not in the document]
diogofcunha: if it cancels automatically it is fine. Copied the pattern from the chat list view 🙂

PR #4399, opened 2024-07-04
rfc: conversational agent

epoberezkin:


docs/rfcs/2024-07-04-conversational-agent.md
```diff
@@ -0,0 +1,55 @@
...
+
+Overall, this appears to be the simplest and the most effective approach.
+
+Conversation(s) with the agent would be stored in the usual database, and the agent would be interactive - it will be called by the code when the message is added to this conversation.
+
+Questions to consider:
+- ux:
+  - should it be one conversation for all user profiles or separate conversations per profile? The latter seems safer from privacy and security points of view.
+- types:
+  - conversation type. Could use "direct" or may be a special type.
```
spaced4ndy, line 51: Local is more close, with direct there would be lots of parallel ifs and/or uncaught bugs. For example, if at some point it requires file functionality files would be uploaded/downloaded instead of simply being stored, etc. Maybe local as well would have unwanted interactions. For example, I don't think we'd want to allow forwarding to/from it. If it's fully scripted "conversation" with no free form I'm not even sure it requires full capabilities of conversation - like chat items, except some rendering logic that could be decomposed; and may require some other capabilities - for example remembering/"understanding" communication context/history. In that case it might be easier to have fully separate abstractions for it

spaced4ndy: It could still be presented as "chat" in UI if necessary, like pending connections and requests are presented as chat previews, but it doesn't necessarily have to use regular chat items, conversations apis, preferences, etc.

PR #4401, opened 2024-07-04
ios: make message corners rounded when selecting context menus

levitatingpineapple:
# Issues

1. Preview for items without frame clipped.

2. During context menu animation shadow is missing corner radius:

3. After context menu animation finishes - default preview corner radius does not match the item:

# Solution

All of the issues can be addressed by
- Ensuring that clipping is applied right before `.contextMenu` modifier
- providing a `.contentShape(.contextMenuPreview)`

# Dynamic Type

There is another glitch, which happens when the height of chat item is less than 2x hardcoded corner radius.
This can happen, when user has set a different text size using `Dynamic Type` (This is used by surprisingly high percentage of users).

This is addressed by using a custom shape - which properly collapses into a pill, in this situation.

### DIFF
--- a/apps/ios/Shared/Views/Helpers/ChatItemClipShape.swift
+++ b/apps/ios/Shared/Views/Helpers/ChatItemClipShape.swift
@@ -1,0 +1,63 @@
+//
+//  ChatItemClipShape.swift
+//  SimpleX (iOS)
+//
+//  Created by Levitating Pineapple on 04/07/2024.
+//  Copyright © 2024 SimpleX Chat. All rights reserved.
+//
+
+import SwiftUI
+import SimpleXChat
+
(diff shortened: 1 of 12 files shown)


apps/ios/Shared/Views/Helpers/ChatItemClipShape.swift
```diff
@@ -0,0 +1,63 @@
...
+                switch chatItem.content {
+                case 
+                    .sndMsgContent, 
+                    .rcvMsgContent, 
+                    .sndDeleted, 
+                    .rcvDeleted, 
+                    .sndModerated, 
+                    .rcvModerated, 
+                    .rcvBlocked: 18
+                default: 8
```
levitatingpineapple, line 47: Items without frame should still have some radius for the lifted preview

PR #4414, opened 2024-07-08
ios: share extension

levitatingpineapple:
Initial work on share-sheet.

Current Scope:
- Share a single file to single chat
- Progress var for direct chats
- Shares from current user

## Preview

 

https://github.com/user-attachments/assets/5ea401e5-94ab-4e82-9932-e81132bd87ed


apps/ios/SimpleX SE/ShareViewController.swift
```diff
@@ -0,0 +1,30 @@
...
+//
+//  Created by Levitating Pineapple on 08/07/2024.
+//  Copyright © 2024 SimpleX Chat. All rights reserved.
+//
+
+import UIKit
+import SwiftUI
+import SimpleXChat
+
+@objc(ShareViewController)
```
levitatingpineapple, line 13: No need for storyboards or wrapping in additional `UIHostingController`

apps/ios/SimpleXChat/hs_init.c
```diff
@@ -39,3 +39,19 @@ void haskell_init_nse(void) {
     char **pargv = argv;
     hs_init_with_rtsopts(&argc, &pargv);
 }
+
+void haskell_init_se(void) {
```
levitatingpineapple, line 43: a copy of `haskell_init_nse`

epoberezkin: I don't think we need a different set of parameters here, we can just use NSE one

apps/ios/SimpleX SE/ShareError.swift
```diff
@@ -0,0 +1,80 @@
...
+//  SimpleX SE
+//
+//  Created by Levitating Pineapple on 11/07/2024.
+//  Copyright © 2024 SimpleX Chat. All rights reserved.
+//
+
+import Foundation
+import SimpleXChat
+
+enum ShareError: LocalizedError {
```
epoberezkin, line 12: don't think we need that, assuming we want to show some alert.

We used that pattern a lot, out of ignorance I think, but you can just use ChatResponse and Alert, not sure there is a need to have some enum in the middle

apps/ios/SimpleX SE/ShareModel.swift
```diff
@@ -0,0 +1,181 @@
...
+                    withAnimation { self.isLoaded = true }
+                }
+            case let .failure(error):
+                await MainActor.run { self.error = error }
+            }
+        }
+
+        // Binding: Filter chats based on debounced search query
+        $search
+            .removeDuplicates()
```
epoberezkin, line 45: why do we remove duplicates here?

levitatingpineapple: To avoid doing unnecessary computations.
But yes - this probably could be done after throttling.

PR #4427, opened 2024-07-09
ios: improve new and existing chat interactions - new chat sheet, one hand ui, info views action buttons; new modes of contact deletion (keep conversation, only delete conversation)

diogofcunha:


apps/ios/Shared/Model/SimpleXAPI.swift
```diff
@@ -341,6 +341,11 @@ func loadChat(chat: Chat, search: String = "") {
         im.reversedChatItems = []
         let chat = try apiGetChat(type: cInfo.chatType, id: cInfo.apiId, search: search)
         m.updateChatInfo(chat.chatInfo)
+        if case let .direct(contact) = chat.chatInfo, cInfo.chatDeleted == false, chat.chatInfo.chatDeleted {
+            var updatedContact = contact
+            updatedContact.chatDeleted = false
+            m.updateContact(updatedContact)
+        }
```
spaced4ndy, line 348: probably not both are needed - we only need to update chat once, right? so either updateChatInfo or updateContact, smth like this:
```suggestion
        if case let .direct(contact) = chat.chatInfo, cInfo.chatDeleted == false, chat.chatInfo.chatDeleted {
            var updatedContact = contact
            updatedContact.chatDeleted = false
            m.updateContact(updatedContact)
        } else {
            m.updateChatInfo(chat.chatInfo)
        }
```

apps/ios/Shared/Views/NewChat/AddGroupView.swift
```diff
@@ -204,7 +206,7 @@ struct AddGroupView: View {
                 chat = c
             }
         } catch {
-            dismiss()
+            dismissAllSheets()
```
spaced4ndy, line 209: is there a reason to dismiss all in case of error? why not go back / allow to go back? maybe it's ok, just wondering

epoberezkin (approved):

PR #4434, opened 2024-07-10
ios: optimise mark-as-read performance

levitatingpineapple:
This PR includes two optimisations that improve the performance when marking items as read:

1. Precalculate indices, which is `O(n)` operation on a background thread.
2. Remove unnecessary `@State`, from `ChatItemWithMenu` what was causing additional `body` re-evaluations.
3. Remove `chatView` reference, because structs are passed as copy - so we where essentially creating a copy of `ChatView` for each loaded `ChatItemWithMenu` view.

* Also revert the scroll to bottom behaviour to always scroll till the end of the list.

### DIFF
--- a/apps/ios/Shared/Views/Chat/ChatView.swift
+++ b/apps/ios/Shared/Views/Chat/ChatView.swift
@@ -597,10 +597,10 @@ struct ChatView: View {
 
     private struct ChatItemWithMenu: View {
         let m = ChatModel.shared
-        let chat: Chat
+        @EnvironmentObject var theme: AppTheme
+        @ObservedObject var chat: Chat
         let chatItem: ChatItem
         let maxWidth: CGFloat
-        @EnvironmentObject var theme: AppTheme
         @Binding var composeState: ComposeState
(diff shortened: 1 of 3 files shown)


apps/ios/Shared/Views/Chat/ChatView.swift
```diff
@@ -250,19 +249,7 @@ struct ChatView: View {
             }
         }
     }
-
-    private func loadGroupMembers(_ groupInfo: GroupInfo, updateView: @escaping () -> Void = {}) async {
-        let groupMembers = await apiListMembers(groupInfo.groupId)
-        await MainActor.run {
-            if chatModel.chatId == groupInfo.id {
-                chatModel.groupMembers = groupMembers.map { GMember.init($0) }
-                chatModel.populateGroupMembersIndexes()
```
epoberezkin, line 259: looks like it's lost

levitatingpineapple: Added back 👍

PR #4435, opened 2024-07-10
multiplatform: improve new and existing chat interactions - new chat sheet, one hand ui, info views action buttons; new modes of contact deletion (keep conversation, only delete conversation)

diogofcunha:
#### Objective
Provide visibilily and easy way to manage all contacts and deleted chats. Give user total contol over what data is saved in device when it comes to contacts

### Contact card actions

Current context menus (...) on individual chats and groups have been a source of confusion for users and are generally missed as valuable features.

#### Objective

Remove confusion and complication associated with ... menus and increase interaction speed.

##### Remove ... menus
Keep only 1 action button per chat type (add member for groups, phone call for individual chats

##### New action buttons to be displayed for contacts
- search
- call
- video
- mute

##### New action buttons to be displayed for members
- message
- call
- video

##### New action buttons to be displayed for groups
- search
- add member
- mute

### One hand UI

Experimental feature in chats list that inverts list and UI.

#### Objective
Provide easy way for users to use app with single hand.

#### Objective
Drive deletion logic by contact type across screen (chat list, contact card, contact list) as opposed to based on screen context.

##### For normal contacts (ready, not deleted, active and chat not deleted):

Delete bring up 2 options:
1. Only delete conversation (show text saying that It can be recovered in new chat sheet)
2. Delete contact
2a) Keep conversation toggle, should be off by default
2b) Buttons delete and notify contact and delete without notification

##### Active contact with deleted conversation (accessible via new chat sheet):
Delete brings confirmation
1) Buttons delete and notify contact and delete without notification

##### Deleted contact:
(Can be someone that deleted me, can be a contact deleted with conversation kept)

Delete brings alert
1) Confirm deletion (Confirm/Delete button)


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatInfoView.kt
```diff
@@ -535,6 +775,165 @@ fun LocalAliasEditor(
   }
 }
 
+// when contact is a "contact card"
+@Composable
+private fun ConnectButton(openedFromChatView: Boolean, chat: Chat, contact: Contact, close: () -> Unit) {
```
spaced4ndy, line 780: I think we don't have a way to open chat info for contact cards now anymore (comparing to initial implementation), so we can remove this button. @diogofcunha correct?

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatInfoView.kt
```diff
@@ -535,6 +775,165 @@ fun LocalAliasEditor(
...
+            }
+          }
+        },
+        hostDevice = hostDevice(chat.remoteHostId),
+      )
+    }
+  )
+}
+
+private fun infoConnectContactViaAddress(openedFromChatView: Boolean, chat: Chat, contact: Contact, incognito: Boolean, close: () -> Unit) {
```
spaced4ndy, line 815: and this function

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ChatListView.kt
```diff
@@ -154,9 +182,14 @@ fun ChatListView(chatModel: ChatModel, settingsState: SettingsViewState, setPerf
 }
 
 @Composable
-private fun OnboardingButtons(openNewChatSheet: () -> Unit) {
+private fun OnboardingButtons() {
```
spaced4ndy, line 185: OnboardingButtons can be removed as well, I think, as they are no longer shown (we always have at least non deletable local notes) and we can't even test whether new changes haven't broken them

PR #4438, opened 2024-07-11
ios: make image utilities accessible from extensions

levitatingpineapple:
- Moves `ImageUtils` to the framework to allow access from extensions
- Adds convenience `intit` to create `UIImage` from base64 encoded string

### DIFF
--- a/apps/ios/SimpleXChat/ImageUtils.swift
+++ b/apps/ios/SimpleXChat/ImageUtils.swift
@@ -7,18 +7,18 @@
 //
 
 import Foundation
-import SimpleXChat
 import SwiftUI
 import AVKit
+import SwiftyGif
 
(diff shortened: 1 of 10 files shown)


apps/ios/SimpleXChat/ImageUtils.swift
```diff
@@ -340,4 +339,12 @@ extension UIImage {
         }
         return self
     }
+
+    public convenience init?(imageString: String?) {
```
epoberezkin, line 343:
```suggestion
    public convenience init?(base64Encoded: String?) {
```

levitatingpineapple: @epoberezkin Updated

PR #4439, opened 2024-07-11
ios: readable errors

spaced4ndy:


apps/ios/SimpleXChat/API.swift
```diff
@@ -292,6 +292,18 @@ func chatErrorString(_ err: ChatError) -> String {
     return String(describing: err)
 }
 
+public func readableError(_ err: Error) -> String {
```
spaced4ndy, line 295: will be used in alerts instead of `responseError` (above)

PR #4444, opened 2024-07-11
core: use disabled servers as known for private message routing

epoberezkin:


src/Simplex/Chat.hs
```diff
@@ -172,15 +172,20 @@ defaultChatConfig =
...
+      (presetServerCfg True)
+      [ "smp://h--vW7ZSkXPeOUpfxlFGgauQmXNFOzGoizak7Ult7cw=@smp15.simplex.im,oauu4bgijybyhczbnxtlggo6hiubahmeutaqineuyy23aojpih3dajad.onion",
+        "smp://hejn2gVIqNU6xjtGM3OwQeuk8ZEbDXVJXAlnSBJBWUA=@smp16.simplex.im,p3ktngodzi6qrf7w64mmde3syuzrv57y55hxabqcq3l5p6oi7yzze6qd.onion",
+        "smp://ZKe4uxF4Z_aLJJOEsC-Y6hSkXgQS5-oc442JQGkyP8M=@smp17.simplex.im,ogtwfxyi3h2h5weftjjpjmxclhb5ugufa5rcyrmg7j4xlch7qsr5nuqd.onion",
+        "smp://PtsqghzQKU83kYTlQ1VKg996dW4Cw4x_bvpKmiv8uns=@smp18.simplex.im,lyqpnwbs2zqfr45jqkncwpywpbtq7jrhxnib5qddtr6npjyezuwd3nqd.onion",
+        "smp://N_McQS3F9TGoh4ER0QstUf55kGnNSd-wXfNPZ7HukcM=@smp19.simplex.im,i53bbtoqhlc365k6kxzwdp5w3cdt433s7bwh3y32rcbml2vztiyyz5id.onion"
+      ]
+      <> map
+        (presetServerCfg False)
+        []
```
spaced4ndy, line 188: probably shouldn't be empty or present

src/Simplex/Chat.hs
```diff
@@ -3225,9 +3227,8 @@ receiveViaCompleteFD user fileId RcvFileDescr {fileDescrText, fileDescrComplete}
     getUnknownSrvs :: [XFTPServer] -> CM [XFTPServer]
     getUnknownSrvs srvs = do
       ChatConfig {defaultServers = DefaultAgentServers {xftp = defXftp}} <- asks config
-      storedSrvs <- map (\ServerCfg {server} -> protoServer server) <$> withStore' (`getProtocolServers` user)
-      let defXftp' = L.map protoServer defXftp
-          knownSrvs = fromMaybe defXftp' $ nonEmpty storedSrvs
+      storedSrvs <- withStore' (`getProtocolServers` user)
+      let knownSrvs = L.map (\ServerCfg {server} -> protoServer server) $ fromMaybe defXftp $ nonEmpty storedSrvs
```
spaced4ndy, line 3231: I think we should refactor shared logic of retrieving user "used" servers that reads server list from db, takes default from config, and returns default if list from db is empty. there are several places (`getUnknownSrvs`, `GetAgentServersSummary`, `APIGetUserProtoServers`) that all do basically this same thing but in different ways and it's hard to find common pattern in what they're doing if you don't know what you're looking for

PR #4446, opened 2024-07-12
core: faster tracking of active subscriptions; ui: only track in foreground

epoberezkin:


spaced4ndy (approved):

apps/ios/Shared/Views/ChatList/ChatListView.swift
```diff
@@ -266,23 +266,19 @@ struct ChatListView: View {
...
     @State private var timer: Timer? = nil
     @State private var timerCounter = 0
     @State private var showServersSummary = false
 
     @AppStorage(DEFAULT_SHOW_SUBSCRIPTION_PERCENTAGE) private var showSubscriptionPercentage = false
 
-    // Constants for the intervals
-    let initialInterval: TimeInterval = 1.0
-    let regularInterval: TimeInterval = 3.0
-    let initialPhaseDuration: TimeInterval = 10.0 // Duration for initial phase in seconds
```
spaced4ndy, line 280: in addition to checking active, there's also following change to timers to simplify logic and reduce number of requests when on servers summary view:

initially there were two timers - one smart with increasing interval to reduce polling after some time in chat list; one simple that did polling every second in summary view. This removes smart timer and only leaves single timer same as in android, with state being passed from chat list view to summary view

PR #4448, opened 2024-07-12
android: implemented one hand ui for chat list screen

diogofcunha:
Adds a usable one hand UI on android.

A recording of the behaviour:

https://github.com/user-attachments/assets/64aed770-ae08-4211-87d9-f0982f7e7866


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ChatListView.kt
```diff
@@ -91,13 +116,20 @@ fun ChatListView(chatModel: ChatModel, settingsState: SettingsViewState, setPerf
     drawerGesturesEnabled = appPlatform.isAndroid,
     floatingActionButton = {
       if (searchText.value.text.isEmpty() && !chatModel.desktopNoUserNoRemote && chatModel.chatRunning.value == true) {
+        var bottom = DEFAULT_PADDING
+        if (oneHandUI.state.value) {
+          bottom = DEFAULT_BOTTOM_PADDING
+        } else {
+          bottom -= 16.dp
+        }
```
spaced4ndy, line 124:
```suggestion
        val bottomPadding = if (oneHandUI.state.value) DEFAULT_BOTTOM_PADDING else DEFAULT_PADDING - 16.dp
```

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ChatListView.kt
```diff
@@ -91,13 +116,20 @@ fun ChatListView(chatModel: ChatModel, settingsState: SettingsViewState, setPerf
...
+
         FloatingActionButton(
           onClick = {
             if (!stopped) {
               if (newChatSheetState.value.isVisible()) hideNewChatSheet(true) else showNewChatSheet()
             }
           },
-          Modifier.padding(end = DEFAULT_PADDING - 16.dp + endPadding, bottom = DEFAULT_PADDING - 16.dp).size(AppBarHeight * fontSizeSqrtMultiplier),
+          Modifier.padding(end = DEFAULT_PADDING - 16.dp + endPadding, bottom = bottom).size(AppBarHeight * fontSizeSqrtMultiplier),
```
spaced4ndy, line 132: something's wrong with floating action button padding in both modes - when it's clicked it's not shown over the initial button shape - the opened menu together with button should be padded so it's on the same height

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ChatListView.kt
```diff
@@ -226,10 +263,26 @@ private fun ChatListTopBar(stopped: Boolean) {
...
   DefaultTopAppBar(
+    navigationButton = {
+      if (chatModel.users.isEmpty() && !chatModel.desktopNoUserNoRemote) {
+        NavigationButtonMenu { scope.launch { if (drawerState.isOpen) drawerState.close() else drawerState.open() } }
+      } else {
+        val users by remember { derivedStateOf { chatModel.users.filter { u -> u.user.activeUser || !u.user.hidden } } }
+        val allRead = users
+          .filter { u -> !u.user.activeUser && !u.user.hidden }
+          .all { u -> u.unreadCount == 0 }
+        UserProfileButton(chatModel.currentUser.value?.profile?.image, allRead) {
```
spaced4ndy, line 277: in regular mode (one hand UI off) user picker is incorrectly located in the bottom

diogofcunha: fixed

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ChatListView.kt
diogofcunha: if you don't mind will skip these as per offline chat with EP. we will both remove the anymation and also stop having this button altogether in one hand

PR #4449, opened 2024-07-14
ios: toolbar and message entry area background color

epoberezkin:


apps/ios/Shared/Views/Chat/ComposeMessage/ComposeView.swift
```diff
@@ -377,6 +377,7 @@ struct ComposeView: View {
                 }
             }
         }
+        .background(Color(uiColor: .tertiarySystemGroupedBackground))
```
spaced4ndy, line 380: compose context/attachment rows look quite bad with this change (worse than just on background)

PR #4453, opened 2024-07-14
ui: different label for subscribed connections: Message push, Active connections

epoberezkin:


apps/ios/Shared/Views/ChatList/ServersSummaryView.swift
```diff
@@ -174,11 +174,11 @@ struct ServersSummaryView: View {
 
     private func smpSubsSection(_ totals: SMPTotals) -> some View {
         Section {
-            infoRow("Connections subscribed", numOrDash(totals.subs.ssActive))
+            infoRow("Active connections", numOrDash(totals.subs.ssActive))
             infoRow("Total", numOrDash(totals.subs.total))
         } header: {
             HStack {
-                Text("Message subscriptions")
+                Text("Message push")
```
spaced4ndy, line 181: I think "push" is more associated with sent messages - it's not clear that you mean "push from server" here. Maybe "Message reception"
```suggestion
                Text("Message reception")
```

apps/multiplatform/common/src/commonMain/resources/MR/base/strings.xml
```diff
@@ -2130,8 +2130,8 @@
   <string name="servers_info_messages_received">Messages received</string>
   <string name="servers_info_details">Details</string>
   <string name="servers_info_private_data_disclaimer">Starting from %s.\nAll data is private to your device.</string>
-  <string name="servers_info_subscriptions_section_header">Messages subscriptions</string>
-  <string name="servers_info_subscriptions_connections_subscribed">Connections subscribed</string>
+  <string name="servers_info_subscriptions_section_header">Message push</string>
```
spaced4ndy, line 2133:
```suggestion
  <string name="servers_info_subscriptions_section_header">Message reception</string>
```

PR #4462, opened 2024-07-16
ios: better errors

spaced4ndy:


apps/ios/SimpleXChat/API.swift
```diff
@@ -239,6 +239,14 @@ private func decodeUser_(_ jDict: NSDictionary) -> UserRef? {
     }
 }
 
+private func errorJson(_ jDict: NSDictionary) -> String? {
```
spaced4ndy, line 242: invalidJson errors in ios included full response json including user - this excludes user from invalidJson string

epoberezkin: wouldn't it exclude the whole thing if it's not error?

apps/ios/SimpleXChat/API.swift
```diff
@@ -268,6 +276,13 @@ func decodeProperty<T: Decodable>(_ obj: Any, _ prop: NSString) -> T? {
     return nil
 }
 
+func json(_ obj: Any) -> String? {
+    if let d = try? JSONSerialization.data(withJSONObject: obj) {
+        return String(decoding: d, as: UTF8.self)
+    }
+    return nil
+}
```
epoberezkin, line 284: we already have some function that does encoding, why this is necessary?

apps/ios/SimpleXChat/API.swift
spaced4ndy: but it can only be error there

apps/ios/SimpleXChat/API.swift
epoberezkin: prettyJSON

spaced4ndy: it's `prettyJSON` below, this one is regular, not pretty. it avoids unnecessary "\n"s which are packed into string anyway

epoberezkin: pass parameter :)

apps/ios/SimpleXChat/API.swift
epoberezkin: theoretically

epoberezkin: it is also used in chatCmdError

epoberezkin: probably it's the key higher up... In any case, seems safer to just keep it as is rather than null it, also avoids nullable return

if let chatError = jDict["chatError"] {
         json(chatError)
     } else {
         json(jDict)
     }

PR #4465, opened 2024-07-16
core: allow sending messages immediately on joinConnection, acceptContact

spaced4ndy:


src/Simplex/Chat/Types.hs
```diff
@@ -1588,9 +1588,9 @@ commandExpectedResponse = \case
   CFCreateConnGrpInv -> t INV_
   CFCreateConnFileInvDirect -> t INV_
   CFCreateConnFileInvGroup -> t INV_
-  CFJoinConn -> t OK_
+  CFJoinConn -> t JOINED_
```
epoberezkin, line 1591: is it backwards compatible?

epoberezkin: that is, is it saved on creation or checked when received?

PR #4472, opened 2024-07-17
ios: coordinate database access between share extension, the app and notifications extension

levitatingpineapple:
# Contents

- Adds database access management by requesting and observing suspension of the main app and NSE processes.
- Properly dismisses the sheet is an error is encountered
- Adds a loading spinner to provide feedback to the user, while waiting for database access


apps/ios/SimpleX SE/ShareModel.swift
```diff
@@ -99,6 +99,8 @@ class ShareModel: ObservableObject {
             )
             try apiSetEncryptLocalFiles(privacyEncryptLocalFilesGroupDefault.get())
             let isRunning = try apiStartChat()
+            // TODO: Avoid duplicate event loop observer
+            // Here we should check, if the chat is already running and only start the polling `recvSimpleXMsg` if `isRunning == false`
```
epoberezkin, line 103: my suggestion is to not rely on isRunning response, and have an internal tracking that this loop already started or not.

apps/ios/SimpleX SE/ShareViewController.swift
```diff
@@ -32,3 +32,22 @@ class ShareViewController: UIHostingController<ShareView> {
...
+/// Handle database access using existing (not-ideal) model, since
+///
+/// 1. Subscribe to lifecycle notifications:
+///     - `NSExtensionHostDidBecomeActive`
+///     - `NSExtensionHostWillResignActive`
+///     - `NSExtensionHostDidEnterBackground`
+///     - `NSExtensionHostWillEnterForeground`
+///
+/// 2. Add `SEState` using `SharedFileSubscriber`
+/// and suspend the chat in case there is a NSE Notification, while share sheet process is active.
```
epoberezkin, line 46: you mean suspend SE chat, right?

apps/ios/SimpleX SE/ShareViewController.swift
```diff
@@ -32,3 +32,22 @@ class ShareViewController: UIHostingController<ShareView> {
...
+///
+/// 3. Handle the case, where share-sheet and app is open simultaneously by sending
+/// a notification over `CFNotificationCenterGetDarwinNotifyCenter` 
+/// so the main app can register database changes and update it's `ChatModel`
+///
+/// 4. Consider protecting all database connection creations with additional as described in
+/// https://swiftpackageindex.com/groue/grdb.swift/v6.28.0/documentation/grdb/databasesharing#Use-the-WAL-mode
```
epoberezkin, line 53: we did try, it's not a reliable enough solution and creates a lot of additional migration challenges. But maybe

apps/ios/SimpleX SE/ShareViewController.swift
levitatingpineapple: Yes - so that the notification can be delivered.
Not entirely sure of the implications of this though.

apps/ios/SimpleX SE/ShareViewController.swift
```diff
@@ -32,3 +32,22 @@ class ShareViewController: UIHostingController<ShareView> {
...
+///     - `NSExtensionHostWillResignActive`
+///     - `NSExtensionHostDidEnterBackground`
+///     - `NSExtensionHostWillEnterForeground`
+///
+/// 2. Add `SEState` using `SharedFileSubscriber`
+/// and suspend the chat in case there is a NSE Notification, while share sheet process is active.
+///
+/// 3. Handle the case, where share-sheet and app is open simultaneously by sending
+/// a notification over `CFNotificationCenterGetDarwinNotifyCenter` 
+/// so the main app can register database changes and update it's `ChatModel`
```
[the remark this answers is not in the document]
levitatingpineapple: You can open share sheet form the app itself for example.
Also I can happen on iPad multitasking

apps/ios/SimpleX SE/ShareViewController.swift
epoberezkin: it's not needed here, as we never enabled subscriptions in the first place, and if we close the database it won't do anything... It might still continue trying to deliver messages though next time apple wakes it up... So maybe we should stop it instead of suspending. need to think

epoberezkin: maybe we should suspend indeed... and then foreground it if it is opened in the same process.

apps/ios/SimpleX SE/ShareViewController.swift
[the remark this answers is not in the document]
epoberezkin: right... How would it work then? do we need to suspend the app in this case? Want the app be in background anyway in this case, so suspended? or it would go into .inactive state?

apps/ios/Shared/Model/SuspendChat.swift
```diff
@@ -36,6 +36,18 @@ private func _suspendChat(timeout: Int) {
     }
 }
 
+let seSubscriber = seMessageSubscriber {
```
epoberezkin, line 39: I don't think it is correct...

apps/ios/SimpleX SE/SEChatState.swift
```diff
@@ -0,0 +1,79 @@
...
+
+    func set(itemId: ChatItem.ID, model: ShareModel) {
+        self.itemId = itemId
+        self.model = model
+    }
+
+    init() {
+        Task {
+            while true {
+                switch recvSimpleXMsg() {
```
epoberezkin, line 45: make async in the next PR

PR #4479, opened 2024-07-18
android, desktop: new chat sheet

diogofcunha:
### New chat sheet in android

https://github.com/user-attachments/assets/bef52457-a9e1-4c1f-99fb-c2bb2487a98e

### New chat sheet in desktop

https://github.com/user-attachments/assets/5f74a504-702d-42ed-8af8-a101266fb30c

https://github.com/user-attachments/assets/628399cc-f747-487e-a4c1-4435c177d770

### Accepting contacts on older version of the app

https://github.com/user-attachments/assets/7d0234aa-c583-4fd4-836b-0075471a9d60

### Accepting contacts on new version of the app

https://github.com/user-attachments/assets/cff64755-f828-40af-9e12-0ed405c4130f


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/newchat/NewChatSheet.kt
```diff
@@ -33,29 +41,58 @@ import kotlinx.coroutines.launch
...
+    Modifier.fillMaxSize(),
+  ) {
+    Box(contentAlignment = Alignment.Center) {
+      val bottomPadding = DEFAULT_PADDING
+      AppBarTitle(
+        stringResource(MR.strings.new_chat),
+        hostDevice(rh?.remoteHostId),
+        bottomPadding = bottomPadding
+      )
+    }
```
spaced4ndy, line 78: header jumps (has different top padding) when opening "Add contact" / "Scan / Past link" views - maybe the issue is in them but it's more noticeable now

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/newchat/NewChatSheet.kt
```diff
@@ -10,6 +13,7 @@ import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.*
 import androidx.compose.runtime.*
+import androidx.compose.runtime.saveable.rememberSaveable
```
spaced4ndy, line 16: many imports are unused

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/contacts/ContactListNavView.kt
```diff
@@ -0,0 +1,112 @@
...
+        else -> {}
+    }
+}
+
+@Composable
+fun ContactMenuItems(chat: Chat, contact: Contact, chatModel: ChatModel, showMenu: MutableState<Boolean>) {
+    if (contact.activeConn != null) {
+        ToggleFavoritesChatAction(chat, chatModel, chat.chatInfo.chatSettings?.favorite == true, showMenu)
+    }
+    DeleteContactAction(chat, chatModel, showMenu)
```
spaced4ndy, line 86: I think DeleteContactAction shouldn't be available on this view - it's not a "new chat" + it's not necessary to find a place to delete only contact w/t deleting conversation now as all options will be available from both chat list and chat info

spaced4ndy: favorite is probably ok to keep as it affects toggle above

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/contacts/ContactListNavView.kt
```diff
@@ -0,0 +1,112 @@
...
+    val showMenu = remember { mutableStateOf(false) }
+    val disabled = chatModel.chatRunning.value == false || chatModel.deletedChats.value.contains(chat.remoteHostId to chat.chatInfo.id)
+    LaunchedEffect(chat.id) {
+        showMenu.value = false
+        delay(500L)
+    }
+    val selectedChat = remember(chat.id) { derivedStateOf { chat.id == chatModel.chatId.value } }
+
+    when (chat.chatInfo) {
+        is ChatInfo.Direct -> {
```
spaced4ndy, line 27: for contact cards (e.g. "SimpleX Chat team" - and we plan to add 1 or 2 more) connectContactViaAddress alert should be shown on new chat sheet, same as contact requests. Also probably both contact cards and contact requests should be colored in accent color to differentiate from contacts and show they lead to some other action than simply opening

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/contacts/ContactsView.kt
```diff
@@ -0,0 +1,385 @@
...
+        }
+        itemsIndexed(filteredContactChats) { index, chat ->
+            val nextChatSelected = remember(chat.id, filteredContactChats) {
+                derivedStateOf {
+                    chatModel.chatId.value != null && filteredContactChats.getOrNull(index + 1)?.id == chatModel.chatId.value
+                }
+            }
+            SectionItemView(padding = PaddingValues(horizontal = DEFAULT_PADDING)) {
+                ContactListNavLinkView(chat, nextChatSelected)
+            }
```
spaced4ndy, line 337: tap area shouldn't be padded from borders of screen, it may be this or some other padding (i.e. when you tap/hold contact nav link view gray tap background should spread up to border)

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/contacts/ContactsView.kt
```diff
@@ -0,0 +1,385 @@
...
+                    onFocusChanged = {
+                        searchFocused = it
+                    }
+                )
+                Divider()
+            }
+
+            if (!searchFocused) {
+                contactActions()
+            }
```
spaced4ndy, line 318: I think these action buttons (Add/Scan/Create group) should have a bit more padding from search, would look better

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/contacts/ContactsView.kt
```diff
@@ -0,0 +1,385 @@
...
+) {
+    ContactsLayout(
+        contactActions = { ContactActionsSection(contactActions, rh) },
+        contactTypes = listOf(ContactType.CARD, ContactType.RECENT, ContactType.REQUEST),
+        contactListTitle = stringResource(MR.strings.contact_list_header_title).uppercase()
+    )
+}
+
+@Composable
+private fun ContactsSearchBar(listState: LazyListState, searchText: MutableState<TextFieldValue>, focused: Boolean, onFocusChanged: (hasFocus: Boolean) -> Unit) {
```
spaced4ndy, line 195: since now it's not simply contacts view, but New chat view, should say "Search or paste SimpleX link" and correctly react to pasting it - offer to connect / open known chats, etc. - basically copy look and behavior of chat list search

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/contacts/ContactListNavView.kt
spaced4ndy: as discussed, let's remove all context menus in this list

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/contacts/ContactListNavView.kt
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    val selectedChat = remember(chat.id) { derivedStateOf { chat.id == chatModel.chatId.value } }
+
+    when (chat.chatInfo) {
+        is ChatInfo.Direct -> {
+            ChatListNavLinkLayout(
+                chatLinkPreview = {
+                    tryOrShowError("${chat.id}ContactListNavLink", error = { ErrorChatListItem() }) {
+                        ContactPreviewView(chat, disabled)
```
spaced4ndy, line 44: interaction with contact cards seems to be broken (tapping it doesn't show any alert on contacts sheet now)

spaced4ndy: also let's filter them out in chat list - they would only be accessible from contacts list in new chat sheet - that was the point of keeping delete action for them there

apps/multiplatform/common/src/commonMain/resources/MR/base/strings.xml
```diff
@@ -2220,4 +2220,13 @@
   <string name="download_errors">Download errors</string>
   <string name="server_address">Server address</string>
   <string name="open_server_settings_button">Open server settings</string>
+
+  <!-- ContactsView.kt -->
+  <string name="contact_type_new">New</string>
+  <string name="contact_type_deleted">Deleted contacts</string>
```
spaced4ndy, line 2226: contact_type_deleted, contact_type_recent and icon_descr_pending_contact_connection seem to be unused translations - better to delete them so that translator don't have to translate them into all languages

apps/multiplatform/common/src/commonMain/resources/MR/base/strings.xml
```diff
@@ -2220,4 +2220,13 @@
...
+  <string name="chat_deleted">Deleted chats</string>
```
spaced4ndy, line 2227:
```suggestion
  <string name="deleted_chats">Deleted chats</string>
```
probably was some other text?

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/contacts/ContactsView.kt
```diff
@@ -0,0 +1,433 @@
...
+                AppBarTitle(
+                    stringResource(MR.strings.chat_deleted),
+                    hostDevice(rh?.remoteHostId),
+                    bottomPadding = bottomPadding
+                )
+            }
+
+            ContactsLayout(
+                contactActions = {},
+                contactTypes = listOf(ContactType.REMOVED),
```
spaced4ndy, line 139: opening deleted conversations (and even writing message) doesn't keep contact in chat list - we can do a simple fix by doing this for contacts with deleted conversation when opening chat with them:

```
        if (contact.chatDeleted) {
          chatModel.updateContact(chat.remoteHostId, contact.copy(chatDeleted = false))
        }
```

(variable names may differ - took it from a different place)

This way it would stay in chat list for current session even if user doesn't write message immediately - they can still re-open from chat list and write later. After app restart it won't be in list, but I think it's ok edge case - can be consider as "user didn't have intention to restore contact" (otherwise we'd have to do an api call to mark chatDeleted true in backend on open, which I think is unnecessary unless somebody complains)

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/contacts/ContactsView.kt
```diff
@@ -0,0 +1,433 @@
...
+    searchChatFilteredBySimplexLink: MutableState<String?>,
+    focused: Boolean,
+    onFocusChanged: (hasFocus: Boolean) -> Unit
+) {
+    Row(verticalAlignment = Alignment.CenterVertically) {
+        val focusRequester = remember { FocusRequester() }
+        Icon(painterResource(MR.images.ic_search), null, Modifier.padding(horizontal = DEFAULT_PADDING_HALF), tint = MaterialTheme.colors.secondary)
+        SearchTextField(
+            Modifier.weight(1f).onFocusChanged { onFocusChanged(it.hasFocus) }.focusRequester(focusRequester),
+            placeholder = stringResource(MR.strings.search_or_paste_simplex_link),
```
spaced4ndy, line 215: interesting - we don't show groups in this list, but if user pasts link to a known group perhaps we should show that single group?

currently it shows "No filtered contacts" and correctly doesn't offer to connect, but we could "surprise" here and allow to open group via known link.

is it a difficult change? if so, maybe better to do it separately from this pr

spaced4ndy: actually it should probably do same as Scan / Paste link view was doing before - open that group chat and show alert "Group already exists" (the interaction is different from chat list where it's shown in list without alert).

Scan / Paste link interaction is also broken btw - it correctly shows alert "Group already exists", but doesn't open chat - instead it goes back one modal, and if you close it the group chat is opened.

spaced4ndy: same for known contact (known by address) - interaction is similarly broken on Scan / Paste link

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/contacts/ContactListNavView.kt
```diff
@@ -0,0 +1,130 @@
...
+
+                            ModalManager.start.closeModals()
+                        }
+                        ContactType.REMOVED -> {
+                            openLoadedChat(chat, chatModel)
+                            chatModel.updateContact(rhId, chat.chatInfo.contact.copy(chatDeleted = false))
+                            ModalManager.start.closeModals()
+                        }
+                        ContactType.CARD -> {
+                            directChatAction(
```
spaced4ndy, line 66: for contact cards alert should be shown on new chat sheet, same as contact requests

spaced4ndy: currently it closes new chat sheet, instead if it's dismissed it would be treated as going back to new chat sheet, if it's accepted (connection started) it would close new chat sheet same as other interactions

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ChatListView.kt
```diff
@@ -35,28 +33,27 @@ import chat.simplex.common.views.onboarding.shouldShowWhatsNew
...
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.serialization.json.Json
 import java.net.URI
-import kotlin.time.Duration
 import kotlin.time.Duration.Companion.seconds
 
+private fun showNewChatSheet() {
+  ModalManager.center.closeModals()
+  ModalManager.center.showModalCloseable { close ->
+    NewChatSheet(rh = chatModel.currentRemoteHost.value, close)
```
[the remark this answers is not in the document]
spaced4ndy: it's not necessarily that interaction with known contact only - simply creating group does the trick as well - it seems to be some logic between opening chat from new chat sheet and opening new chat sheet itself

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/contacts/ContactsView.kt
```diff
@@ -0,0 +1,448 @@
...
+) {
+    ContactsLayout(
+        contactActions = { ContactActionsSection(contactActions, rh) },
+        contactTypes = listOf(ContactType.CARD, ContactType.RECENT, ContactType.REQUEST),
+        contactListTitle = stringResource(MR.strings.contact_list_header_title).uppercase()
+    )
+}
+
+@Composable
+private fun ContactsSearchBar(
```
spaced4ndy, line 204: on desktop after pasting a known contact address it requires 2 clicks on that contact to open it (first click reveals action buttons, only second click opens chat)

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ChatListView.kt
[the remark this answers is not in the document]
spaced4ndy: maybe some incorrect interaction NewChatConnectionStage?

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/contacts/ContactListNavView.kt
```diff
@@ -0,0 +1,131 @@
...
+        delay(500L)
+    }
+
+    LaunchedEffect(contactType) {
+        if (contactType == ContactType.RECENT) {
+            withBGApi {
+                loadedChat.value = chatModel.controller.apiGetChat(rhId, chat.chatInfo.chatType, chat.chatInfo.apiId)
+            }
+        }
+    }
```
spaced4ndy, line 35: what does this do? does it preload chat before it's opened, and does so for all "recent" (active) contacts in list? if so, it can be slowing ui quite a lot if there's a lot of active contacts in list, better to only load on opening

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ChatListView.kt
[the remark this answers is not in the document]
diogofcunha: was most likely bugs in the enum assignment, seems all good now  🙂

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/contacts/ContactsView.kt
diogofcunha: it seems like focus re-render cancels the click events with the movement of the screen. https://github.com/simplex-chat/simplex-chat/pull/4479/commits/1dbfb7aeea0a993d16b9a11723bc712386621b35 seems to be an hack that works on the issue

PR #4481, opened 2024-07-18
ios: add image message sending to share extension

levitatingpineapple:
- Adds utility for downsampling an image directly from disk instead of loading the original image in memory
- Adds support for sending image messages from share extension

# Comparing image compression

https://github.com/user-attachments/assets/7fd2e161-d8d1-4d65-8d49-85a116d506a9

### DIFF
--- a/apps/ios/SimpleX SE/ShareAPI.swift
+++ b/apps/ios/SimpleX SE/ShareAPI.swift
@@ -102,6 +102,9 @@ enum ShareError: LocalizedError {
     case missingAttachment
     case unsupportedFormat
     case encryptFile
+    case imageDownsampling
+    case previewGeneration
+    case mediaLoading
     case loadFileRepresentation(Error)
 
     public var errorDescription: String? {
@@ -120,6 +123,12 @@ enum ShareError: LocalizedError {
             "This file format is not supported"
         case .encryptFile:
             "File encryption has failed"
+        case .imageDownsampling:
+            "Failed to downsample selected image"
+        case .previewGeneration:
+            "Failed to generate Preview"
+        case .mediaLoading:
+            "Unable to load media"
         case .loadFileRepresentation(let error):
             "File could not be loaded: \(String(describing: error))"
         }
--- a/apps/ios/SimpleX SE/ShareModel.swift
+++ b/apps/ios/SimpleX SE/ShareModel.swift
@@ -10,6 +10,12 @@ import UniformTypeIdentifiers
 import SwiftUI
 import SimpleXChat
 
+/// Maximum size of hex encoded media previews
+private let MAX_DATA_SIZE: Int64 = 14000
+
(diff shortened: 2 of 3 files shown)


apps/ios/SimpleX SE/ShareAPI.swift
```diff
@@ -120,6 +123,12 @@ enum ShareError: LocalizedError {
             "This file format is not supported"
         case .encryptFile:
             "File encryption has failed"
+        case .imageDownsampling:
```
epoberezkin, line 126: let's please get rid of ShareError, and change sendMessage type to:

private func sendMessage() async -> ErrorAlert? {

apps/ios/SimpleX SE/ShareModel.swift
```diff
@@ -71,13 +77,58 @@ class ShareModel: ObservableObject {
...
+                if attachment.hasItemConformingToTypeIdentifier(UTType.gif.identifier) {
+                    guard let data = try? Data(contentsOf: url), let image = UIImage(data: data) else {
+                        return .failure(.mediaLoading)
+                    }
+                    guard let file = saveAnimImage(image) else {
+                        return .failure(.encryptFile)
+                    }
+                    guard let imageString = resizeImageToStrSize(image, maxDataSize: MAX_DATA_SIZE) else {
+                        return .failure(.previewGeneration)
+                    }
```
epoberezkin, line 97: can be one guard, with multiple conditions

apps/ios/SimpleX SE/ShareModel.swift
```diff
@@ -71,13 +77,58 @@ class ShareModel: ObservableObject {
...
+                } else {
+                    guard let image = downsampleImage(at: url, to: MAX_DOWNSAMPLE_SIZE) else {
+                        return .failure(.imageDownsampling)
+                    }
+                    guard let file = saveImage(image) else {
+                        return .failure(.encryptFile)
+                    }
+                    guard let imageString = resizeImageToStrSize(image, maxDataSize: MAX_DATA_SIZE) else {
+                        return .failure(.previewGeneration)
+                    }
```
epoberezkin, line 111: also can be one guard

apps/ios/SimpleX SE/ShareModel.swift
```diff
@@ -71,13 +77,58 @@ class ShareModel: ObservableObject {
...
+                    cryptoFile = file
+                    msgContent = .image(text: comment, image: imageString)
+                }
+
+            // Prepare Data message
+            case .data:
+                msgContent = .file(comment)
+                guard let file = saveFileFromURL(url) else { return .failure(.encryptFile) }
```
epoberezkin, line 119: this is not necessarily encryption error, it can be IO error, so it should be a generic "Error saving file"

PR #4484, opened 2024-07-19
core: fix nix build for iOS library with tagged JSON (e.g., Flutter apps)

epoberezkin:


tomquas:
confirmed to fix segfaults during opensllinit on ipad pro (model MPDY2FD/A). thx a lot!

PR #4486, opened 2024-07-19
ios: improve share extension completion handling

levitatingpineapple:
- reworks event loop to include additional conditions
- removes `ShareError` in favour of `ErrorAlert`

* To keep the scope managable - this PR does not include adding background tasks or chat shutdown


apps/ios/SimpleX SE/ShareViewController.swift
```diff
@@ -30,12 +30,32 @@ class ShareViewController: UIHostingController<ShareView> {
...
+                    self.extensionContext!.completeRequest(returningItems: [item]) { expired in
+                        self.handleShutDown(expired: expired)
                     }
                 }
             }
             Task { await MainActor.run { model.item = item } }
         }
     }
+
+    override func viewDidDisappear(_ animated: Bool) {
```
levitatingpineapple, line 42: Calls completion if the sheet gets manually dismissed.

apps/ios/SimpleX SE/SEChatState.swift
```diff
@@ -31,34 +31,56 @@ class SEChatState {
...
+                    guard let eventHandling, aChatItem.chatItem.id == itemId else { continue }
+                    switch eventHandling {
+                    case .directMessage:
+                        await MainActor.run { model?.bottomBar = .loadingBar(progress: 1) }
+                    case .group:
+                        try? await Task.sleep(for: .seconds(5))
+                        model?.completion?(nil)
                     }
-                default: break
+                // For direct chats run completion only after receiving `.complete` item state
```
levitatingpineapple, line 76: Here we check that the file description upload is completed
Is `aChatItem.chatItem.meta.itemStatus == .sndSent(sndProgress: .complete)` the right thing to check for?

apps/ios/SimpleX SE/ShareModel.swift
```diff
@@ -59,9 +67,17 @@ class ShareModel: ObservableObject {
...
             case let .success(item):
-                await MainActor.run { self.bottomBar = .loadingBar(progress: .zero) }
+                if item.chatInfo.chatType != .group {
+                    await MainActor.run {
+                        self.bottomBar = .loadingBar(progress: .zero)
+                    }
+                }
                 // Listen to the event loop for progress events
-                EventLoop.shared.set(itemId: item.chatItem.id, model: self)
+                EventLoop.shared.set(
```
epoberezkin, line 76: On fast networks we may have a race condition here that event would arrive before sendMessage returns (particularly when sending small file to a large group it is very likely to happen on any network), so event loop has to be fully configured for processing before calling sendMessage.

One option could be to stop and create the new loop on the next invocation. But it may result in the process being suspended before stopping the loop, so the next invocation needs to be able to stop the old loop too - probably it has to be tracked in some global variable...

apps/ios/SimpleX SE/ShareViewController.swift
```diff
@@ -30,12 +31,51 @@ class ShareViewController: UIHostingController<ShareView> {
...
+            if model.bottomBar.isLoading {
+                self.extensionContext!.completeRequest(returningItems: [item]) { expired in
+                    self.handleShutDown(expired: expired)
+                }
+            }
+        }
+    }
+
+    /// System will call this function twice:
+    /// - first time with `expired = false` giving time to do background processing
```
epoberezkin, line 55: don't we need to lock on semaphore in the first call?

apps/ios/SimpleX SE/ShareViewController.swift
```diff
@@ -30,12 +31,51 @@ class ShareViewController: UIHostingController<ShareView> {
...
+    /// - second time with `expired = true` shortly before terminating the process
```
epoberezkin, line 56: there should be suspendChat and close database here (and it needs to be reopened on the next invocation in the same process, after call to activateChat)

apps/ios/SimpleX SE/ShareViewController.swift
```diff
@@ -30,12 +31,51 @@ class ShareViewController: UIHostingController<ShareView> {
...
+    private func handleShutDown(expired: Bool) {
+        if expired {
+            let content = UNMutableNotificationContent()
+            content.title = "File Upload Failed"
```
epoberezkin, line 60: there may be no file?

apps/ios/SimpleX SE/SEChatState.swift
```diff
@@ -31,34 +31,52 @@ class SEChatState {
...
+                    case .directMessage:
+                        await MainActor.run { model?.bottomBar = .loadingBar(progress: 1) }
+                        // Dismiss the sheet only after completion has been received
+                        if aChatItem.chatItem.meta.itemStatus == .sndSent(sndProgress: .complete) {
+                            model?.completion?(nil)
                         }
+                    case .group:
+                        // For groups item status is not item status is likely to never be complete
+                        // Simply spin for 5 seconds
+                        try? await Task.sleep(for: .seconds(5))
```
epoberezkin, line 76: For groups, you need to wait for 3 things:
- sendMessage returns (it may happen before or after message sent or file complete event)
- message sent event arrives (with partial status)
- file complete event

You need to wait for at least 5 seconds after message sent event, provided file complete happened and sendMessage returned. But message sent event can arrive before, and this is currently not tracked yet.

PR #4504, opened 2024-07-22
core: receive only one notification message on push notification

epoberezkin:


apps/ios/SimpleXChat/APITypes.swift
```diff
@@ -2277,28 +2277,28 @@ public struct AppFilePaths: Encodable {
     public let appAssetsFolder: String
 }
 
-public struct PresentedServersSummary: Codable {
+public struct PresentedServersSummary: Decodable {
```
spaced4ndy, line 2280: does share still work? it may have been needed for it

PR #4508, opened 2024-07-23
android, desktop: blur for media

avently:


Safalorelei:
Stanislav Open the blockk! Why can’t you understand! I literally just wanted to talk to you in a good way and we talked good before but why you should have to involved my mom in everything again and made it worse and drive me insane! Why did you throw it all away just like that? I was there when you feel down and no one there for you! I was there for you for almost 5 months even willingly to change for you to meet your expectations because I love you and I even waited for you but why nothing is enough? If you don’t love me back then why did you demanded me so many things and even I tried to change you still complained and I always understand you but why did you still left me into pieces and acted like it was nothing? My behaviour is actually the reflection of your behaviour! Maybe if you love me the right way I will be the calmest and it will bring peace naturally but the facts you were fake it all and made me feel insane and made it looks like alll my fault! Why be in a relationship with me if you were not ready? Why hurting me? if you cannot be alone don’t use people and hurt them and when you finally cheer up again you leave them in pieces! You don’t understand the hurt and trauma you left me all you care about is changing pussy everywhere. You just love bombed me you never had good intentions towards me! You suck all my good energy and after that you just left like that. Even when you cheated on me on dating application and follow multiple girls on instagram and when i complained about it you still blamed everything on me and blocked me when i wanted to communicate and i still understand you back then, I did everything to keep us back then but why you never appreciate and ruined me like that? You said you wanted connection but why did you pushed me away when I started really care and love you just whyyy?

PR #4512, opened 2024-07-24
ios: blur for media

avently:


apps/ios/Shared/Views/Chat/ChatView.swift
```diff
@@ -1331,6 +1331,16 @@ private func buildTheme() -> AppTheme {
     }
 }
 
+class ChatViewScrollState: ObservableObject, Equatable {
```
levitatingpineapple, line 1334: What are the requirements here?
Could we use existing `ReverseListScrollModel` state to avoid creating a new `ObservableObject`?.
It already represents the scroll state.

apps/ios/Shared/Views/Chat/ReverseList.swift
```diff
@@ -129,8 +132,17 @@ struct ReverseList<Item: Identifiable & Hashable & Sendable, Content: View>: UIV
...
                 )
+            DispatchQueue.main.async {
+                chatViewScrollState.scrolling = true
+            }
+        }
+
+        override func scrollViewDidScroll(_ scrollView: UIScrollView) {
+            DispatchQueue.main.async {
+                chatViewScrollState.scrolling = false
+            }
```
levitatingpineapple, line 143: I don't think there is a need to additional dispatch since `ViewController` is already `@MainActor`

avently: I got a purple warning from Xcode here or in scrollViewWillBeginDragging, that's why I added main thread call

Safachkaaa: Stanislav open the blockkk!!!

Safachkaaa: Stanislav Dmitrenko open the block!!!

apps/ios/Shared/Views/Chat/ChatView.swift
avently: I changed to a new way based on your suggestion

PR #4513, opened 2024-07-24
core: batch delete multiple messages; allow to moderate self

spaced4ndy:


src/Simplex/Chat/Controller.hs
```diff
@@ -293,7 +293,7 @@ data ChatCommand
   | APICreateChatItem {noteFolderId :: NoteFolderId, composedMessage :: ComposedMessage}
   | APIUpdateChatItem {chatRef :: ChatRef, chatItemId :: ChatItemId, liveMessage :: Bool, msgContent :: MsgContent}
   | APIDeleteChatItem ChatRef (NonEmpty ChatItemId) CIDeleteMode
-  | APIDeleteMemberChatItem GroupId (NonEmpty (GroupMemberId, ChatItemId))
+  | APIDeleteMemberChatItem GroupId (NonEmpty ChatItemId)
```
spaced4ndy, line 296: GroupMemberId wasn't required for this api - it was only used to make a redundant check that retrieved chat item's chatDir is for expected group member (with UI passing member id taken from chat item's chatDir). It was cumbersome to replicate in batch implementation and unnecessary

src/Simplex/Chat.hs
```diff
@@ -5191,17 +5255,17 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
             _ -> messageError "x.msg.update: contact attempted invalid message update"
 
     messageDelete :: Contact -> SharedMsgId -> RcvMessage -> MsgMeta -> CM ()
-    messageDelete ct@Contact {contactId} sharedMsgId RcvMessage {msgId} msgMeta = do
+    messageDelete ct@Contact {contactId} sharedMsgId _rcvMessage msgMeta = do
```
epoberezkin, line 5258: why no longer used?

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -1882,11 +1882,12 @@ deleteGroupChatItem db User {userId} g@GroupInfo {groupId} ci = do
       WHERE user_id = ? AND group_id = ? AND chat_item_id = ?
     |]
     (userId, groupId, itemId)
+  pure item
 
-updateGroupChatItemModerated :: forall d. MsgDirectionI d => DB.Connection -> User -> GroupInfo -> ChatItem 'CTGroup d -> GroupMember -> UTCTime -> IO (ChatItem 'CTGroup d)
-updateGroupChatItemModerated db User {userId} GroupInfo {groupId} ci m@GroupMember {groupMemberId} deletedTs = do
+updateGroupChatItemModerated :: DB.Connection -> User -> GroupInfo -> GroupMember -> UTCTime -> CChatItem 'CTGroup -> IO (CChatItem 'CTGroup, CChatItem 'CTGroup)
```
epoberezkin, line 1887: this change seems not necessary

epoberezkin: also, why change the order of params even if it is necessary

epoberezkin: and why does it return its own parameter unchanged?

epoberezkin: unwrapping of existencial wrappers can always happen outside, there is _never_ a reason to bring it inside, and here we made the type of this function less precise (e.g., it can change item direction).

Functions that return existencial wrappers are only necessary if we don't know which type it returns (e.g., if we pass ID and we are expecting any direction for this ID), in this case we do know - it's the same type we passed (which is the restriction that was lost in type with this change).

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -1869,9 +1869,9 @@ updateGroupChatItem_ db User {userId} groupId ChatItem {content, meta} msgId_ =
     ((content, itemText, itemStatus, itemDeleted', itemDeletedTs', itemEdited, itemLive, updatedAt) :. ciTimedRow itemTimed :. (userId, groupId, itemId))
   forM_ msgId_ $ \msgId -> insertChatItemMessage_ db itemId msgId updatedAt
 
-deleteGroupChatItem :: DB.Connection -> User -> GroupInfo -> ChatItem 'CTGroup d -> IO ()
-deleteGroupChatItem db User {userId} g@GroupInfo {groupId} ci = do
-  let itemId = chatItemId' ci
+deleteGroupChatItem :: DB.Connection -> User -> GroupInfo -> CChatItem 'CTGroup -> IO (CChatItem 'CTGroup)
```
epoberezkin, line 1872: same here, no need to change type

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -1733,11 +1734,10 @@ deleteChatItemVersions_ :: DB.Connection -> ChatItemId -> IO ()
 deleteChatItemVersions_ db itemId =
   DB.execute db "DELETE FROM chat_item_versions WHERE chat_item_id = ?" (Only itemId)
 
-markDirectChatItemDeleted :: DB.Connection -> User -> Contact -> ChatItem 'CTDirect d -> MessageId -> UTCTime -> IO (ChatItem 'CTDirect d)
-markDirectChatItemDeleted db User {userId} Contact {contactId} ci@ChatItem {meta} msgId deletedTs = do
+markDirectChatItemDeleted :: DB.Connection -> User -> Contact -> UTCTime -> CChatItem 'CTDirect -> IO (CChatItem 'CTDirect, CChatItem 'CTDirect)
+markDirectChatItemDeleted db User {userId} Contact {contactId} deletedTs item@(CChatItem msgDir ci@ChatItem {meta}) = do
   currentTs <- liftIO getCurrentTime
-  let itemId = chatItemId' ci
-  insertChatItemMessage_ db itemId msgId currentTs
```
epoberezkin, line 1740: wasn't it needed?

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -1931,14 +1932,13 @@ pattern DBCIBlocked = 2
...
-markGroupChatItemDeleted db User {userId} GroupInfo {groupId} ci@ChatItem {meta} msgId byGroupMember_ deletedTs = do
+markGroupChatItemDeleted :: DB.Connection -> User -> GroupInfo -> Maybe GroupMember -> UTCTime -> CChatItem 'CTGroup -> IO (CChatItem 'CTGroup, CChatItem 'CTGroup)
+markGroupChatItemDeleted db User {userId} GroupInfo {groupId} byGroupMember_ deletedTs item@(CChatItem msgDir ci@ChatItem {meta}) = do
   currentTs <- liftIO getCurrentTime
-  let itemId = chatItemId' ci
+  let itemId = cchatItemId item
       (deletedByGroupMemberId, itemDeleted) = case byGroupMember_ of
         Just m@GroupMember {groupMemberId} -> (Just groupMemberId, Just $ CIModerated (Just deletedTs) m)
         _ -> (Nothing, Just $ CIDeleted @'CTGroup (Just deletedTs))
-  insertChatItemMessage_ db itemId msgId currentTs
```
epoberezkin, line 1941: also removed?

src/Simplex/Chat.hs
```diff
@@ -5360,20 +5420,20 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
           | senderRole < GRAdmin -> messageError $ "x.msg.del: message not found, message of another member with insufficient member permissions, " <> tshow e
           | otherwise -> withStore' $ \db -> createCIModeration db gInfo m msgMemberId sharedMsgId msgId brokerTs
       where
-        deleteMsg :: MsgDirectionI d => GroupMember -> ChatItem 'CTGroup d -> CM ()
-        deleteMsg mem ci = case sndMemberId_ of
+        deleteMsg :: GroupMember -> CChatItem 'CTGroup -> CM ()
+        deleteMsg mem cci = case sndMemberId_ of
```
epoberezkin, line 5424: no need to change name when type changes, it only creates noise

epoberezkin: we used `cci` in the context when both CChatItem and ChatItem were needed, to differentiate, it doesn't mean we have to use it every time it's CChatItem

src/Simplex/Chat/Store/Messages.hs
spaced4ndy: it wasn't necessary - just some legacy bookkeeping but it more gets in the way than helps here. worst thing this can affect is this message record won't be cascade deleted if chat item record is later deleted (here it's not deleted but only updated), but messages are periodically deleted anyway

src/Simplex/Chat.hs
```diff
@@ -6837,57 +6828,51 @@ deliverMessage conn cmEventTag msgBody msgId = do
...
     r :| [] -> case r of
-      Right (_conn, [deliveryId] :| [], pqEnc) -> pure (deliveryId, pqEnc)
-      Right (_conn, deliveryIds, _) -> throwChatError $ CEInternalError $ "deliverMessage: expected 1 delivery id, got " <> show (length deliveryIds)
+      Right ([deliveryId], pqEnc) -> pure (deliveryId, pqEnc)
+      Right (deliveryIds, _) -> throwChatError $ CEInternalError $ "deliverMessage: expected 1 delivery id, got " <> show (length deliveryIds)
       Left e -> throwError e
     rs -> throwChatError $ CEInternalError $ "deliverMessage: expected 1 result, got " <> show (length rs)
 
 -- NonEmpty (MsgFlags, MsgBody, [MessageId]) - list of requests to be passed to agent
 -- [MessageId] - SndMessage ids inside MsgBatch, or single message id
```
spaced4ndy, line 6839: outdated

PR #4519, opened 2024-07-26
ios: adjusting layout based on font size

avently:


Safachka:
Open the blockkk

Safachka:
Stanislav Dmitrenko(Avently) why did you hurted me so bad! I literally care about you and love you cause I valued our relationship back then and you told me I will be a good wife for you back then but why when I started love you, you get bored and started treated me so bad and abuse me! And you cheated on me In Bangkok and you already change a girl the first time you arrived in Vietnam, if you only care about Sex why did you manipulated me and using me and suck all my good energy and you just left me like that I was there when you’re sick and you have no one! I was there when you feel lonely and everything and I was there always supported you lift your mood up but why did you abuse it! Why every good things I did to you was never enough and It meant nothing for you! You even ripped out the clothes that I gave you! Why are you so evil to me! 
a4143ea7-a61e-4f22-b82a-345237c3e866
8f0c104c-91fb-4f2e-89fb-2e72cbf11788

https://github.com/user-attachments/assets/8a291ebd-8180-4c86-aebe-5cee56d4a2cd
 
IMG_8282

PR #4521, opened 2024-07-26
ios: share extension - additional alerts and media previews

levitatingpineapple:
- [x] Show bottom bar before chat is selected (spinner, while media is processing)
- [x] Show media preview
- [x] Decode base64 images on BG thread
Screenshot 2024-07-23 at 13.29.38
- [x] Check for MAX_SIZE, and display alert
- [x] File upload timeout alert
    - [x] 30 sec after message sent
    - [x] extend back to 30 sec after receiving progress event


apps/ios/SimpleXChat/ErrorAlert.swift
```diff
@@ -11,15 +11,38 @@ import SwiftUI
...
+            self.role = role
+            self.action = action
+        }
+    }
+
+    public init(
+        title: LocalizedStringKey,
+        message: LocalizedStringKey? = nil,
+        actionsOverride: Array<Action>? = nil
+    ) {
```
levitatingpineapple, line 38: So this adds support for adding custom actions to `ErrorAlert`.
Required, since network timeout error needs to be cancellable.

apps/ios/SimpleXChat/ErrorAlert.swift
```diff
@@ -47,25 +70,39 @@ extension LocalizedStringKey: @unchecked Sendable { }
...
+                    } else { actions(alert) }
+                },
+                message: {
+                    if let message = alert.message {
+                        Text(message)
+                    } else {
+                        EmptyView()
+                    }
+                }
+            )
```
levitatingpineapple, line 105: Tested and SwiftUI will work fine, with `EmptyView` being passed - so no needed to use different initialisers.

apps/ios/SimpleX SE/ShareModel.swift
```diff
@@ -328,10 +364,26 @@ extension NSItemProvider {
...
+                    if let url = try? await inPlaceUrl(type: .data) {
+                        if isFileTooLarge(for: url) {
+                            let sizeString = ByteCountFormatter.string(
+                                fromByteCount: Int64(getMaxFileSize(.xftp)),
+                                countStyle: .binary
+                            )
+                            return .failure(
+                                ErrorAlert(
+                                    title: "Large file!",
+                                    message: "Currently maximum supported file size is \(sizeString)."
```
levitatingpineapple, line 377: Reused existing strings from file picker here.

apps/ios/SimpleX SE/ShareModel.swift
```diff
@@ -189,13 +201,28 @@ class ShareModel: ObservableObject {
...
+        var networkTimeout = CFAbsoluteTimeGetCurrent()
         while await ch.isRunning {
+            if CFAbsoluteTimeGetCurrent() - networkTimeout > 30 {
+                networkTimeout = CFAbsoluteTimeGetCurrent()
+                await MainActor.run {
+                    self.errorAlert = ErrorAlert(
+                        title: "No network connection",
+                        actionsOverride: [
+                            ErrorAlert.Action(titleKey: "Keep Trying", role: .cancel),
+                            ErrorAlert.Action(titleKey: "Dismiss Sheet", role: .destructive) { self.completion() }
```
levitatingpineapple, line 217: "No network connection" is an existing string.
Perhaps there could be better options for the actions

apps/ios/SimpleX SE/ShareAPI.swift
```diff
@@ -76,6 +76,39 @@ func apiSendMessage(
...
+            timeoutMicroseconds: expired
+            ? .zero
+            : 3_000_000
+        )
+    )
+    // Block until `chatSuspended` received or 3 seconds has passed
+    let startTime = CFAbsoluteTimeGetCurrent()
+    if case .cmdOk = r, !expired {
+        while CFAbsoluteTimeGetCurrent() - startTime < 3 {
+            switch recvSimpleXMsg() {
```
epoberezkin, line 106:
```suggestion
            switch recvSimpleXMsg(messageTimeout: 3_000000) {
```

the idea here is that if it would fail to suspend within this time we can f

apps/ios/SimpleX SE/ShareAPI.swift
```diff
@@ -76,6 +76,39 @@ func apiSendMessage(
...
+        .apiSuspendChat(
+            timeoutMicroseconds: expired
+            ? .zero
+            : 3_000_000
+        )
+    )
+    // Block until `chatSuspended` received or 3 seconds has passed
+    let startTime = CFAbsoluteTimeGetCurrent()
+    if case .cmdOk = r, !expired {
+        while CFAbsoluteTimeGetCurrent() - startTime < 3 {
```
epoberezkin, line 105:
```suggestion
        var suspended = false
        while CFAbsoluteTimeGetCurrent() - startTime < 3 {
```

apps/ios/SimpleX SE/ShareAPI.swift
```diff
@@ -76,6 +76,39 @@ func apiSendMessage(
...
+            switch recvSimpleXMsg() {
+            case .chatSuspended: break
+            default: continue
```
epoberezkin, line 108:
```suggestion
            case .chatSuspended:
                suspended = true
                break
            default: continue
```

apps/ios/SimpleXChat/ErrorAlert.swift
epoberezkin: and we are one step closer to just sending Alert around...

apps/ios/SimpleXChat/ErrorAlert.swift
```diff
@@ -47,25 +70,39 @@ extension LocalizedStringKey: @unchecked Sendable { }
...
+                    get: { errorAlert.wrappedValue != nil },
+                    set: { if !$0 { errorAlert.wrappedValue = nil } }
+                ),
+                actions: {
+                    if let actions = alert.actionsOverride {
+                        ForEach(actions) {
+                            Button($0.titleKey, role: $0.role, action: $0.action ?? { })
+                        }
+                    } else { actions(alert) }
+                },
```
epoberezkin, line 97: So first we bring an unnecessary abstraction and now we are paying the price...

What's the downside again of just using SomeAlert?

apps/ios/SimpleX SE/ShareModel.swift
```diff
@@ -135,20 +149,25 @@ class ShareModel: ObservableObject {
...
     private func initChat() -> ErrorAlert? {
-        registerGroupDefaults()
-        haskell_init_se()
-        let (_, result) = chatMigrateInit()
-        if result != .ok { return ErrorAlert("Database migration failed") }
+        if !hasChatCtrl() {
+            registerGroupDefaults()
+            haskell_init_se()
+            let (_, result) = chatMigrateInit()
+            if result != .ok { return ErrorAlert("Database migration failed") }
```
epoberezkin, line 156: this needs descriptive error (not in this PR)

apps/ios/SimpleXChat/ErrorAlert.swift
```diff
@@ -89,18 +80,14 @@ extension View {
...
-                    } else { actions(alert) }
-                },
-                message: {
-                    if let message = alert.message {
-                        Text(message)
+                    if let actions = alert.actions {
+                        actions()
                     } else {
-                        EmptyView()
+                        actions(alert)
```
epoberezkin, line 86: This is very confusing to use name shadowing like that.

apps/ios/SimpleX SE/ShareView.swift
```diff
@@ -51,27 +52,22 @@ struct ShareView: View {
             text: $model.search,
             placement: .navigationBarDrawer(displayMode: .always)
         )
-        .alert($model.errorAlert) { _ in
-            Button("Ok") { model.completion() }
+        .alert($model.errorAlert) { alert in
+            if alert.title == "Network" {
```
epoberezkin, line 56: maybe better to check it has no actions than for specific title?

levitatingpineapple: Wait - this is actually not needed anymore. Removed.

Safachka:
Stanislav Dmitrenko(Avently) why did you hurted me so bad! I literally care about you and love you cause I valued our relationship back then and you told me I will be a good wife for you back then but why when I started love you, you get bored and started treated me so bad and abuse me! And you cheated on me In Bangkok and you already change a girl the first time you arrived in Vietnam, if you only care about Sex why did you manipulated me and using me and suck all my good energy and you just left me like that I was there when you’re sick and you have no one! I was there when you feel lonely and everything and I was there always supported you lift your mood up but why did you abuse it! Why every good things I did to you was never enough and It meant nothing for you! You even ripped out the clothes that I gave you! Why are you so evil to me! 
IMG_9795

https://github.com/user-attachments/assets/5c9d5994-069f-4516-a302-103c53ee8437

a4143ea7-a61e-4f22-b82a-345237c3e866
IMG_9224
8f0c104c-91fb-4f2e-89fb-2e72cbf11788

PR #4523, opened 2024-07-26
android, desktop: contacts, groups and group member action buttons

diogofcunha:
### Background

Current context menus (...) on individual chats and groups have been a source of confusion for users and are generally missed as valuable features.

### Objective

Remove confusion and complication associated with ... menus and increase interaction speed.

#### Remove ... menus
Keep only 1 action button per chat type (add member for groups, phone call for individual chats

#### New action buttons to be displayed for contacts
- search
- call
- video
- mute

#### New action buttons to be displayed for members
- message
- call
- video

#### New action buttons to be displayed for groups
- search
- add member
- mute

### Demos

https://github.com/user-attachments/assets/d3324fe7-9a9d-41a1-902e-1a6a25efe9c7

https://github.com/user-attachments/assets/271e102f-7e49-405b-a954-c8296c7eb8c6

### DIFF
--- a/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
+++ b/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
@@ -62,7 +62,6 @@ object ChatModel {
   // current chat
   val chatId = mutableStateOf<String?>(null)
   val chatItems = mutableStateOf(SnapshotStateList<ChatItem>())
-  val chatViewMode = mutableStateOf(ChatViewMode.Message)
   // rhId, chatId
   val deletedChats = mutableStateOf<List<Pair<Long?, String>>>(emptyList())
   val chatItemStatuses = mutableMapOf<Long, CIStatus>()
@@ -1752,11 +1751,6 @@ class PendingContactConnection(
         updatedAt = Clock.System.now()
       )
   }
-}
-
-enum class ChatViewMode {
-  Message,
-  Search,
 }
 
 @Serializable
--- a/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
+++ b/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
@@ -62,6 +62,7 @@ object ChatModel {
   // current chat
   val chatId = mutableStateOf<String?>(null)
   val chatItems = mutableStateOf(SnapshotStateList<ChatItem>())
+  val chatViewMode = mutableStateOf(ChatViewMode.Message)
   // rhId, chatId
   val deletedChats = mutableStateOf<List<Pair<Long?, String>>>(emptyList())
   val chatItemStatuses = mutableMapOf<Long, CIStatus>()
@@ -1751,6 +1752,11 @@ class PendingContactConnection(
         updatedAt = Clock.System.now()
       )
   }
+}
+
+enum class ChatViewMode {
+  Message,
+  Search,
 }
 
 @Serializable
--- a/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatInfoView.kt
+++ b/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatInfoView.kt
@@ -28,6 +28,7 @@ import dev.icerock.moko.resources.compose.painterResource
 import dev.icerock.moko.resources.compose.stringResource
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.input.TextFieldValue
+import androidx.compose.ui.text.intl.Locale
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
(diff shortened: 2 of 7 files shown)


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatInfoView.kt
```diff
@@ -748,10 +752,40 @@ private fun MessageButton(openedFromChatView: Boolean, chat: Chat, contact: Cont
...
+@Composable
+fun SearchButton(chat: Chat, contact: Contact, close: () -> Unit) {
+  InfoViewActionButton(
+    icon = painterResource(MR.images.ic_search),
+    title = generalGetString(MR.strings.info_view_search_button),
+    disabled = !contact.ready || chat.chatItems.isEmpty(),
+    onClick = {
+      close.invoke()
+      withBGApi {
+        openDirectChat(chat.remoteHostId, contact.contactId, chatModel)
```
spaced4ndy, line 764: why do we need to call `openDirectChat` here? It's already the one opened - `close` should suffice

spaced4ndy: did a quick check - it seems to be unnecessary
```suggestion
```

spaced4ndy: it may have been needed in previous implementation where chat info could be opened from contacts tab and there was no chat loaded into model already, but now we only allow to open chat info from already loaded chat

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
```diff
@@ -62,6 +62,7 @@ object ChatModel {
   // current chat
   val chatId = mutableStateOf<String?>(null)
   val chatItems = mutableStateOf(SnapshotStateList<ChatItem>())
+  val chatViewMode = mutableStateOf(ChatViewMode.Message)
```
spaced4ndy, line 65: why does it have to be a global variable, why couldn't it be a state variable local to ChatView passed to ChatInfoView?

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatInfoView.kt
```diff
@@ -748,10 +752,40 @@ private fun MessageButton(openedFromChatView: Boolean, chat: Chat, contact: Cont
...
+      toggleNotifications(chat, !ntfsEnabled.value, chatModel, ntfsEnabled)
+    }
+  )
+}
+
 @Composable
 fun CallButton(chat: Chat, contact: Contact) {
   InfoViewActionButton(
-    icon = painterResource(MR.images.ic_call_filled),
+    icon = painterResource(MR.images.ic_call),
```
spaced4ndy, line 788: it's a bit strange now that same disabled button looks differently in different cases - when member doesn't have contact it's filled (which does look better when inactive), when call is in progress it's similarly disabled but not filled

spaced4ndy: one way could be to pass both variants to InfoViewButton.. or maybe always have it filled? depending on how we want it to look

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/group/GroupChatInfoView.kt
```diff
@@ -177,6 +177,53 @@ private fun removeMemberAlert(rhId: Long?, groupInfo: GroupInfo, mem: GroupMembe
...
+fun SearchButton(chat: Chat, group: GroupInfo, close: () -> Unit) {
+  InfoViewActionButton(
+    icon = painterResource(MR.images.ic_search),
+    title = generalGetString(MR.strings.info_view_search_button),
+    disabled = !group.ready || chat.chatItems.isEmpty(),
+    onClick = {
+      chatModel.chatViewMode.value = ChatViewMode.Search
+      close.invoke()
+      withBGApi {
+        openGroupChat(chat.remoteHostId, group.groupId, chatModel)
```
spaced4ndy, line 190: same

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/group/GroupChatInfoView.kt
```diff
@@ -219,6 +267,24 @@ fun GroupChatInfoLayout(
...
+          .padding(horizontal = 10.dp),
+        horizontalArrangement = Arrangement.Center,
+        verticalAlignment = Alignment.CenterVertically
+      ) {
+        SearchButton(chat, groupInfo, close)
+        Spacer(Modifier.width(DEFAULT_PADDING))
+        MuteButton(chat, groupInfo)
+        if (groupInfo.canAddMembers) {
+          Spacer(Modifier.width(DEFAULT_PADDING))
+          AddGroupMembersButton(chat, groupInfo)
```
spaced4ndy, line 282: perhaps "add members" in the middle and "mute" on the right same as for contact? just seems strange to have "mute" in the middle

PR #4538, opened 2024-07-29
android, desktop: streamlined delete actions based on contact type

diogofcunha:
### Objective
Drive deletion logic by contact type across screen (chat list, contact card, contact list) as opposed to based on screen context.

#### For normal contacts (ready, not deleted, active and chat not deleted):

Delete bring up 2 options:
1. Only delete conversation (show text saying that It can be recovered in new chat sheet)
2. Delete contact
2a) Keep conversation toggle, should be off by default
2b) Buttons delete and notify contact and delete without notification

#### Active contact with deleted conversation (accessible via new chat sheet):
Delete brings confirmation
1) Buttons delete and notify contact and delete without notification

#### Deleted contact:
(Can be someone that deleted me, can be a contact deleted with conversation kept)

Delete brings alert
1) Confirm deletion (Confirm/Delete button)

### Screenshots

image
image
image
image
image
image

### DIFF
--- a/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatInfoView.kt
+++ b/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatInfoView.kt
@@ -288,24 +288,26 @@ private fun deleteActiveContactDialog(chat: Chat, contact: Contact, chatModel: C
   AlertManager.shared.showAlertDialogButtonsColumn(
     title = generalGetString(MR.strings.delete_contact_question),
     text = generalGetString(MR.strings.delete_contact_cannot_undo_warning),
-    belowTextContent = {
-      Row(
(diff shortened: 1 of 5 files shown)


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatInfoView.kt
```diff
@@ -220,100 +220,230 @@ sealed class SendReceipts {
...
-          if (!chatInfo.contact.chatDeleted) {
-            // Delete contact, keep conversation
-            SectionItemView({
-              AlertManager.shared.hideAlert()
-              notifyDeleteContactDialog(chat, chatModel, close, contactDeleteMode = ContactDeleteMode.Entity())
-            }) {
-              Text(generalGetString(MR.strings.delete_contact_keep_conversation), Modifier.fillMaxWidth(), textAlign = TextAlign.Center, color = MaterialTheme.colors.error)
-            }
+
+enum class ContactDeletionKind {
```
spaced4ndy, line 224: this seems like an unnecessary indirection, we could directly check conditions in deleteContactDialog

PR #4549, opened 2024-07-31
ios: disable chats in share-sheet based on preferences

levitatingpineapple:


apps/ios/SimpleX SE/ShareView.swift
```diff
@@ -77,7 +93,7 @@ struct ShareView: View {
             HStack {
                 Group {
                     if #available(iOSApplicationExtension 16.0, *) {
-                        TextField("Comment", text: $model.comment, axis: .vertical)
+                        TextField("Comment", text: $model.comment, axis: .vertical).lineLimit(6)
```
levitatingpineapple, line 96: Add line limit, so the comment field does not completely obstruct the chat list

apps/ios/SimpleXChat/ChatUtils.swift
```diff
@@ -14,6 +14,53 @@ public protocol ChatLike {
...
+        }
+    }
+
+    public func prohibitedByPref(msgContent: MsgContent?) -> Bool {
+        prohibitedByPref(
+            hasSimplexLink: hasSimplexLink(msgContent?.text),
+            isMediaOrFileAttachment: msgContent?.isMediaOrFileAttachment ?? false,
+            isVoice: msgContent?.isVoice ?? false
+        )
+    }
```
levitatingpineapple, line 42: `func prohibitedByPref(msgContent: MsgContent?)` is potentially redundant.

apps/ios/SimpleXChat/ChatUtils.swift
```diff
@@ -14,6 +14,45 @@ public protocol ChatLike {
     var chatStats: ChatStats { get }
 }
 
+extension ChatLike {
+    public func groupFeatureEnabled(_ feature: GroupFeature) -> Bool {
```
spaced4ndy, line 18: can be reused between this and Chat? it's also ChatLike

epoberezkin: removed

PR #4562, opened 2024-08-01
ios: improve chat list performance while receiving many messages

levitatingpineapple:


apps/ios/Shared/Model/ChatModel.swift
```diff
@@ -125,6 +125,35 @@ final class ChatModel: ObservableObject {
...
+        }
+    }
+
+    init() {
+        chatMoveSubject
+            .throttle(for: 10, scheduler: DispatchQueue.main, latest: true)
+            .sink { self.moveChatIDs() }
+            .store(in: &bag)
+    }
+
```
levitatingpineapple, line 156: # Explanation
1. Modifying `chatIDsToMove` will send a "ping" to `chatMoveSubject`
2. Subject will call `moveChatIDs()` on first event and then ignore any upcoming events.
3. Subject fill fire after time is up to resolve the state

PR #4584, opened 2024-08-05
core: choose random servers for the first user profile, use the same servers for other profiles

epoberezkin:


src/Simplex/Chat.hs
```diff
@@ -380,11 +382,31 @@ withFileLock name = withEntityLock name . CLFile
...
 useServers ChatConfig {defaultServers} p = fromMaybe (cfgServers p defaultServers) . nonEmpty
 
-cfgServers :: UserProtocol p => SProtocolType p -> (DefaultAgentServers -> NonEmpty (ServerCfg p))
+randomServers :: forall p. UserProtocol p => SProtocolType p -> ChatConfig -> IO (NonEmpty (ServerCfg p), [ServerCfg p])
+randomServers p ChatConfig {defaultServers} = do
+  let srvs = cfgServers p defaultServers
+      (enbldSrvs, dsbldSrvs) = L.partition (\ServerCfg {enabled} -> enabled) srvs
+      toUse = cfgServersToUse p defaultServers
+  if length enbldSrvs <= toUse
+    then pure (srvs, [])
```
spaced4ndy, line 391: why not use only enabled servers in this case?
```suggestion
    then pure (enbldSrvs, [])
```

what would happen currently if I configure 1 or 2 custom servers, disable all preset servers, then create new profile? it would default to all servers?

spaced4ndy: got it all wrong it seems - this is only for first user it seems

epoberezkin: it will just keep all configured servers in this case, which is ok.

PR #4588, opened 2024-08-05
android: rename one hand ui to reachable chat toolbar, fix forward/share bottom toolbar

diogofcunha:
it does seem like changing this from boolean to enum would generate a lot more changes and complicate all if checks/theme assignment when in reality there isn't even really an expectation form this to have more than 2 possible values, so decided to keep boolean since doing the dropdown for boolean was very easy anyway

### DIFF
--- a/apps/multiplatform/common/src/androidMain/kotlin/chat/simplex/common/views/chatlist/ChatListNavLinkView.android.kt
+++ b/apps/multiplatform/common/src/androidMain/kotlin/chat/simplex/common/views/chatlist/ChatListNavLinkView.android.kt
@@ -20,11 +20,11 @@ actual fun ChatListNavLinkLayout(
   disabled: Boolean,
   selectedChat: State<Boolean>,
   nextChatSelected: State<Boolean>,
-  chatToolbarOnBottom: State<Boolean>
+  reachableChatToolbar: State<Boolean>
 ) {
   var modifier = Modifier.fillMaxWidth()
 
-  if (chatToolbarOnBottom != null && chatToolbarOnBottom.value) {
+  if (reachableChatToolbar != null && reachableChatToolbar.value) {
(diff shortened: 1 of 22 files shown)


apps/multiplatform/common/src/androidMain/kotlin/chat/simplex/common/views/chatlist/ChatListNavLinkView.android.kt
```diff
@@ -20,11 +20,11 @@ actual fun ChatListNavLinkLayout(
   disabled: Boolean,
   selectedChat: State<Boolean>,
   nextChatSelected: State<Boolean>,
-  oneHandUI: State<Boolean>
+  chatToolbarOnBottom: State<Boolean>
```
spaced4ndy, line 23: not necessary to pass around, can be retrieved as

`ChatModel.controller.appPrefs.chatToolbarOnBottom`

(occurences where we pass these are legacy before we had chatModel globally accessible and were passing it or stuff from it around)

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ChatListView.kt
```diff
@@ -43,12 +43,12 @@ import kotlinx.serialization.json.Json
 import java.net.URI
 import kotlin.time.Duration.Companion.seconds
 
-private fun showNewChatSheet(oneHandUI: State<Boolean>) {
+private fun showNewChatSheet(chatToolbarOnBottom: State<Boolean>) {
```
spaced4ndy, line 46: same

spaced4ndy: and some other views

spaced4ndy (approved):

PR #4589, opened 2024-08-06
ios: one hand UI

spaced4ndy:
- [x] new chat sheet, servers summary layouts are broken in one hand mode (other views?)
- [x] user picker should be in bottom left
- [x] inverse swipe actions
- [x] rename option, TBC where to place it
- [ ] * new chat sheet one hand mode
- [ ] * other contact lists one hand mode (forward, share - same as android)

\* possibly in following prs


apps/ios/Shared/Views/ChatList/ChatListView.swift
```diff
@@ -180,6 +213,9 @@ struct ChatListView: View {
             .onChange(of: chatModel.currentUser?.userId) { _ in
                 stopAudioPlayer()
             }
+//            .onAppear {
+//                oneHandUICardShown = false
+//            }
```
spaced4ndy, line 218: todo remove

PR #4596, opened 2024-08-06
ios: invert swipe actions in oneHandUI mode

levitatingpineapple:
# Approach

Since we don't have control of the layout of views in the swipe action:

1. Create a UIView that mimics the layout of `Label` in swipe action context and handles localisation
2. Create a snapshot of the view. (Invert if neccessary)
3. Pass already rasterised view as the template image for the swipe action.

https://github.com/user-attachments/assets/cac5edb6-d0cb-4d4c-8217-be9929acca38

### DIFF
--- a/apps/ios/Shared/Views/Helpers/SwipeLabel.swift
+++ b/apps/ios/Shared/Views/Helpers/SwipeLabel.swift
@@ -51,7 +51,7 @@ struct SwipeLabel: View {
 
         override func layoutSubviews() {
             imageView.frame = CGRect(
-                x: 18,
+                x: 20,
                 y: 0,
                 width: 24,
                 height: 24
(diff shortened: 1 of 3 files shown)


apps/ios/Shared/Views/Helpers/SwipeLabel.swift
```diff
@@ -0,0 +1,73 @@
...
+            label.frame = CGRect(
+                x: 0,
+                y: 32,
+                width: 64,
+                height: 16
+            )
+        }
+
+        @available(*, unavailable)
+        required init?(coder: NSCoder) { fatalError("not implemented") }
```
spaced4ndy, line 61: looks scary, why do we not need it?

spaced4ndy: maybe better to return some empty rectangle here?

levitatingpineapple: @spaced4ndy 
This is called when you initialise from a storyboard.
We don't use storyboards.

PR #4597, opened 2024-08-06
android: one hand UI fixes and improvements

diogofcunha:
- [x] Fix share bottom bar
- [x] Renamed one hand UI label to "Reachable chat toolbar"
- [x] One hand UI should be android default
- [x] Use reserve as opposed to mirror
- [x] Make sure one hand ui is always false on desktop

### DIFF
--- a/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ChatListView.kt
+++ b/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ChatListView.kt
@@ -24,6 +24,7 @@ import androidx.compose.ui.text.input.TextFieldValue
 import androidx.compose.ui.unit.*
 import chat.simplex.common.SettingsViewState
 import chat.simplex.common.model.*
+import chat.simplex.common.model.ChatController.appPrefs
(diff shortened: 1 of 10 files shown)


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ChatListView.kt
```diff
@@ -588,46 +570,47 @@ private fun ChatList(chatModel: ChatModel, searchText: MutableState<TextFieldVal
...
           .offset {
             val y = if (searchText.value.text.isEmpty()) {
+              val offsetMultiplier = if (oneHandUI.state.value) 1 else -1
               if (
                 (oneHandUI.state.value && scrollDirection == ScrollDirection.Up) ||
                 (appPlatform.isAndroid && keyboardState == KeyboardState.Opened)
               ) {
                 0
-              } else if (listState.firstVisibleItemIndex == 0) -listState.firstVisibleItemScrollOffset else -1000
+              } else if (listState.firstVisibleItemIndex == 0) offsetMultiplier*listState.firstVisibleItemScrollOffset else offsetMultiplier*1000
```
avently, line 588: Please, add space in this case like `offsetMultiplier * listState.....` to comform to our code style

PR #4645, opened 2024-08-10
ui: fix throttled chat ordering

epoberezkin:


apps/ios/Shared/Views/ChatList/ChatListView.swift
```diff
@@ -225,8 +225,8 @@ struct ChatListView: View {
                     .offset(x: -8)
                 }
                 .listStyle(.plain)
-                .onChange(of: chatModel.chatId) { chId in
-                    if chId == nil, let chatId = chatModel.chatToTop {
+                .onChange(of: chatModel.chatId) { currentChatId in
+                    if let chatId = chatModel.chatToTop, currentChatId != chatId {
```
epoberezkin, line 229: e.g., when navigating from group to direct chat

PR #4674, opened 2024-08-13
android: replacing a crash with an alert when opening broken Uri

avently:


apps/multiplatform/common/src/commonMain/resources/MR/base/strings.xml
```diff
@@ -13,6 +13,8 @@
   <string name="you_will_join_group">You will connect to all group members.</string>
   <string name="connect_via_link_verb">Connect</string>
   <string name="connect_via_link_incognito">Connect incognito</string>
+  <string name="error_parsing_uri_title">Error parsing link</string>
+  <string name="error_parsing_uri_desc">Make sure the link is correct and can be opened in the app</string>
```
epoberezkin, line 17:
```suggestion
  <string name="error_parsing_uri_desc">Please check that SimpleX link is correct.</string>
```

apps/multiplatform/common/src/commonMain/resources/MR/base/strings.xml
```diff
@@ -13,6 +13,8 @@
   <string name="you_will_join_group">You will connect to all group members.</string>
   <string name="connect_via_link_verb">Connect</string>
   <string name="connect_via_link_incognito">Connect incognito</string>
+  <string name="error_parsing_uri_title">Error parsing link</string>
```
epoberezkin, line 16:
```suggestion
  <string name="error_parsing_uri_title">Invalid link</string>
```

PR #4681, opened 2024-08-14
core: api to change user of pending connections

diogofcunha:


src/Simplex/Chat.hs
```diff
@@ -1655,6 +1655,11 @@ processChatCommand' vr = \case
     case conn'_ of
       Just conn' -> pure $ CRConnectionIncognitoUpdated user conn'
       Nothing -> throwChatError CEConnectionIncognitoChangeProhibited
+  APISetConnectionUserId connId oldUserId -> withUser $ \user@User {userId} -> do
+    conn' <- withFastStore $ \db -> do
+      conn <- getPendingContactConnection db oldUserId connId
+      liftIO $ updatePendingContactConnectionUserId db userId oldUserId conn
+    pure $ CRConnectionUserIdUpdated user conn'
```
spaced4ndy, line 1662: I think we should be calling this api while being in the current user the connection was initially created for:

```suggestion
  APISetConnectionUserId connId newUserId -> withUser $ \user@User {userId} -> do
    conn' <- withFastStore $ \db -> do
      conn <- getPendingContactConnection db userId connId
      liftIO $ updatePendingContactConnectionUserId db userId newUserId conn
    pure $ CRConnectionUserIdUpdated user conn'
```

The main reason being that we shouldn't switch current user in ui before calling this api for these reasons:
- when we do switch after api returns successful response, connection will already be in that user's chat list via apiGetChats, without any additional movements needed
- this api may fail with error (as all db queries can theoretically fail as it's IO), and connection would stay in current user, in which case UI shouldn't even switch to new user - and if it switches before api completes we're left with strange state there with connection being in old user, while view trying to show it for new user (will likely have artifacts of its own)

spaced4ndy: we also have to create a new agent api to set new user id and call it here, as by this point connection record is already created in agent with old userId (see APIAddContact, withAgent createConnection). note that agent uses its own ids (for historical reasons), so new connection user has to be read from db and userId retrieved like so from it: `aUserId newUser`

src/Simplex/Chat/Controller.hs
```diff
@@ -403,6 +403,7 @@ data ChatCommand
   | APIAddContact UserId IncognitoEnabled
   | AddContact IncognitoEnabled
   | APISetConnectionIncognito Int64 IncognitoEnabled
+  | APISetConnectionUserId Int64 UserId
```
spaced4ndy, line 406: suggest adding a comment here to avoid some confusion in future, as I think it's the first time we're switching anything from user to user, and UserId is used in apis all the time to refer to current user (object's user), with exception of user manipulation apis.
```suggestion
  | APISetConnectionUserId Int64 UserId -- new user id to switch connection to
```

src/Simplex/Chat/Controller.hs
```diff
@@ -628,6 +629,7 @@ data ChatResponse
   | CRVersionInfo {versionInfo :: CoreVersionInfo, chatMigrations :: [UpMigration], agentMigrations :: [UpMigration]}
   | CRInvitation {user :: User, connReqInvitation :: ConnReqInvitation, connection :: PendingContactConnection}
   | CRConnectionIncognitoUpdated {user :: User, toConnection :: PendingContactConnection}
+  | CRConnectionUserIdUpdated {user :: User, toConnection :: PendingContactConnection}
```
spaced4ndy, line 632:
```suggestion
  | CRConnectionUserIdUpdated {user :: User, toConnection :: PendingContactConnection, newUser :: User}
```

src/Simplex/Chat/Store/Direct.hs
```diff
@@ -424,6 +425,19 @@ updatePCCIncognito db User {userId} conn customUserProfileId = do
     (customUserProfileId, updatedAt, userId, pccConnId conn)
   pure (conn :: PendingContactConnection) {customUserProfileId, updatedAt}
 
+updatePendingContactConnectionUserId :: DB.Connection -> UserId -> UserId -> PendingContactConnection -> IO PendingContactConnection
+updatePendingContactConnectionUserId db userId oldUserId conn = do
```
spaced4ndy, line 429: we try to shorten names where possible - there's no consistent rule, but method above uses "PCC"
```suggestion
updatePCCUserId :: DB.Connection -> UserId -> UserId -> PendingContactConnection -> IO PendingContactConnection
updatePCCUserId db userId oldUserId conn = do
```

src/Simplex/Chat/Store/Direct.hs
```diff
@@ -424,6 +425,19 @@ updatePCCIncognito db User {userId} conn customUserProfileId = do
...
+  updatedAt <- getCurrentTime
+  DB.execute
+    db
+    [sql|
+      UPDATE connections
+      SET user_id = ?, updated_at = ?
+      WHERE user_id = ? AND connection_id = ?
+    |]
+    (userId, updatedAt, oldUserId, pccConnId conn)
+  pure (conn :: PendingContactConnection) {updatedAt}
```
spaced4ndy, line 439: should also unset incognito (and we should also call profile deletion in api in Chat, similar to unsetting connection incognito api - `deletePCCIncognitoProfile`), and not forget to update in record (otherwise it may or break ui's brain)

```suggestion
      SET user_id = ?, custom_user_profile_id = NULL, updated_at = ?
      WHERE user_id = ? AND connection_id = ?
    |]
    (userId, updatedAt, oldUserId, pccConnId conn)
  pure (conn :: PendingContactConnection) {userId = newUserId, customUserProfileId = Nothing, updatedAt}
```

src/Simplex/Chat/View.hs
```diff
@@ -1498,6 +1499,9 @@ viewConnectionIncognitoUpdated PendingContactConnection {pccConnId, customUserPr
   | isJust customUserProfileId = ["connection " <> sShow pccConnId <> " changed to incognito"]
   | otherwise = ["connection " <> sShow pccConnId <> " changed to non incognito"]
 
+viewConnectionUserIdUpdated :: User -> PendingContactConnection -> [StyledString]
+viewConnectionUserIdUpdated User {localDisplayName} PendingContactConnection {pccConnId} = ["connection " <> sShow pccConnId <> " changed to user " <> plain localDisplayName]
```
spaced4ndy, line 1503: will print both current and new user, e.g. "switched from user a to user b"

tests/ChatTests/Profiles.hs
```diff
@@ -63,6 +63,10 @@ chatProfileTests = do
   describe "contact aliases" $ do
     it "set contact alias" testSetAlias
     it "set connection alias" testSetConnectionAlias
+  describe "pending connection users" $ do
+    it "change user id for pending connection" testUpdatePendingConnectionUserId
+    it "change user id for pending connection multiple times" testUpdatePendingConnectionUserIdMulti
+    it "can't change user id for non pending connections" testCantUpdateNonPendingConnectionUserId
```
spaced4ndy, line 69: \+ test when connection is created as incognito, then switched to new user (should unset incognito)
\+ test when switched to new user, then to incognito (should be incognito in that user)
\+ test when switched to new user, then to incognito, then to non incognito (should stay non incognito in that new user)

maybe last one is a bit overkill.. better safe than sorry I guess

src/Simplex/Chat.hs
spaced4ndy: actually we also have to consider re-creating on that user's configured servers.. let's discuss

src/Simplex/Chat.hs
```diff
@@ -1655,6 +1655,43 @@ processChatCommand' vr = \case
     case conn'_ of
       Just conn' -> pure $ CRConnectionIncognitoUpdated user conn'
       Nothing -> throwChatError CEConnectionIncognitoChangeProhibited
+  APISetConnectionUserId connId newUserId -> withUser $ \user@User {userId} -> do
+    cfg <- asks config
+    newUser <- privateGetUser newUserId
+    conn' <- withFastStore $ \db -> do
+      conn <- getPendingContactConnection db userId connId
+      newUserServers <- liftIO $ getServers db cfg newUser SPSMP
+      pure $ checkConnection conn newUserServers
```
spaced4ndy, line 1664: doesn't need to be inside withStore

spaced4ndy: also conn' is misleading as it's not simply a conn but a tuple

src/Simplex/Chat.hs
```diff
@@ -1655,6 +1655,43 @@ processChatCommand' vr = \case
...
+        subMode <- chatReadVar subscriptionMode
+        (newConnId, cReq) <- withAgent $ \a -> createConnection a (aUserId user) True SCMInvitation Nothing IKPQOn subMode
+        newConn <- withFastStore' $ \db -> do
+          deleteConnectionRecord db user connId
+          createDirectConnection db newUser newConnId cReq ConnNew Nothing subMode initialChatVersion PQSupportOn
+        withAgent $ \a -> deleteConnectionAsync a False (aConnId' conn)
+        pure $ CRInvitation user cReq newConn
+      _ -> throwChatError CEInvalidConnReq
+    where
+      getServers :: (ProtocolTypeI p, UserProtocol p) => DB.Connection -> ChatConfig -> User -> SProtocolType p -> IO (NonEmpty (ProtocolServer p))
```
spaced4ndy, line 1682: not necessary to define general function as it's only used for smp

src/Simplex/Chat.hs
```diff
@@ -1655,6 +1655,43 @@ processChatCommand' vr = \case
...
+    conn' <- withFastStore $ \db -> do
+      conn <- getPendingContactConnection db userId connId
+      newUserServers <- liftIO $ getServers db cfg newUser SPSMP
+      pure $ checkConnection conn newUserServers
+    case conn' of
+      (True, Just conn) -> do
+        -- Safe to update connection
+        withAgent $ \a -> changeConnectionUser a (aUserId user) (aConnId' conn) (aUserId newUser)
+        _ <- withFastStore' $ \db -> updatePCCUser db userId conn newUserId
+        pure $ CRConnectionUserChanged user conn newUser
```
spaced4ndy, line 1670: should return updated connection object to ui, so it has less chance of mixing up state (e.g. connection incognito resets in store)

src/Simplex/Chat.hs
```diff
@@ -1655,6 +1655,43 @@ processChatCommand' vr = \case
...
+    newUser <- privateGetUser newUserId
+    conn' <- withFastStore $ \db -> do
+      conn <- getPendingContactConnection db userId connId
+      newUserServers <- liftIO $ getServers db cfg newUser SPSMP
+      pure $ checkConnection conn newUserServers
+    case conn' of
+      (True, Just conn) -> do
+        -- Safe to update connection
+        withAgent $ \a -> changeConnectionUser a (aUserId user) (aConnId' conn) (aUserId newUser)
+        _ <- withFastStore' $ \db -> updatePCCUser db userId conn newUserId
```
spaced4ndy, line 1669: also in both branches we should delete incognito profile in case it was created for connection previously

spaced4ndy: \+ can check in tests it was present and got removed via `hasContactProfiles`

src/Simplex/Chat.hs
```diff
@@ -1655,6 +1655,43 @@ processChatCommand' vr = \case
...
+      getServers db cfg user p = L.map (\ServerCfg {server} -> protoServer server) . useServers cfg p <$> getProtocolServers db user
+      checkConnection :: PendingContactConnection -> NonEmpty (ProtocolServer 'PSMP) -> (Bool, Maybe PendingContactConnection)
+      checkConnection conn@PendingContactConnection {pccConnStatus, connReqInv} newUserServers =
+        case (pccConnStatus, connReqInv) of
+          (ConnNew, Just req) ->
+            let r = connectionData req
+                ConnReqUriData {crSmpQueues = q :| _} = r
+                SMPQueueUri {queueAddress = SMPQueueAddress {smpServer}} = q
+             in (smpServer `elem` newUserServers, Just conn)
+          _ -> (False, Nothing)
```
spaced4ndy, line 1692: this currently creates some indirection and may turn out better if we check pccConnStatus, connReqInv outside and simply throw error based on that, then in good path check servers and do stuff depending on whether server has to change (so this function would only check server is in list)

spaced4ndy: also would be more straightforward to separate reading servers into the good branch then - as we don't need them if link is in bad status

src/Simplex/Chat.hs
```diff
@@ -1655,6 +1655,43 @@ processChatCommand' vr = \case
...
+      connectionData :: ConnReqInvitation -> ConnReqUriData
+      connectionData (CRInvitationUri crData _) = crData
```
spaced4ndy, line 1694:
```suggestion
```

src/Simplex/Chat.hs
```diff
@@ -1655,6 +1655,43 @@ processChatCommand' vr = \case
...
+      _ -> throwChatError CEInvalidConnReq
+    where
+      getServers :: (ProtocolTypeI p, UserProtocol p) => DB.Connection -> ChatConfig -> User -> SProtocolType p -> IO (NonEmpty (ProtocolServer p))
+      getServers db cfg user p = L.map (\ServerCfg {server} -> protoServer server) . useServers cfg p <$> getProtocolServers db user
+      checkConnection :: PendingContactConnection -> NonEmpty (ProtocolServer 'PSMP) -> (Bool, Maybe PendingContactConnection)
+      checkConnection conn@PendingContactConnection {pccConnStatus, connReqInv} newUserServers =
+        case (pccConnStatus, connReqInv) of
+          (ConnNew, Just req) ->
+            let r = connectionData req
+                ConnReqUriData {crSmpQueues = q :| _} = r
```
spaced4ndy, line 1689:
```suggestion
          (ConnNew, Just (CRInvitationUri crData _)) ->
            let ConnReqUriData {crSmpQueues = q :| _} = crData
```

src/Simplex/Chat.hs
```diff
@@ -1655,6 +1655,43 @@ processChatCommand' vr = \case
...
+        pure $ CRConnectionUserChanged user conn newUser
+      (False, Just conn) -> do
+        -- Servers do not match, create new connection
+        subMode <- chatReadVar subscriptionMode
+        (newConnId, cReq) <- withAgent $ \a -> createConnection a (aUserId user) True SCMInvitation Nothing IKPQOn subMode
+        newConn <- withFastStore' $ \db -> do
+          deleteConnectionRecord db user connId
+          createDirectConnection db newUser newConnId cReq ConnNew Nothing subMode initialChatVersion PQSupportOn
+        withAgent $ \a -> deleteConnectionAsync a False (aConnId' conn)
+        pure $ CRInvitation user cReq newConn
```
spaced4ndy, line 1679: I think response should be the same in both cases - as we hide what's happening from ui it doesn't need to differentiate responses

src/Simplex/Chat.hs
```diff
@@ -1655,6 +1655,43 @@ processChatCommand' vr = \case
...
+        _ <- withFastStore' $ \db -> updatePCCUser db userId conn newUserId
+        pure $ CRConnectionUserChanged user conn newUser
+      (False, Just conn) -> do
+        -- Servers do not match, create new connection
+        subMode <- chatReadVar subscriptionMode
+        (newConnId, cReq) <- withAgent $ \a -> createConnection a (aUserId user) True SCMInvitation Nothing IKPQOn subMode
+        newConn <- withFastStore' $ \db -> do
+          deleteConnectionRecord db user connId
+          createDirectConnection db newUser newConnId cReq ConnNew Nothing subMode initialChatVersion PQSupportOn
+        withAgent $ \a -> deleteConnectionAsync a False (aConnId' conn)
```
spaced4ndy, line 1678: I see this is still used in a couple of places (not sure why - maybe there was reason, maybe got overlooked), but I think here we should just use `deleteAgentConnectionAsync` to catch as we already replaced connection in db anyway so there's no recovery for failing to delete it

src/Simplex/Chat.hs
spaced4ndy: would be smth like this:

`L.map (\ServerCfg {server} -> protoServer server) . useServers cfg SPXFTP <$> withStore' (`getProtocolServers` user)`

spaced4ndy: we need to move this logic of retrieving user servers into some separate function already.. I was about to do this refactoring previous time it got repeated again with minor variation. There's 4 or 5 places already that do the same simple thing of reading servers from db and substituting with default if empty, that do it slightly differently every time to accommodate to downstream type instead of mapping after, and each time you have to figure out what's the logic there. not proposing to do it in this pr.

src/Simplex/Chat/Controller.hs
```diff
@@ -403,6 +403,7 @@ data ChatCommand
   | APIAddContact UserId IncognitoEnabled
   | AddContact IncognitoEnabled
   | APISetConnectionIncognito Int64 IncognitoEnabled
+  | APISetConnectionUserId Int64 UserId -- new user id to switch connection to
```
epoberezkin, line 406: let's rename to APIChangeConnectionUser

epoberezkin: (reflects the purpose, not technicality)

src/Simplex/Chat.hs
```diff
@@ -7790,6 +7827,7 @@ chatCommandP =
       "/_connect " *> (APIConnect <$> A.decimal <*> incognitoOnOffP <* A.space <*> ((Just <$> strP) <|> A.takeByteString $> Nothing)),
       "/_connect " *> (APIAddContact <$> A.decimal <*> incognitoOnOffP),
       "/_set incognito :" *> (APISetConnectionIncognito <$> A.decimal <* A.space <*> onOffP),
+      "/_set user :" *> (APISetConnectionUserId <$> A.decimal <* A.space <*> A.decimal),
```
epoberezkin, line 7830:
```suggestion
      "/_set conn user :" *> (APISetConnectionUserId <$> A.decimal <* A.space <*> A.decimal),
```

src/Simplex/Chat.hs
epoberezkin, line 1664:
```suggestion
    newUserServers <- withFastStore $ \db -> do
      conn <- getPendingContactConnection db userId connId
      liftIO $ getServers db cfg newUser SPSMP
```

src/Simplex/Chat.hs
```diff
@@ -1655,6 +1655,43 @@ processChatCommand' vr = \case
...
       Just conn' -> pure $ CRConnectionIncognitoUpdated user conn'
       Nothing -> throwChatError CEConnectionIncognitoChangeProhibited
+  APISetConnectionUserId connId newUserId -> withUser $ \user@User {userId} -> do
+    cfg <- asks config
+    newUser <- privateGetUser newUserId
+    conn' <- withFastStore $ \db -> do
+      conn <- getPendingContactConnection db userId connId
+      newUserServers <- liftIO $ getServers db cfg newUser SPSMP
+      pure $ checkConnection conn newUserServers
+    case conn' of
```
epoberezkin, line 1665:
```suggestion
    case checkConnection conn newUserServers of
```

epoberezkin, line 1665:
```suggestion
    case checkConnection conn newUserServers of
```

PR #4683, opened 2024-08-14
ios: add text to share extension link preview

levitatingpineapple:


apps/ios/Shared/Views/Chat/ComposeMessage/ComposeLinkView.swift
```diff
@@ -55,7 +55,7 @@ struct ComposeLinkView: View {
                     .foregroundColor(theme.colors.secondary)
             }
             .padding(.vertical, 5)
-            .frame(maxWidth: .infinity, minHeight: 60, maxHeight: 60)
```
levitatingpineapple, line 58: Setting height limit will make the view's text go out of bounds with larger sizes.

Screenshot 2024-08-14 at 10 29 48

PR #4695, opened 2024-08-15
ios: improve animation when receiving messages

levitatingpineapple:


apps/ios/Shared/Views/Chat/ReverseList.swift
```diff
@@ -171,16 +171,32 @@ struct ReverseList<Item: Identifiable & Hashable & Sendable, Content: View>: UIV
             Task { representer.scrollState = .atDestination }
         }
 
-        func update(items: Array<Item>) {
+        var updateInProgress = false
+
+        func update(items: [Item]) {
+            if updateInProgress { return }
```
levitatingpineapple, line 177: Technically some item updates could get lost due to an early return
In practice seems to work fine
Required, since non-animated update can take longer than 0.25s to return completion callback

PR #4697, opened 2024-08-15
android: target API level 34 (Android 14)

avently:


apps/multiplatform/android/src/main/AndroidManifest.xml
```diff
@@ -133,15 +139,19 @@
         android:name=".SimplexService"
         android:enabled="true"
         android:exported="false"
-        android:stopWithTask="false"></service>
+        android:stopWithTask="false"
+        android:foregroundServiceType="remoteMessaging"
```
dpad85, line 143: Hey, of all the available types, `remoteMessaging` does look like the most relevant, but [the description in the doc](https://developer.android.com/develop/background-work/services/fg-service-types#remote-messaging) is ambiguous : 
> Transfer text messages from one device to another. Assists with continuity of a user's messaging tasks when they switch devices.

That description could be interpreted as only appropriate to a one-time data migration from a user's old device to a new one, but not to frequent and long-lived data transfers with a service.

For context, I'm working on [Phoenix](https://github.com/ACINQ/phoenix), a self-custodial bitcoin wallet, and we have similar constraints (though instead of text messages, our app receives payments in the background). I've started implementing a permanent foreground service to do that (kinda similar [to what you've done](https://simplex.chat/blog/20220404-simplex-chat-instant-notifications.html)), but my main issue has been with finding a nice foreground type that would not cause any issues with the new GooglePlay policies. 

So I'd like to know your thoughts on all this. If my comment is not relevant, feel free to dismiss it. I can also move it to a GitHub discussion.

avently: https://support.google.com/googleplay/android-developer/answer/13392821?hl=en

> Assists without continuity of a user's messaging tasks when they switch devices. For example, allow users to send text messages via a web client by scanning a QR code and establishing a connection between the web client and the phone. Not limited to SMS.

In the case of our app, it's kind of required to have a persistent connection in order to send and receive messages. We don't have any Google-backed push service. And we also use that foreground service when establishing a connection with a remote desktop, which means a user can link mobile with desktop and send messages from mobile using his desktop (which is exactly what they write about that flag). 

And as you said:

>of all the available types, remoteMessaging does look like the most relevant

And since we only have another option called "specialUse" (we can't just skip specifying any type at all) it's probably better to use remote messaging than to try to explain to the reviewer what we actually do in the app to have a specialUse (which is probably unoptimized by Android). 

In terms of background activity having a connection to a web client (as they say) is the same thing as having a connection to SMP servers (but much more connections, of course). So their optimization for this specific flag should be enough to handle it gracefully for a user device in terms of memory/CPU, etc.

So, they give us requirements. If they don't think we do the right thing, they will give us a hint of any other alternative. I don't know how choosing specialUse can be better than remoteMessaging.

PR #4704, opened 2024-08-16
core: multi forward api

spaced4ndy:


src/Simplex/Chat.hs
```diff
@@ -7006,59 +7165,33 @@ sendGroupMessage user gInfo members chatMsgEvent = do
       currentTs <- liftIO getCurrentTime
       withStore' $ \db -> updateUserMemberProfileSentAt db user gInfo currentTs
 
-type GroupSndResultData = (([Either ChatError ([Int64], PQEncryption)], [(GroupMember, Connection)]), ([Either ChatError ()], [GroupMember]), [GroupMember])
-
 data GroupSndResult = GroupSndResult
-  { sentTo :: [GroupMember],
-    pending :: [GroupMember],
+  { sentTo :: [(GroupMemberId, Either ChatError [MessageId], Either ChatError ([Int64], PQEncryption))],
```
spaced4ndy, line 7169: It's a different GroupSndResult from previous. This one is closer to GroupSndResultData in purpose. GroupSndResultData is a worse name, so new code uses name from one and purpose from other. Diff is bad regardless of its position.

src/Simplex/Chat.hs
```diff
@@ -977,92 +978,119 @@ processChatCommand' vr = \case
...
+          forwardCIFF :: CChatItem c -> Maybe CIForwardedFrom -> Maybe CIForwardedFrom
+          forwardCIFF (CChatItem _ ChatItem {meta = CIMeta {itemForwarded}}) ciff = case itemForwarded of
             Nothing -> ciff
             Just CIFFUnknown -> ciff
             Just prevCIFF -> Just prevCIFF
-          forwardCryptoFile :: ChatItem c d -> CM (Maybe CryptoFile)
-          forwardCryptoFile ChatItem {file = Nothing} = pure Nothing
-          forwardCryptoFile ChatItem {file = Just ciFile} = case ciFile of
-            CIFile {fileName, fileStatus, fileSource = Just fromCF@CryptoFile {filePath}}
-              | ciFileLoaded fileStatus ->
```
spaced4ndy, line 1046: ui may offer to preemptively download files in case it's possible for any files. if any files are missing it can show a bar near multi select buttons saying "x messages will be sent without files", or alert with confirmation because it doesn't make sense to show for delete actions

src/Simplex/Chat.hs
```diff
@@ -2858,77 +2886,152 @@ processChatCommand' vr = \case
...
+                Nothing -> pure (Nothing, Nothing)
+            prepareMsgs :: NonEmpty (ComposeMessageReq, Maybe FileInvitation) -> Maybe CITimed -> CM (NonEmpty (MsgContainer, Maybe (CIQuote 'CTDirect)))
+            prepareMsgs cmsFileInvs timed_ =
+              forM cmsFileInvs $ \((ComposedMessage {quotedItemId, msgContent = mc}, itemForwarded), fInv_) ->
+                case (quotedItemId, itemForwarded) of
+                  (Nothing, Nothing) -> pure (MCSimple (ExtMsgContent mc fInv_ (ttl' <$> timed_) (justTrue live)), Nothing)
+                  (Nothing, Just _) -> pure (MCForward (ExtMsgContent mc fInv_ (ttl' <$> timed_) (justTrue live)), Nothing)
+                  (Just qiId, Nothing) -> do
+                    CChatItem _ qci@ChatItem {meta = CIMeta {itemTs, itemSharedMsgId}, formattedText, file} <-
+                      withFastStore $ \db -> getDirectChatItem db user contactId qiId
```
spaced4ndy, line 2933: it doesn't seem critical to allow quote + multi select because UI would be complicated as well, so api prohibits it and doesn't bother to batch here. all other message preparation steps except this are pure - so we only do single db read in case of 1 quoted item

src/Simplex/Chat/View.hs
```diff
@@ -120,7 +120,11 @@ responseToView hu@(currentRH, user_) ChatConfig {logLevel, showReactions, showRe
   CRConnectionVerified u verified code -> ttyUser u [plain $ if verified then "connection verified" else "connection not verified, current code is " <> code]
   CRContactCode u ct code -> ttyUser u $ viewContactCode ct code testView
   CRGroupMemberCode u g m code -> ttyUser u $ viewGroupMemberCode g m code testView
-  CRNewChatItem u (AChatItem _ _ chat item) -> ttyUser u $ unmuted u chat item $ viewChatItem chat item False ts tz <> viewItemReactions item
+  CRNewChatItems u chatItems ->
```
epoberezkin, line 123: will it correctly show multiple lines? don't think so...

I think `ttyUser u` should be added to each line? or is it?

src/Simplex/Chat/Terminal/Output.hs
```diff
@@ -147,7 +147,7 @@ runTerminalOutput ct cc@ChatController {outputQ, showLiveItems, logFilePath} Cha
   forever $ do
     (_, outputRH, r) <- atomically $ readTBQueue outputQ
     case r of
-      CRNewChatItem u ci -> when markRead $ markChatItemRead u ci
+      CRNewChatItems u (ci : _) -> when markRead $ markChatItemRead u ci -- At the moment of writing received items are created one at a time
```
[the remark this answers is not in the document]
spaced4ndy: no, because we don't create more than 1 received item at a time, and sent items we don't mark as read - that's what's comment is about

PR #4739, opened 2024-08-22
ios: multi send & forward api

spaced4ndy:


apps/ios/Shared/Model/SimpleXAPI.swift
```diff
@@ -380,10 +380,13 @@ private func processSendMessageCmd(toChatType: ChatType, cmd: ChatCommand) async
...
-        if case let .newChatItem(_, aChatItem) = r {
-            cItem = aChatItem.chatItem
-            chatModel.messageDelivery[aChatItem.chatItem.id] = endTask
-            return cItem
+        if case let .newChatItems(_, aChatItems) = r {
+            let cItems = aChatItems.map { $0.chatItem }
+            if let cItemLast = cItems.last {
+                cItem = cItemLast
+                chatModel.messageDelivery[cItemLast.id] = endTask
+            }
```
spaced4ndy, line 388: only creating UI delivery record for last item - this seems sufficient as they are sent in order in single connection, so events also should come in order, and receiving event for last item would mean all have been processed. I don't think I'm missing anything, am I?

PR #4741, opened 2024-08-22
android, desktop: allow for chat profile selection on new chat screen

diogofcunha:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
```diff
@@ -784,7 +784,8 @@ object ChatModel {
 data class ShowingInvitation(
   val connId: String,
   val connReq: String,
-  val connChatUsed: Boolean
+  val connChatUsed: Boolean,
+  val conn: PendingContactConnection
```
spaced4ndy, line 788: connId, connReq may be already available in PCC

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/newchat/NewChatView.kt
```diff
@@ -192,6 +204,244 @@ private fun RetryButton(onClick: () -> Unit) {
...
+    horizontalArrangement = Arrangement.SpaceBetween,
+    verticalAlignment = Alignment.CenterVertically
+  ) {
+    Row(verticalAlignment = Alignment.CenterVertically) {
+      image()
+      TextIconSpaced(false)
+      Text(title, modifier = Modifier.align(Alignment.CenterVertically))
+      if (infoIcon != null) {
+        TextIconSpaced(false)
+        infoIcon()
```
spaced4ndy, line 231: tappable area should be circle and bigger, like on the previous view

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/newchat/NewChatView.kt
```diff
@@ -192,6 +207,254 @@ private fun RetryButton(onClick: () -> Unit) {
...
+  disabled: Boolean,
+  onSelected: () -> Unit,
+  image: @Composable () -> Unit,
+  onInfo: (@Composable () -> Unit)? = null
+) {
+  Row(
+    Modifier
+      .fillMaxWidth()
+      .sizeIn(minHeight = DEFAULT_MIN_SECTION_ITEM_HEIGHT)
+      .then(if (disabled) Modifier else Modifier.clickable(onClick = onSelected))
```
avently, line 223: why not just .clickable(enabled = !disabled, onClick = onSelected)

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/newchat/NewChatView.kt
```diff
@@ -192,6 +207,254 @@ private fun RetryButton(onClick: () -> Unit) {
...
 }
 
+@Composable
+private fun ProfilePickerOption(
+  title: String,
+  selected: Boolean,
+  disabled: Boolean,
+  onSelected: () -> Unit,
+  image: @Composable () -> Unit,
+  onInfo: (@Composable () -> Unit)? = null
```
avently, line 217: no reason for onInfo to be @Composable

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/newchat/NewChatView.kt
```diff
@@ -192,6 +207,254 @@ private fun RetryButton(onClick: () -> Unit) {
...
+      .sizeIn(minHeight = DEFAULT_MIN_SECTION_ITEM_HEIGHT)
+      .then(if (disabled) Modifier else Modifier.clickable(onClick = onSelected))
+      .padding(horizontal = DEFAULT_PADDING),
+    horizontalArrangement = Arrangement.SpaceBetween,
+    verticalAlignment = Alignment.CenterVertically
+  ) {
+    Row(verticalAlignment = Alignment.CenterVertically) {
+      image()
+      TextIconSpaced(false)
+      Text(title, modifier = Modifier.align(Alignment.CenterVertically))
```
avently, line 231: You have the same thing on the whole Row, are you sure you need the second one?

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/newchat/NewChatView.kt
```diff
@@ -192,6 +207,254 @@ private fun RetryButton(onClick: () -> Unit) {
...
+    progressByTimeout = if (switchingProfile.value) {
+      delay(500)
+      switchingProfile.value
+    } else {
+      false
+    }
+  }
+
+  @Composable
+  fun profilePickerUserOption(user: User) {
```
avently, line 315: `@Composables` should always have Capitalized function names. And please don't include functions inside functions. It's a bad practice in Kotlin and should be avoided at any costs except if you need to pass millions of arguments to such function and you prefer to get those arguments from outer function (it makes code unreadable). If it is the case, keep your version

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/newchat/NewChatView.kt
```diff
@@ -192,6 +207,254 @@ private fun RetryButton(onClick: () -> Unit) {
...
+      },
+      onInfo = { ModalManager.start.showModal { IncognitoView() } },
+    )
+  }
+
+  BoxWithConstraints {
+    Column(
+      Modifier
+        .fillMaxSize()
+        .then(
```
avently, line 404: why not just

```kotlin
            Modifier
              .alpha(if (progressByTimeout) 0.6f else 1f)
```

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/newchat/NewChatView.kt
```diff
@@ -204,23 +467,75 @@ private fun InviteView(rhId: Long?, connReqInvitation: String, contactConnection
...
-  KeyChangeEffect(incognito.value) {
-    withBGApi {
-      val contactConn = contactConnection.value ?: return@withBGApi
-      val conn = controller.apiSetConnectionIncognito(rhId, contactConn.pccConnId, incognito.value) ?: return@withBGApi
-      withChats {
-        contactConnection.value = conn
-        updateContactConnection(rhId, conn)
+
+  val currentUser by remember(chatModel.currentUser.value) {
+    derivedStateOf { chatModel.currentUser.value }
```
avently, line 478: why not just
```kotlin

val currentUser by remember { chatModel.currentUser } 
```

Same thing in different place

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/newchat/NewChatView.kt
diogofcunha: yes that is the case here, would require 6 arguments to be passed around

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/newchat/NewChatView.kt
```diff
@@ -204,23 +467,75 @@ private fun InviteView(rhId: Long?, connReqInvitation: String, contactConnection
...
+  }
+
+  currentUser?.let {
```
avently, line 481: the block of code is quite large so if you want to use .let here, specify it like so at least:
```kotlin
currentUser?.let { user ->
```

it will be easier to track that you want to use the value, not just checking for non-null this way. 
Also we try to not use .let at all because it makes some people crazy when they read it:)
So this is better:
```kotlin
val user = currentUser.value
if (user != null) {
...
}
```

avently:
Is it intended that sorting changes all the time when you change a user? Maybe it's better to have constant place for each user like we do in UserProfilesView. In current situation since we need to have incognito option would be better to do like:
1. Always active user selected globally (user.activeUser)
2. Always incognito option
3. Other users with constant sorting

diogofcunha:
> Is it intended that sorting changes all the time when you change a user? Maybe it's better to have constant place for each user like we do in UserProfilesView. In current situation since we need to have incognito option would be better to do like:
> 
> 1. Always active user selected globally (user.activeUser)
> 2. Always incognito option
> 3. Other users with constant sorting
> 
> Edit: ah, I see that something like this is happening now. I didn't notice that active user is changing in response to changing it in connection

yes, the logic we settled on is similar to what you proposed.

Only difference is that the selected is always on top (and when it is incognito that is the one that's on top).

1. Selected (profile or incognito)
2. Incognito (if not selected)
3. Alphabetical ordering

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/newchat/NewChatView.kt
```diff
@@ -192,6 +207,254 @@ private fun RetryButton(onClick: () -> Unit) {
...
+    val selected = selectedProfile?.userId == user.userId && !incognito
+
+    ProfilePickerOption(
+      title = user.chatViewName,
+      disabled = switchingProfile.value || selected,
+      selected = selected,
+      onSelected = {
+        switchingProfile.value = true
+        withApi {
+          if (contactConnection != null) {
```
avently, line 324: put the whole block into try { ... } finally { switchingProfile.value = false } because if something will crash, that thing will never become `false`

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/newchat/NewChatView.kt
avently:
```kotlin
val currentUser = remember { chatModel.currentUser }.value
  if (currentUser != null) {
```

PR #4743, opened 2024-08-22
ios: time based message grouping

levitatingpineapple:


apps/ios/Shared/Model/ChatModel.swift
```diff
@@ -428,19 +436,17 @@ final class ChatModel: ObservableObject {
 
     private func _upsertChatItem(_ cInfo: ChatInfo, _ cItem: ChatItem) -> Bool {
         if let i = getChatItemIndex(cItem) {
-            withConditionalAnimation {
-                _updateChatItem(at: i, with: cItem)
-            }
```
levitatingpineapple, line 433: Two instances of `withConditionalAnimation ` removed.
- They haven't had effect since moving to reverse list
- They where interfering with chat item view update, when item below it causes the change

apps/ios/Shared/Views/Chat/ChatItemView.swift
```diff
@@ -9,9 +9,21 @@
 import SwiftUI
 import SimpleXChat
 
+extension EnvironmentValues {
```
epoberezkin, line 12: why is this needed

apps/ios/Shared/Views/Chat/ChatItemView.swift
```diff
@@ -9,9 +9,21 @@
...
+    var showTimestamp: Bool {
+        get { self[ShowTimestamp.self] }
+        set { self[ShowTimestamp.self] = newValue }
+    }
+}
+
 struct ChatItemView: View {
     @ObservedObject var chat: Chat
     @EnvironmentObject var theme: AppTheme
+    @Environment(\.showTimestamp) var showTimestamp
```
epoberezkin, line 26: why does it need to be Environment? Seems some fancy way to pass parameter.

apps/ios/Shared/Views/Chat/ChatView.swift
```diff
@@ -716,31 +717,53 @@ struct ChatView: View {
 
         var revealed: Bool { chatItem == revealedChatItem }
 
+        struct TimeSeparation {
```
epoberezkin, line 720: why is it a struct?

epoberezkin: we are passing two parameters, any encapsulation beyond maybe ad hoc tuple, is not necessary here. Or just two parameters

apps/ios/Shared/Views/Chat/ChatView.swift
```diff
@@ -716,31 +717,53 @@ struct ChatView: View {
...
+            let isTimestampShown: Bool
+            let isGapLarge: Bool
+
+            init(for chatItem: ChatItem, at index: Int?) {
```
epoberezkin, line 724: inits are not supposed to do any work, they are supposed to be dummy constructor.

Making inits do some logic is OOP anti-pattern

PR #4744, opened 2024-08-22
ios: connection profile search, incognito info in selection list and improved loader

diogofcunha:


apps/ios/Shared/Views/NewChat/NewChatView.swift
```diff
@@ -487,6 +502,12 @@ private struct ActiveProfilePicker: View {
                     incognitoProfileImage()
                     Text("Incognito")
                         .foregroundColor(theme.colors.onBackground)
+                    Image(systemName: "info.circle")
+                        .foregroundColor(theme.colors.primary)
+                        .font(.system(size: 14))
+                        .onTapGesture {
+                            showIncognitoSheet = true
+                        }
```
spaced4ndy, line 510: add Spacer? should look better with info in the end. I know there's a checkmark on selection - maybe it should replace info, or be placed left from it

apps/ios/Shared/Views/NewChat/NewChatView.swift
```diff
@@ -459,19 +459,34 @@ private struct ActiveProfilePicker: View {
...
+                }
+            }
+    }
+    
+    private func filteredProfiles() -> [User] {
+        let s = trimmedSearchTextOrPassword
+        let lower = s.localizedLowercase
+        
+        return profiles.filter { u in
+            if (u.activeUser || !u.hidden) && (s == "" || u.chatViewName.localizedLowercase.contains(lower)) {
```
spaced4ndy, line 486: switching connection to hidden user produces "unexpected" error. As the user already successfully entered the password it should let him into profile on switch

apps/ios/Shared/Views/NewChat/NewChatView.swift
```diff
@@ -495,7 +516,8 @@ private struct ActiveProfilePicker: View {
                     }
                 }
             }
-            ForEach(profiles) { item in
+            let filteredProfiles = filteredProfiles()
+            ForEach(filteredProfiles) { item in
```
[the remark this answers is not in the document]
epoberezkin: maybe, don't have strong opinion here. I think it's ok as is.

apps/ios/Shared/Views/NewChat/NewChatView.swift
epoberezkin: not better. Let's keep where it was before please

epoberezkin: The reason why it is a bad idea to put info in the end is because if you are right handed, as the majority of people are, you are tapping the row where you see the tick. Now you have info there, so you can't tap this place...

PR #4787, opened 2024-08-28
ios: make CallKit calls fire in time after cold start

avently:


apps/ios/Shared/Views/Call/CallController.swift
```diff
@@ -156,6 +174,22 @@ class CallController: NSObject, CXProviderDelegate, PKPushRegistryDelegate, Obse
...
+            return true
+        }
+        logger.debug("CallController waiting until chat started")
+        var t: UInt64 = 0
+        repeat {
+            _ = try? await Task.sleep(nanoseconds: stepMs * 1000000)
+            t += stepMs
+            if ChatModel.shared.chatInitialized, ChatModel.shared.chatRunning == true, case .active = AppChatState.shared.value {
+                return true
+            }
```
epoberezkin, line 188:
```suggestion
            if ChatModel.shared.chatInitialized, ChatModel.shared.chatRunning == true, case .active = AppChatState.shared.value {
                return true
            }
            _ = try? await Task.sleep(nanoseconds: stepMs * 1000000)
            t += stepMs
```

PR #4789, opened 2024-08-28
ios: add floating date separator

levitatingpineapple:


levitatingpineapple:
https://github.com/simplex-chat/simplex-chat/pull/4801

PR #4799, opened 2024-08-29
ios: add material background to elements over image and video messages

levitatingpineapple:


levitatingpineapple:
https://github.com/simplex-chat/simplex-chat/pull/4808

PR #4801, opened 2024-08-29
ios: add floating date separator

levitatingpineapple:


apps/ios/Shared/Views/Chat/ChatView.swift
```diff
@@ -454,50 +454,55 @@ struct ChatView: View {
...
         let scrollOffset = CurrentValueSubject<Double, Never>(0)
+
         private var bag = Set<AnyCancellable>()
 
+        typealias ListState = (
+            topItemDate: Date?,
+            bottomItemId: ChatItem.ID?
+        )
+
+        private struct ViewUpdate: Equatable {
```
levitatingpineapple, line 467: `ViewUpdate` struct added to satisfy`.removeDuplicates()` requirement for `Equatable` conformance.

PR #4820, opened 2024-09-03
ios: make message information on media readable

levitatingpineapple:


apps/ios/Shared/Views/Chat/ChatItem/CIVideoView.swift
```diff
@@ -288,32 +289,23 @@ struct CIVideoView: View {
...
-                .font(.caption)
-                .padding(.vertical, 3)
-                .padding(.horizontal, 6)
-                .background(Color.black.opacity(0.35))
-                .cornerRadius(10)
-                .padding(.top, 6)
-            }
-        }
+            .padding(.vertical, 6)
+            .padding(.horizontal, 12)
```
levitatingpineapple, line 308: Matches the padding with metadata in bottom trailing corner

apps/ios/Shared/Views/Chat/ChatItem/CIMetaView.swift
```diff
@@ -45,7 +49,7 @@ func ciMetaText(
     _ meta: CIMeta,
     chatTTL: Int?,
     encrypted: Bool?,
-    color: Color = .clear,
+    color: Color? = .clear,
```
levitatingpineapple, line 52: `Color?` -  passing nil here means that `.foregroundColor` modifier gets completely omitted.

PR #4857, opened 2024-09-10
ios: bulk forward

levitatingpineapple:


apps/ios/Shared/Views/Chat/ChatView.swift
```diff
@@ -701,6 +714,80 @@ struct ChatView: View {
...
+                    }
+                }
+            }
+        }
+
+        func openForwardingSheet(_ items: [Int64]) {
+            let im = ItemsModel.shared
+            forwardedChatItems = items
+                .compactMap { id in im.reversedChatItems.first { $0.id == id } }
+                .sorted { $0.meta.itemTs == $1.meta.itemTs }
```
diogofcunha, line 787: It seems there is an issue with forwarding order in IOS, we do quite a bit of rounding to the second on the backend and it seems doing this sort isn't reliable. Can't we just keep the selection of the ids as before, and take advantage of the ordering already in the chat model and just filter here? That's the approach we took on android and it seems to be accurate

PR #4860, opened 2024-09-10
core: bulk forward missing files error handling

diogofcunha:


src/Simplex/Chat/View.hs
```diff
@@ -2034,7 +2034,11 @@ viewChatError isCmd logLevel testView = \case
     CEFallbackToSMPProhibited fileId -> ["recipient tried to accept file " <> sShow fileId <> " via old protocol, prohibited"]
     CEInlineFileProhibited _ -> ["A small file sent without acceptance - you can enable receiving such files with -f option."]
     CEInvalidQuote -> ["cannot reply to this message"]
-    CEInvalidForward -> ["cannot forward this message"]
+    CEInvalidForward -> ["cannot forward this message(s)"]
```
spaced4ndy, line 2037:
```suggestion
    CEInvalidForward -> ["cannot forward message(s)"]
```

src/Simplex/Chat/Messages.hs
```diff
@@ -595,6 +595,27 @@ ciFileLoaded = \case
   CIFSRcvWarning {} -> False
   CIFSInvalid {} -> False
 
+data ForwardFileError = FFENotAccepted FileTransferId | FFEInProgress | FFEFailed | FFEMissing
+  deriving (Eq, Ord)
+
+ciFileForwardError :: FileTransferId -> CIFileStatus d -> Maybe ForwardFileError
```
spaced4ndy, line 601: to not duplicate logic ciFileLoaded function could be defined via this one (isJust, seems like it's full match)

src/Simplex/Chat/Messages.hs
```diff
@@ -595,6 +595,27 @@ ciFileLoaded = \case
...
+  CIFSSndStored -> Nothing
+  CIFSSndTransfer {} -> Nothing
+  CIFSSndComplete -> Nothing
+  CIFSSndCancelled -> Nothing
+  CIFSSndError {} -> Nothing
+  CIFSSndWarning {} -> Nothing
+  CIFSRcvInvitation -> Just $ FFENotAccepted fId
+  CIFSRcvAccepted -> Just FFEInProgress
+  CIFSRcvTransfer {} -> Just FFEInProgress
+  CIFSRcvAborted -> Just FFEFailed
```
spaced4ndy, line 612: possibly this status should translate into FFENotAccepted as it's happening on NOT_APPROVED error (unknown servers)

src/Simplex/Chat.hs
```diff
@@ -980,86 +980,64 @@ processChatCommand' vr = \case
...
-              (mc, _) <- forwardMC ci
-              file <- forwardCryptoFile ci
-              let ciff = forwardCIFF ci Nothing
-              pure (ComposedMessage file Nothing mc, ciff)
+          items <- withFastStore $ \db -> mapM (getLocalChatItem db user fromChatId) itemIds
+          mapM (ciComposeMsgReq Nothing) items
         CTContactRequest -> throwChatError $ CECommandError "not supported"
         CTContactConnection -> throwChatError $ CECommandError "not supported"
         where
+          ciComposeMsgReq :: Maybe (MsgDirection -> Maybe Int64 -> Maybe ChatItemId -> CIForwardedFrom) -> CChatItem d -> CM (Either (ForwardFileError, Bool) ComposeMessageReq)
```
spaced4ndy, line 1032: unclear what's the reason for this refactoring. if there's no reason better to revert it

src/Simplex/Chat/Messages.hs
diogofcunha: ciFileLoaded can now be removed, it's unused

src/Simplex/Chat/Messages.hs
spaced4ndy: FFECanReceive would be a better name then

PR #4861, opened 2024-09-11
ios: batch previews, when sending media messsages

levitatingpineapple:


apps/ios/Shared/Views/Chat/ComposeMessage/ComposeView.swift
```diff
@@ -167,6 +167,13 @@ struct ComposeState {
         }
     }
 
+    var manyMediaPreviews: Bool {
```
spaced4ndy, line 170: limited live message to a single attachment but there's still a bug when starting live message with a single attachment that each new live message send is a separate send

spaced4ndy (approved):

Issue #4837, opened 2024-09-06
[Bug]: [Xcode 16] iOS App crashes with `EXC_BAD_ACCESS` when decoding JSON in `chatResponse()`


acevif:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Platform

iOS

### OS version

iOS 17.6.1

### App version

v6.0.4

### Current Behavior

When building v6.0.4 with Xcode 16, the app crashes immediately upon launch due to a runtime exception. 

The error is `EXC_BAD_ACCESS code=2`, which typically indicates an attempt to access memory that the app does not have permission to access. The crash occurs when attempting to decode a JSON response using `JSONDecoder.decode(_:from:)`.

The exception is thrown at line 186 in `API.swift` (`public func chatResponse(_ s: String) -> ChatResponse`).

line 186: `let r = try jsonDecoder.decode(APIResponse.self, from: d)`

The content of the string `s` that is causing the crash is as follows:
```
{"resp":{"_owsf":true,"chatCmdError":{"chatError":{"_owsf":true,"error":{"errorType":{"_owsf":true,"noActiveUser":{}}}}}}}
```

### Expected Behavior

The app should not crash. When building with Xcode 15, the app runs as expected without any crashes.

### Steps To Reproduce

1. `$ nix build .\#aarch64-darwin-ios:lib:simplex-chat`
2. Unzip `pkg-ios-aarch64-swift-json.zip` to `apps/ios/Libraries` directory.
3. Open the project in Xcode 16 (16.0 Beta 6).
4. Adjust project settings (e.g., developer team, etc.)
5. Build and run the app on an iPhone(physical device).

### Relevant log output

**Log:**

```
DEBUGGING: startChatAndActivate
scenePhase was inactive, now active
CallController: didUpdate push credentials for type PKPushTypeVoIP
AppDelegate: didRegisterForRemoteNotificationsWithDeviceToken xxxxxxxxxxxxxxxxxxxxxxxx
pushEnvironment: development
initializeChat
chatMigrateInit uses keychain: true
chatMigrateInit DB path: /private/var/mobile/Containers/Shared/AppGroup/967A9074-0264-4F0B-9E0B-E827566BA09E/simplex_v1, confirm: yesUp
chatSendCmd apiSetAppFilePaths
setNetworkInfo Network changed: UserNetworkInfo(networkType: SimpleXChat.UserNetworkType.wifi, online: true)
[...]

PR #4867, opened 2024-09-12
ios: Workaround for stack overflow crash with Xcode 16

acevif:
- Increased stack size to 2MiB
- Fix: https://github.com/simplex-chat/simplex-chat/issues/4837


acevif:

acevif:
In my environment, 992KiB works, but 976KiB does not.

acevif:
I am thinking of setting the stack size to 2MiB. 4MiB seems too large, and 2MiB should provide sufficient headroom.

epoberezkin:
It does look like something that Xcode might need to fix, isn't it?

Is there some bug/issue/discussion where they track it?

acevif:
@epoberezkin 

It might be a bug in Xcode, but I can't completely rule out the possibility that the required stack size has increased due to (for example) changes in the specifications related to Swift concurrency support.

I searched for relevant materials and related bugs/issues/discussions, but I haven't been able to find anything so far.

What version of Xcode is your team currently using?

P.S. I plan to check later if this issue has been resolved in Xcode 16.1 or Xcode 16.2.

yspreen:
@acevif I think I just ran into a similar bug in YouTube player kit. I am using Xcode 16.2

avently:
@acevif first of all, thank you for this workaround/fix. Because it allowed me to run the build from Xcode 16. And sorry for such a slow reply - we just didn't plan to update to a new Xcode until it gets fixed by Apple but seems like it isn't considered a bug if it's still in 16.2.

> What version of Xcode is your team currently using?

We have been using Xcode version prior to 16.0 (maybe I had 15.4). I updated to 16.2 now and I see no changes in behavior of the app, so it still crashes on jsonDecoder.decode.

Just for a context. Recenty we discovered crashes in production version of the app (not development) on iOS 18 only and the problem as I found was in stack overflow as well but it was in random places in ChatView and didn't look like stackoverflow. So just rendering views made the app crashed. Very hard to diagnose the crash right in the release day. Anyway, I added this line and it helped with that bug: https://github.com/simplex-chat/simplex-chat/pull/5734/files

But. It doesn't help with the bug you made a fix for because that line only changes stack size for a main thread.

How did you discover that solution?

avently:
> In my environment, 992KiB works, but 976KiB does not.

In my case on the current code base 1120 works but 1104 doesn't. 

After I added 11 new lines to APIResponse enum (like `case apiSettings2(appSettings: AppSettings)`), the app started to crash with 1120KiB. So the stack size we need to set directly depends of the size of enum. And the app started to work in these conditions (+11 lines in enum) when I changed stack size to 1184KiB.

So I think setting 2 MB is ok for now. If it starts to crash again, it can be increased later. For now we can test how this solution works in real life.

Note that I'm on iOS 16, so not only iOS 18 has this problem.

epoberezkin:
pushed to #5771 (to not pull fork)

epoberezkin:
@acevif thanks for the solution. It's weird that simple JSON parsing creates stack overflow... It's merged now via another branch, as Apple requires Xcode 16 from mid-April.

PR #4868, opened 2024-09-12
android, desktop: bulk forward

diogofcunha:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ContextItemView.kt
```diff
@@ -62,7 +62,7 @@ fun ContextItemView(
     )
   }
 
-  fun attachment(): ImageResource? =
+  fun attachment(contextItem: ChatItem): ImageResource? =
```
spaced4ndy, line 65: shouldn't show attachment icon if file is not loaded

spaced4ndy: unless it's an image

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ShareListView.kt
```diff
@@ -65,6 +65,16 @@ fun ShareListView(chatModel: ChatModel, stopped: Boolean) {
           hasSimplexLink = hasSimplexLink(mc.text)
         }
       }
+      is SharedContent.BulkForward -> {
+        sharedContent.chatItems.forEach {
```
spaced4ndy, line 69: this seems to be wrong - it should be doing any (or w/e it's called in kotlin) instead of forEach - right now if you select multiple items, preferences are only checked against last item

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatView.kt
```diff
@@ -1416,6 +1440,76 @@ private fun TopEndFloatingButton(
...
+@Composable
+private fun HideForwardConfirmationButton(
+  text: String,
+  modifier: Modifier = Modifier,
+  contentPadding: PaddingValues = ButtonDefaults.TextButtonContentPadding
+) {
+  TextButton(onClick = { AlertManager.shared.hideAlert() }, modifier = modifier, contentPadding = contentPadding) {
+    Text(text, textAlign = TextAlign.Center, color = MaterialTheme.colors.primary)
+  }
+}
```
spaced4ndy, line 1482: can be removed
```suggestion
```

spaced4ndy (approved):

PR #4889, opened 2024-09-16
ui: translations

epoberezkin:


summoner001:
Can you please open the weblate? 
There are suggestions for the Hungarian translation, but those cannot be accepted because the weblate is closed.

PR #4914, opened 2024-09-20
android, desktop:  time based message grouping and day separators

diogofcunha:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
```diff
@@ -2335,6 +2335,10 @@ sealed class CIDirection {
     is LocalSnd -> true
     is LocalRcv -> false
   }
+
+  fun sameDirection(dir: CIDirection): Boolean {
```
avently, line 2339: incorrectly call this function as sameDirection because it's name suggests to think about same as snd == snd, or DirectSnd == GroupSnd. Unclear the meaning, you have to read the code to understand. Since you use it only once, make this function local, or better just use it as an oneliner, like `val sameMemberAndDirection = ...`

avently:
Overall looks good to me. Pushed some not-important changes. 
I would say that in landscape mode on Android it would be better to show multiple profiles instead of one long profile. Screen is very long, it should help with profile switching if more profiles can be visible at once

avently:
It's a little weird that padding for hovered message in chat in selectable mode is not the bottom (should be half at top and half at bottom).  If it's hard to do, can be postponed to after release

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatView.kt
```diff
@@ -1518,6 +1537,22 @@ private fun ButtonRow(horizontalArrangement: Arrangement.Horizontal, content: @C
   }
 }
 
+@Composable
+fun DateSeparator(date: Instant) {
```
avently, line 1541: can be private

avently:
Chat bubble is jumping now when any checkmark appears (delivery receipt). Is it expected?

avently:
Didn't check correctness of `showTimestamp` param passed everywhere, I hope you did it.
Those two issues with padding in selection UI and not preserved space are the only issues I noticed. Looks good

PR #4927, opened 2024-09-23
ios: hide user picker sheet instantly, when opening another sheet

levitatingpineapple:


apps/ios/Shared/Views/NewChat/NewChatView.swift
```diff
@@ -693,6 +693,7 @@ struct ScannerInView: View {
                             .frame(maxWidth: .infinity, maxHeight: .infinity)
                             .foregroundColor(Color.clear)
                         switch cameraAuthorizationStatus {
+                        case .authorized, nil: EmptyView()
```
levitatingpineapple, line 696: Avoids flashing "Tap to scan" button, if user has already granted camera permission.

PR #4939, opened 2024-09-25
android, desktop: scrolling user profiles

diogofcunha:


avently:
Overall looks good to me. Pushed some not-important changes.
I would say that in landscape mode on Android it would be better to show multiple profiles instead of one long profile. Screen is very long, it should help with profile switching if more profiles can be visible at once

PR #4946, opened 2024-09-27
android, desktop: make space on chat bubble end consistent

diogofcunha:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/item/CIMetaView.kt
```diff
@@ -125,9 +125,9 @@ fun reserveSpaceForMeta(
   showViaProxy: Boolean = false,
   showTimestamp: Boolean
 ): String {
-  val iconSpace = "    "
-  val whiteSpace = " "
-  var res = iconSpace
+  val iconSpace = "\u00A0\u00A0\u00A0\u00A0"
```
diogofcunha, line 128: Rendering without non breaking spaces was ignoring all spaces when rendered inside priority layout (Image attachments, video...)

PR #4951, opened 2024-09-29
android, desktop: add floating date separator to chatview

diogofcunha:


avently:
Date will not gonna hide if you don't scroll in ChatView (desktop). Choose a chat, keep it as is without scrolling. Data should disappear but it is not

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatView.kt
```diff
@@ -1345,6 +1348,24 @@ fun BoxWithConstraintsScope.FloatingButtons(
         firstItemIsVisible = firstVisibleIndex == 0
       }
   }
+  LaunchedEffect(listState) {
+    snapshotFlow { listState.layoutInfo.visibleItemsInfo }
+      .collect { visibleItemsInfo ->
+        if (visibleItemsInfo.find { it.index == 0 } != null) {
```
avently, line 1354: if you need to do something like this, you can user .filter() on snapshotFlow before .collect()

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatView.kt
```diff
@@ -1497,6 +1541,78 @@ private fun TopEndFloatingButton(
   }
 }
 
+@Composable
+private fun TopCenterFloatingButton(
```
avently, line 1545: it's not a button. It's maybe FloatingDate but for sure not a button

diogofcunha:
> ```kotlin
> index
> ```

Never happened to me but I guess it could be a race condition in the updates of all the indexes. I will make the code more defensive

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatView.kt
avently: and since we have only one floatingDate, there is no need to call it `TopCenter`. For floating buttons it makes sense because there are two buttons in ChatView

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatView.kt
```diff
@@ -1345,6 +1348,24 @@ fun BoxWithConstraintsScope.FloatingButtons(
         firstItemIsVisible = firstVisibleIndex == 0
       }
   }
+  LaunchedEffect(listState) {
```
avently, line 1351: Make sure that you really need to put listState here. Isn't `LaunchedEffect(Unit)` is enough?

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatView.kt
```diff
@@ -1401,6 +1422,29 @@ fun BoxWithConstraintsScope.FloatingButtons(
...
+      if (lastIndexOfVisibleItems >= 0) {
+        val lastVisibleChatItemIndex = chatItems.value.lastIndex - firstVisibleIndex - lastIndexOfVisibleItems
+        chatItems.value[lastVisibleChatItemIndex]
+      } else {
+        null
+      }
+    }
+  }
+
+  val firstVisibleItemScrollOffset by remember {
```
avently, line 1437: such solution makes at least two functions to recompose on every scroll. Have you checked the `price` we pay for that solution? 
To check, put this and try to scroll:

```kotlin

SideEffect {
println("LALAL recompose $firstVisibleItemScrollOffset")
}
```

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatView.kt
```diff
@@ -1497,6 +1541,78 @@ private fun TopEndFloatingButton(
...
+    } else if (showDate.value) {
+      showDate.value = false
+    }
+  }
+
+  KeyChangeEffect(firstVisibleItemScrollOffset) {
+    hideDateWhenNotScrolling?.cancel()
+
+    hideDateWhenNotScrolling = coroutineScope.launch {
+      delay(1000)
```
avently, line 1573: If you need a delay, why not to add it before `hideDateWhenNotScrolling = coroutineScope.launch {`?
Because if the coroutine gets canceled before reaching the delay, it still try to display date which is not needed at that point of time, probably

avently:
Please, remove the whole code from FloatingButtons and make separate function for all this date-related things. 
`FloatingButtons` should be fixed later (by me) because it's design is bad and makes more recomposes than needed. By separating your code, the code you write will produce recompositions only for your use-case, not for floating buttons.

avently:
Currently, if you choose another chat on desktop, the floating date updates after one frame. So you see new chat's items but the date update will happen on next drawing frame, looks like. Try to make it instant if possible. Which means to depend more on derivedState than on LaunchedEffect/KeyChangeEffect. And also check recompositions to make sure you don't request to recompose all scrolling movements

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatView.kt
avently: I was wrong, I understood your idea incorrectly. Reverted your commit to return that old behaviour and simplified it a little here: 
https://github.com/simplex-chat/simplex-chat/pull/4951/commits/b213e34df0c324d578f63cee53951393cafd1d33

avently:
The floating date is a bit off placed compared to sticky date you made before. Scroll to the first message in converstation and you'll notice that floating date is lower than fixed date. Maybe some content padding should be added at top part of ChatView column? To make both dates to appear in the same place

avently:
When the floating date is close to the previous message from different date, it shows that different date while fixed date shows the date for the next messages (correct).

So, 
floating date: 27 september
fixed date underneath of floating date: 28 september

Try to scroll app until you see fixed date in the same spot as floating date and compare them. They should show the same data

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatView.kt
```diff
@@ -1497,6 +1504,108 @@ private fun TopEndFloatingButton(
   }
 }
 
+@Composable
+private fun FloatingDate(
+  modifier: Modifier,
+  listState: LazyListState,
+) {
+  var nearBottomIndex by remember { mutableStateOf(-1) }
```
avently, line 1512: if you put it here you'll notice that recompositions happen every scroll offset changes, this shouldn't happen at all
```kotlin
SideEffect { println("LALAL RELOAD") }
```

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatView.kt
```diff
@@ -1497,6 +1504,108 @@ private fun TopEndFloatingButton(
...
+  var isNearBottom by remember { mutableStateOf(true) }
+  val lastVisibleItemDate = remember(listState.layoutInfo.visibleItemsInfo.lastIndex, listState.firstVisibleItemIndex, chatModel.chatItems) {
```
avently, line 1514: try to remove everything in `remember(...)` to `remember {}` and see if it reduces recompositions and gives correct date at the same time

avently: maybe to keep chatModel.chatItems there if it will not work correctly without it

PR #4961, opened 2024-10-01
core: add support for server operators

epoberezkin:


src/Simplex/Chat/Controller.hs
```diff
@@ -948,6 +957,12 @@ data AProtoServersConfig = forall p. ProtocolTypeI p => APSC (SProtocolType p) (
 
 deriving instance Show AProtoServersConfig
 
+data UserServersError
+  = USEStorageMissing
+  | USEProxyMissing
+  | USEDuplicate {server :: AProtoServerWithAuth}
```
spaced4ndy, line 963: can also be invalid name - not parsable

src/Simplex/Chat/Operators.hs
```diff
@@ -0,0 +1,86 @@
...
+    "simplex" -> Just OTSimplex
+    "xyz" -> Just OTXyz
+    _ -> Nothing
+  textEncode = \case
+    OTSimplex -> "simplex"
+    OTXyz -> "xyz"
+
+data ConditionsAcceptance
+  = CAAccepted {date :: UTCTime}
+  | CARequired {deadline :: Maybe UTCTime, acceptedConditionsCommit :: Maybe Text}
```
spaced4ndy, line 61: instead of having acceptedConditionsCommit on operator better to return old and new conditions (possibly only hash for old) in CRUsageConditions

src/Simplex/Chat/Operators.hs
```diff
@@ -0,0 +1,86 @@
...
+  deriving (Show)
+
+data ServerOperator = ServerOperator
+  { operatorId :: OperatorId,
+    operatorTag :: Maybe OperatorTag,
+    tradeName :: Text,
+    legalName :: Maybe Text,
+    acceptedConditions :: Maybe ConditionsAcceptance,
```
spaced4ndy, line 69: why is it optional? if no conditions are accepted, backend would map it into required for current conditions, no? how to interpret no data about conditions in UI?

src/Simplex/Chat/Migrations/chat_schema.sql
```diff
@@ -589,6 +590,23 @@ CREATE TABLE note_folders(
...
+  trade_name TEXT NOT NULL,
+  legal_name TEXT,
+  enabled INTEGER NOT NULL DEFAULT 1,
+  role_storage INTEGER NOT NULL DEFAULT 1,
+  role_proxy INTEGER NOT NULL DEFAULT 1
+);
+CREATE TABLE usage_conditions_acceptance(
+  usage_conditions_acceptance_id INTEGER PRIMARY KEY AUTOINCREMENT,
+  server_operator_id INTEGER REFERENCES server_operators(server_operator_id),
+  conditions_required_at TEXT NOT NULL,
```
spaced4ndy, line 605: we may need to set it in addtional api actually, because we don't know when conditions will be shown to user. so this would be nullable

src/Simplex/Chat/Controller.hs
```diff
@@ -352,6 +353,11 @@ data ChatCommand
   | SetUserProtoServers AProtoServersConfig
   | APITestProtoServer UserId AProtoServerWithAuth
   | TestProtoServer AProtoServerWithAuth
+  | APIGetUserServers UserId
+  | APISetUserServers UserId (NonEmpty UserServers)
+  | APIValidateServers (NonEmpty UserServers) -- response is CRUserServersValidation
+  | APIGetUsageConditions
+  | APIAcceptConditions UTCTime (NonEmpty OperatorId)
```
spaced4ndy, line 360:
```suggestion
  | APIGetUsageConditions
  | APIUsageConditionsShown UTCTime
  | APIAcceptConditions UTCTime (NonEmpty OperatorId)
```
- to set required_at (comment in schema)

src/Simplex/Chat/Migrations/chat_schema.sql
```diff
@@ -589,6 +590,23 @@ CREATE TABLE note_folders(
...
+  conditions_accepted_at TEXT,
+  conditions_commit TEXT NOT NULL,
+  accepted_at TEXT NOT NULL
```
spaced4ndy, line 608: conditions_accepted_at, accepted_at - duplicate?

src/Simplex/Chat/Migrations/M20241027_server_operators.hs
```diff
@@ -0,0 +1,55 @@
...
+  trade_name TEXT NOT NULL,
+  legal_name TEXT,
+  enabled INTEGER NOT NULL DEFAULT 1,
+  role_storage INTEGER NOT NULL DEFAULT 1,
+  role_proxy INTEGER NOT NULL DEFAULT 1
+);
+
+ALTER TABLE protocol_servers ADD COLUMN server_operator_id INTEGER REFERENCES server_operators ON DELETE SET NULL;
+
+CREATE TABLE usage_conditions_acceptance (
```
spaced4ndy, line 23: why change from previous many-to-many schema? I think it made sense:

adding specific conditions (without text but with commit) only once, linking operators to that single record

PR #4963, opened 2024-10-01
load sheet navigation titles before presenting

levitatingpineapple:


apps/ios/Shared/Views/RemoteAccess/ConnectDesktopView.swift
```diff
@@ -14,7 +14,6 @@ struct ConnectDesktopView: View {
     @EnvironmentObject var m: ChatModel
     @EnvironmentObject var theme: AppTheme
     @Environment(\.dismiss) var dismiss: DismissAction
-    var viaSettings = false
```
levitatingpineapple, line 17: no longer used

PR #4970, opened 2024-10-02
desktop: fix typescript sdk ability to send / receive messages

Pixelycia:
As far I understand, recently you could send one message at once, now it changed and SDK wasn't able to handle properly message sending \ receiving (it just crushes). I fix it, so instead of sending \ receiving single message now you send \ receive array of messages at once


Pixelycia:

epoberezkin:
Thank you!

PR #4977, opened 2024-10-04
core: add chat message tail and roundness settings

diogofcunha:


src/Simplex/Chat/AppSettings.hs
```diff
@@ -50,6 +50,8 @@ data AppSettings = AppSettings
     iosCallKitEnabled :: Maybe Bool,
     iosCallKitCallsInRecents :: Maybe Bool,
     uiProfileImageCornerRadius :: Maybe Double,
+    uiChatItemRoundness :: Maybe Double,
```
spaced4ndy, line 53: these also seem missing in AppSettings.swift in master

PR #4980, opened 2024-10-04
ios: remove `showSettings` and `withNavigation`

levitatingpineapple:


apps/ios/Shared/ContentView.swift
```diff
@@ -86,7 +85,7 @@ struct ContentView: View {
                 callView(call)
             }
 
-            if !showSettings, let la = chatModel.laRequest {
+            if let la = chatModel.laRequest {
```
levitatingpineapple, line 88: Removal of `!showSettings`

Is this condition still required?
If so - In which flow is it needed?

epoberezkin: it changes where authentication view is attached I think - it is used both to lock the whole app and to lock, for example, chat profiles, console, stopping the chat that open inside settings.

PR #4988, opened 2024-10-07
rfc: infinite scrolling

diogofcunha:


docs/rfcs/2024-10-07-chat-infinite-scrolling.md
```diff
@@ -0,0 +1,322 @@
...
+### Pagination integration:
+
+```
+fetch(scrollingUp, chatItemId)
+    if scrollingUp
+        fetchBefore(chatItemId)
+    else
+        fetchAfter(chatItemId)
+
+-- Another option would be if the api exposed a way to do before and after, possibly already supported since they are just separate query strings
```
spaced4ndy, line 32: it is already supported, yes

spaced4ndy: current approach to pagination in backend (based on ts, database id) is problematic though, see comments below - maybe we should rework backend simultaneously with this.
https://github.com/simplex-chat/simplex-chat/pull/3677#issuecomment-1889326757
https://github.com/simplex-chat/simplex-chat/pull/3661#issuecomment-1884824486

diogofcunha: For my understanding, is the major issue we face that IDs are not stable and therefore results of pagination may not be immutable? (Ex if I hit the same page twice totally different results may be retrieved)

diogofcunha: How does the issue manifest today for users? Duplicates? Missing messages?

diogofcunha: I guess we also have no way of knowing if a permanent delete happens, since id may be reused

docs/rfcs/2024-10-07-chat-infinite-scrolling.md
```diff
@@ -38,6 +38,15 @@ fetchLatest()
     fetchLast()
 ```
 
+
+```haskell
+-- Chat command
+APIGetInitialChatItems ChatRef
+
+-- Chat result
+CRInitialChatItems {user :: User, latestChatItems :: [AChatItem], unreadChatItems :: [AChatItem]}
```
diogofcunha, line 47: could also just be part of the API we currently use for loading a chat, if we can safely make breaking changes

docs/rfcs/2024-10-07-chat-infinite-scrolling.md
spaced4ndy: There's multiple issues arising out of it, the main reason for all of them is that IDs are not guaranteed to be monotonically increasing, because deleted records free up old IDs that can be reused.

So for a given item ID is stable, but sorting based on ID doesn't guarantee correct item order. So we sort on ts first*, then ID. Using ID for pagination cutoff is also incorrect (as done for contact and local chat queries currently; groups use ts instead - see `getDirectChat`, `getGroupChat`, `getLocalChat` pagination queries). For example, you search for items before item X, but it reused old ID - current query would incorrectly filter out items with greater IDs that really are before item X chronologically. At the very least contact and local queries should be reworked similar to group queries, but it wouldn't solve all problems.

* Using ts is also not perfect, depending on where it originated: server ts used for sorting received items in groups are rounded to second, so messages that have same ts on server may be incorrectly sorted on ID fallback; local ts is unreliable because it depends on time set on user device, which user can change arbitrarily. For example, if user sets device time forward 10 mins, in a group all sent messages are created with it, received messages are created with server time; conversation happens, and re-opening chat shows only sent items last (it looks like received items have disappeared unless you scroll up - it's not a conjecture, it practically happened to a user).

How it can manifest:
- possible chronologically incorrect item ordering / different item ordering on chat reload from one when you are inside chat and receiving events
- possible skips on pagination (even cutting off based on ts does not guarantee no skips I think in case items had same ts and were sorted on ID)
- messages fail to be marked as read (via range api)

What we need to solve all these issues is some counter tables with monotonically increasing counters per conversation, linking back to chat items.

docs/rfcs/2024-10-07-chat-infinite-scrolling.md
```diff
@@ -0,0 +1,331 @@
...
+**Pros:**
+- Keeping two sections makes it simple for users to scroll through messages and jump to recent ones and doesn't add extreme complexity in code design.
+- Most common action for users (going back to the bottom of the chat) are totally addressed and instant without exponetially increasing the ammount of data stored in device runtime memory.
+
+**Cons:**
+- Managing two sections can make the code harder to maintain and could introduce bugs if not handled carefully.
+- Not reusing anchors means the app might have to access the database more often.
+- Deduplication of items as they are fetched becomes of extreme importance as not doing it effectively can lead to duplicated messages being displayed.
+
+### Option 3
```
avently, line 177: This is how I thought to implement that feature. The difference is that you can't "swap" sections because in real code you should scroll to that item in question. Because "swapping" would make the whole list of items changed in one frame, this is not what a user expect.

So the idea to have multiple sections is good, but all of them should be rendered in LazyColumn using items{} blocks. Datasource should provide all available sections (taking care of intersections of items in sections). 

When new random item is loaded (by searching or clicking on reply), that item has preceding items and following items. They become a section if none of the items have intersections between existing sections. If they have, they are merged into such section. If the items have intersections with more than one section, non-duplicated items are added to the closest section and duplicated items (from next sections) just dropped (so the next sections are not merged together. I think having more small sections is better than having one big section because you can drop some of them when memory exhausted).

Datasource should provide active chat item key (which is Pair<chatItemId,chatItem.createdAt> currently) when you want to scroll somewhere so LazyColumn would know what item to scroll to. After LazyColumn scrolled to that item, active chat item key should become null (no need to track action chat item after initial scroll to it).

Important part which should be covered by that design is handling of events (the ones that can be expanded/collapsed in UI). All events should be put into separate sections. Let's say you found one event chat item without preceding/following events. It has to be put into separate section too. Because next time when you load chat item it can be event too and you will put that new item into that section. 

So you will have at least two types of sections: ordinary chat items and event chat items. Items with different types cannot be in one section. Why? Because by putting them in different buckets you can easily implement "reveal" functionality which will not depend on running in composition like it is now (which gives a lot of unneeded recompositions) but instead will rely on pre-populated data in the datasource (model). That special "events" section can have all required elements to construct dependant event items.

The whole idea look complex, yes, but it gives enough flexibility to render anything, even section headers can be put here to make main thread code less compute-intensive.

avently: before implementing this concept, check that having multiple sections like this can be done in LazyColumn without side effects like recomposition of unchanged sections. Because items{} block located out of composition and should be constructed dynamically like 
listOfSections.forEach( section -> items(section, key=...){} }

I'm not sure that it can be done easily or done at all without sideeffects

docs/rfcs/2024-10-07-chat-infinite-scrolling.md
epoberezkin: These are all arguments for causality-based sorting really.

docs/rfcs/2024-10-07-chat-infinite-scrolling.md
epoberezkin: Scrolling is a rather strong argument to keep up to 3 sections: bottom, current (can be the same as bottom), destination of scroll (also can be the same as bottom).

Also the third section is needed only temporarily during the far navigation, while simply scrolling would add items to the top of the current section and remove them from the bottom of it.

This way, when navigating from second section upwards, an additional third section will be added above current and scroll triggered. Once scroll completes the below section can be removed, and the third section becomes current.

There is no reason to generalise it to N, and it's easier to code the logic around 3, with removing unneeded sections.

So we would have to cater for these scenarios:

1. Slow scroll up when the current section is bottom

At some point when N items is loaded (N is max items per section, say, 500), the current section will be split to two sections - current and bottom with M items in each (M is min items per section, say 200).

In parallel items can be added to either of the sections, when added to the bottom, the bottom section will start dropping items on top of it.

2. Far scroll up when the current section is bottom.

Destination section is loaded with M items, with destination item being in the centre of it. This section is added to top, and scroll is triggered (so it looks like scroll).

3. Slow scroll upwards when current section is not bottom.

Add items at the direction of scroll in steps of L (say 50), start dropping items from another end as the section exceeds N (500) - drop enough for it to become M (200). It's the basic principles of all regulating apparata to have two thresholds and only invoke a corrective action once threshold is reached. While 200 and 500 can be tweaked to find the correct balance, it is necessary to have min/max thresholds for loaded section size.

4. Slow scroll downards when current section is not bottom.

Same as upwards, but it should cater for the scenario when the current section and bottom section merge. That suggests that max size for the bottom section should be bigger than max size for other sections – say K, that logically should be N + M.

5. Far scroll upwards when the current section is not bottom.

Same as far scroll from the bottom section, with the difference that once scroll is complete the current section will be dropped.

6. Far scroll downwards when the current section is not bottom.

The first is to make the decision whether we will be in bottom section or it will be another section above bottom. The decision could be simple - if the destination item is higher than N + M/2 we should create a new section. If the destination item is lower than that, it should extend bottom section to N + M items and scroll to destination.

Other considerations:

All these arguments assume that for UI to work we have to have a single array of items. This is not necessarily optimal from addition/removal points of view.

A posible alternative to have exactly 3 arrays, titled as Bottom, Current (optional), Destination (optional), + indication whether destination is up/down or bottom during the scroll, and during view rendering we can use all arrays in the correct order. This is certainly possible in Android, where we just render items in a look (cc @avently) but may be harder in iOS where we may need a single array (@levitatingpineapple or don't we?)

epoberezkin: Maybe we can just simulate scrolling from one view to another by rendering lists of lists and reordering them as necessary:

```
ChatView {
  if let upDestBlock {
    ForEach(upDestBlock) { item }
  }
  if let currBlock {
    ForEach(currBlock) { item }
  }
  if let downDestBlock {
    ForEach(downDestBlock) { item }
  }
  ForEach(bottomBlock) { item }
}
```

Why the above can't work? I mean, what's the reason to need N section instead of 4 with max 3 being rendered at a given time, and max 2 once scrolling completes?

Programming the same logic around the arbitrary number of sections seems more complex.

epoberezkin: that also suggests that when chat is opened 1 or 2 blocks may be loaded - only bottom, in case the last unread is not higher than N + M/2, or bottom (M items) and current (M items), with the last unread in the centre of the current block.

epoberezkin: Merged items will just increase the size of the current section in case they are expanded - the worst being "connected" that we need to get read of somehow.

avently: > Merged items will just increase the size of the current section in case they are expanded - the worst being "connected" that we need to get rid of somehow.

Without having a solution for this problem your arguments (about 3 sections) does not have a sense.

diogofcunha: I think we could take some learnings from signal here, each sort of merges your two approaches. 

I think it’s ok to have areas for let’s call it “threads” where we load those interaction areas on click for expand, so they would never load on the list itself, but rather on a bottom sheet or a similar primitive in android. We could also load them on the main list but they would need to be their separate area each (stas is saying I think). But anyway I think we should stop aggregating this items client side, we want page while scrolling fetches to be as fast as possible

avently: https://github.com/simplex-chat/simplex-chat/tree/av/multiplatform-test-chatview

Small demonstration of the idea from my comments. It works in terms of displaying groups and updating data in it dynamically. So the whole concept in my comments can be applied without problems. Even more than that: chat items can be stored in one list like we do now. Then every update require one single pass of the whole list to put each items to sections. It's pretty fast even on old device. Around 100 items per one millisecond (32 ms per 3200 items) on cold start. On continue using the app and opening chats the time reduces to 0.3-0.5 ms per 100 items. So it's not important at all to optimize that thing. Couple of milliseconds, even if for some devices it could be 10, it's not noticeable and will only happen on data refresh (like arriving new item or scrolling to top, etc). If we skip all optimizations we thought about, the whole structure becomes veeeery easy to understand and to code/maintain. Of course there is no need to store 3200 items like in my example, unneeded items can removed automatically on each pass.

avently: On desktop result of forEach'ing 3200 items is 7 ms :) So I think we need to focus on correctness, not on performance. And take just the simplest thing: chatItems list like we have now, the whole logic of loading new items and putting them into one list like we have now, putting all this into buckets, displaying them in sections, reducing size of list from unneeded items

PR #5008, opened 2024-10-09
ios: fix group member sheet load animation

levitatingpineapple:


apps/ios/Shared/Views/Chat/Group/GroupMemberInfoView.swift
```diff
@@ -253,6 +153,114 @@ struct GroupMemberInfoView: View {
         }
         .modifier(ThemedBackground(grouped: true))
     }
+    
+    @ViewBuilder
+    private func sections(_ member: GroupMember) -> some View {
```
levitatingpineapple, line 158: moved without changes

epoberezkin: why does it need to move?

apps/ios/Shared/Views/Chat/Group/GroupMemberInfoView.swift
```diff
@@ -94,129 +95,28 @@ struct GroupMemberInfoView: View {
...
                     // this condition prevents re-setting picker
                     if !justOpened { return }
                 }
                 justOpened = false
-                DispatchQueue.main.async {
-                    newRole = member.memberRole
-                    do {
-                        let (_, stats) = try apiGroupMemberInfo(groupInfo.apiId, member.groupMemberId)
-                        let (mem, code) = member.memberActive ? try apiGetGroupMemberCode(groupInfo.apiId, member.groupMemberId) : (member, nil)
+                newRole = member.memberRole
```
epoberezkin, line 107: this is state change on background thread?

apps/ios/Shared/Views/Chat/Group/GroupMemberInfoView.swift
epoberezkin: without moving, it weirdly works differently...

PR #5016, opened 2024-10-10
ios: fix empty qr code reader when swapping to connect via link

diogofcunha:
Tested when permission is enabled and disabled, it replicates the behaviour that happens when this view is open directly from new chat sheet

### DIFF
--- a/apps/ios/Shared/Views/NewChat/NewChatView.swift
+++ b/apps/ios/Shared/Views/NewChat/NewChatView.swift
@@ -97,6 +97,11 @@ struct NewChatView: View {
             }
             .pickerStyle(.segmented)
             .padding()
+            .onChange(of: $selection.wrappedValue) { opt in
+                if opt == NewChatOption.connect {
+                    showQRCodeScanner = true
+                }
+            }
 
             VStack {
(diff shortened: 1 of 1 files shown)


apps/ios/Shared/Views/NewChat/NewChatView.swift
```diff
@@ -97,6 +97,11 @@ struct NewChatView: View {
             }
             .pickerStyle(.segmented)
             .padding()
+            .onChange(of: $selection.wrappedValue) { opt in
```
spaced4ndy, line 100: I think old behavior when swiping from "new link" to "scan link" was to show scan button, and not immediately scanner (camera) itself.. but it's probably ok

diogofcunha: on one hand seems like a good idea, on the other hand this is more consistent with what happens when you open it directly so not totally sure

diogofcunha: i would say that if you go there is a lot more likely that you want to scan then to paste the link (since paste is supported everywhere else) and we would be adding an extra tap for the most common action

PR #5018, opened 2024-10-10
core: do not regenerate key when accepting connection to avoid invalidating invitation link on bad networks

spaced4ndy:


src/Simplex/Chat.hs
```diff
@@ -1297,14 +1297,37 @@ processChatCommand' vr = \case
...
+        Just conn -> do
+          let profileToSend = profileToSendOnAccept user incognitoProfile False
+          dm <- encodeConnInfoPQ pqSupport chatV $ XInfo profileToSend
+          (_, sqSecured) <- withAgent $ \a -> acceptContact a (aConnId conn) True invId dm pqSupport subMode
+          ucl <- withFastStore $ \db -> getUserContactLinkById db userId userContactLinkId
+          let contactUsed = (\(_, groupId_, _) -> isNothing groupId_) ucl
+              connStatus_ = if sqSecured then Just ConnSndReady else Nothing
+          ct' <- withStore $ \db -> do
+            liftIO $ deleteContactRequestByName db user cName
+            updateContactAccepted db user ct connStatus_ contactUsed
```
epoberezkin, line 1328: this probably can accept conn as parameter to avoid additional error

src/Simplex/Chat/Store/Direct.hs
```diff
@@ -779,29 +780,42 @@ createAcceptedContact db user@User {userId, profile = LocalProfile {preferences}
...
+            chatTs = Just createdAt,
+            contactGroupMemberId = Nothing,
+            contactGrpInvSent = False,
+            uiThemes = Nothing,
+            chatDeleted = False,
+            customData = Nothing
+          }
+  pure (ct, conn)
+
+deleteContactRequestByName :: DB.Connection -> User -> ContactName -> IO ()
```
epoberezkin, line 809: why is it using name and not ID?

apps/ios/Shared/Model/SimpleXAPI.swift
```diff
@@ -462,7 +462,7 @@ func apiGetNtfToken() -> (DeviceToken?, NtfTknStatus?, NotificationsMode, String
     let r = chatSendCmdSync(.apiGetNtfToken)
     switch r {
     case let .ntfToken(token, status, ntfMode, ntfServer): return (token, status, ntfMode, ntfServer)
-    case .chatCmdError(_, .errorAgent(.CMD(.PROHIBITED))): return (nil, nil, .off, nil)
+    case .chatCmdError(_, .errorAgent(.CMD(.PROHIBITED, "_"))): return (nil, nil, .off, nil)
```
spaced4ndy, line 465:
```suggestion
    case .chatCmdError(_, .errorAgent(.CMD(.PROHIBITED, _))): return (nil, nil, .off, nil)
```

PR #5022, opened 2024-10-11
android: re-apply custom language when webview appears

avently:
There is a bug on Android related to including WebView. App language changes to system language regardless of what was set before in context's configuration. Re-set needed to be done twice: after creating of WebView and after removing it from a view

https://issuetracker.google.com/issues/109833940


apps/multiplatform/common/src/androidMain/kotlin/chat/simplex/common/views/call/CallView.android.kt
```diff
@@ -775,6 +778,15 @@ private fun updateActiveCall(initial: Call, transform: (Call) -> Call) {
   }
 }
 
+/*
+* Creating WebView automatically drops user's custom app locale to default system locale.
+* Preventing it by re-applying custom locale
```
epoberezkin, line 783:
```suggestion
* Preventing it by re-applying custom locale
* https://issuetracker.google.com/issues/109833940
```

PR #5047, opened 2024-10-15
core: ntf get many wip

spaced4ndy:


apps/ios/SimpleX NSE/NotificationService.swift
```diff
@@ -219,51 +266,87 @@ class NotificationService: UNNotificationServiceExtension {
...
         deliverBestAttemptNtf(urgent: true)
     }
 
-    func processReceivedNtf(_ ntf: NSENotification) -> Bool {
-        guard let ntfInfo = notificationInfo, let expectedMsgTs = ntfInfo.expectedMsg_?.msgTs else { return false }
+    // TODO
+    // for cases where we couldn't process notification:
+    // - return `!expectedMessages.allSatisfy { $0.value.ready }` instead of false?
+    //   (so that outside loop goes on to expected messages from other connections)
+    // - or instead of waiting in outside loop, do one more attempt, if another matching thread is not found proceed?
```
spaced4ndy, line 299: Currently just returning `!expectedMessages.allSatisfy { $0.value.ready }`, which means message loop continues while there are not ready "expected message" entries, and that messages previously returning false do not get processed by other threads if they exist. Simply returning false is not correct because then if there are no other matching threads loop will be stuck on waiting for a single message, and batch of expected messages would not finish processing.

Couple of options:
- We could change return to some enum instead of bool signalling "go onto next", "limited retry" (e.g. when not all expected messages are ready - try once or a few more times to find another thread, but shorter than current wait limit), "wait" (when all expected messages are ready - ok to wait as before).
- We could put these rejected messages to some other queue where they would be waiting for matching threads as before, without holding processing of this batch of expected messages.

apps/ios/SimpleX NSE/NotificationService.swift
```diff
@@ -219,51 +266,82 @@ class NotificationService: UNNotificationServiceExtension {
...
+        guard
+            let expectedMessage = expectedMessages[id],
+            let expectedMsgTs = expectedMessage.ntfMessage.expectedMsg_?.msgTs
+        else {
+            return !expectedMessages.allSatisfy { $0.value.ready }
+        }
+        let ntfInfo = expectedMessage.ntfMessage
         if !ntfInfo.user.showNotifications {
-            self.setBestAttemptNtf(.empty)
+            expectedMessages[id]?.msgBestAttempt = .empty
```
spaced4ndy, line 304: Currently just returning `!expectedMessages.allSatisfy { $0.value.ready }`, which means message loop continues while there are not ready "expected message" entries, and that messages previously returning false do not get processed by other threads if they exist. Simply returning false is not correct because then if there are no other matching threads loop will be stuck on waiting for a single message, and batch of expected messages would not finish processing (or would take longer than necessary).

Couple of options:
- We could change return to some enum instead of bool signalling "go onto next", "limited retry" (e.g. when not all expected messages are ready - try once or a few more times to find another thread, but shorter than current wait limit), "wait" (when all expected messages are ready - ok to wait as before).
- We could put these rejected messages to some other queue where they would be waiting for matching threads as before, without holding processing of this batch of expected messages.

apps/ios/SimpleX NSE/NotificationService.swift
```diff
@@ -219,51 +266,82 @@ class NotificationService: UNNotificationServiceExtension {
...
             } else if let msgTs = info.msgTs_, msgTs > expectedMsgTs {
-                logger.debug("NotificationService processNtf: msgInfo msgId = \(info.msgId, privacy: .private): unexpected msgInfo, let other instance to process it, stopping this one")
+                logger.debug("NotificationService processNtf: msgInfo msgId = \(info.msgId, privacy: .public): unexpected msgInfo, let other instance to process it, stopping this one")
+                expectedMessages[id]?.ready = true
                 self.deliverBestAttemptNtf()
-                return false
-            } else if allowedGetNextAttempts > 0, let receiveConnId = receiveConnId {
-                logger.debug("NotificationService processNtf: msgInfo msgId = \(info.msgId, privacy: .private): unexpected msgInfo, get next message")
-                allowedGetNextAttempts -= 1
+                return !expectedMessages.allSatisfy { $0.value.ready }
```
spaced4ndy, line 317: Currently just returning `!expectedMessages.allSatisfy { $0.value.ready }`, which means message loop continues while there are not ready "expected message" entries, and that messages previously returning false do not get processed by other threads if they exist. Simply returning false is not correct because then if there are no other matching threads loop will be stuck on waiting for a single message, and batch of expected messages would not finish processing (or would take longer than necessary).

Couple of options:
- We could change return to some enum instead of bool signalling "go onto next", "limited retry" (e.g. when not all expected messages are ready - try once or a few more times to find another thread, but shorter than current wait limit), "wait" (when all expected messages are ready - ok to wait as before).
- We could put these rejected messages to some other queue where they would be waiting for matching threads as before, without holding processing of this batch of expected messages.

apps/ios/SimpleX NSE/NotificationService.swift
```diff
@@ -79,7 +124,7 @@ class NSEThreads {
 
     private func rcvEntityThread(_ id: ChatId) -> (UUID, NotificationService)? {
         NSEThreads.queue.sync {
-            activeThreads.first(where: { (_, nse) in nse.receiveEntityId == id })
+            activeThreads.first(where: { (_, nse) in nse.expectedMessages[id]?.ready == false })
```
epoberezkin, line 127:
```suggestion
            // this selects the first thread that:
            // 1) has this connection in nse.expectedMessages
            // 2) has not completed processing messages for this connection
            activeThreads.first(where: { (_, nse) in nse.expectedMessages[id]?.ready == false })
```

apps/ios/SimpleX NSE/NotificationService.swift
```diff
@@ -192,116 +242,130 @@ class NotificationService: UNNotificationServiceExtension {
...
-                        shouldProcessNtf = true
-                        return
-                    }
+               let ntfMessages = apiGetNtfMessage(nonce: nonce, encNtfInfo: encNtfInfo) {
+                logger.debug("NotificationService: receiveNtfMessages: apiGetNtfMessage ntfMessages count = \(ntfMessages.count)")
+                for ntfMessage in ntfMessages {
+                    addExpectedMessage(ntfMessage: ntfMessage)
+                }
+                if !expectedMessages.isEmpty {
+                    shouldProcessNtf = true
```
spaced4ndy, line 251:
```suggestion
                if !expectedMessages.isEmpty {
                    // shouldProcessNtf = false
                    // process rejected messages queue
                    // don't wait in `processNotification`
                    shouldProcessNtf = true
```

PR #5061, opened 2024-10-16
ios: fix screen protect for sheets

levitatingpineapple:


levitatingpineapple:
https://github.com/simplex-chat/simplex-chat/pull/5062

Issue #1673, opened 2023-01-01
dyld[1204]: Library not loaded: /usr/local/opt/openssl@1.1/lib/libcrypto.1.1.dylib


adamshand:
I just downloaded simplex chat to give the terminal client a try.  On most recent macOS I get this error:

❯ uname -a
Darwin chimera.local 22.2.0 Darwin Kernel Version 22.2.0: Fri Nov 11 02:04:44 PST 2022; root:xnu-8792.61.2~4/RELEASE_ARM64_T8103 arm64

❯ ~/bin/darwin/simplex-chat
dyld[1204]: Library not loaded: /usr/local/opt/openssl@1.1/lib/libcrypto.1.1.dylib
  Referenced from: <54D5EF15-53B9-351E-B65E-5A3B45502488> /Users/adam/bin/darwin/simplex-chat
  Reason: tried: '/usr/local/opt/openssl@1.1/lib/libcrypto.1.1.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/usr/local/opt/openssl@1.1/lib/libcrypto.1.1.dylib' (no such file), '/usr/local/opt/openssl@1.1/lib/libcrypto.1.1.dylib' (no such file), '/usr/local/lib/libcrypto.1.1.dylib' (no such file), '/usr/lib/libcrypto.1.1.dylib' (no such file, not in dyld cache)
zsh: abort      ~/bin/darwin/simplex-chat

nicfab:
I have the same issue.
macOS Ventura 13.1
I hope it will be solved.

epoberezkin:
Are you setting up cabal.project.local as described in this doc: https://github.com/simplex-chat/simplex-chat/blob/stable/docs/CLI.md#in-any-os

Specifically, see the comments in the script related to Mac OS.

The most likely reason it cannot find openssl is because you didn't do one of the steps in the doc (install it with brew, add cabal.project.local, make sure it points to correct openssl location).

nicfab:
> Are you setting up cabal.project.local as described in this doc: https://github.com/simplex-chat/simplex-chat/blob/stable/docs/CLI.md#in-any-os
> 
> Specifically, see the comments in the script related to Mac OS.
> 
> The most likely reason it cannot find openssl is because you didn't do one of the steps in the doc (install it with brew, add cabal.project.local, make sure it points to correct openssl location).

I did as per your instruction, but after running `cabal install`, I saw the following:

```
[...]

PR #5070, opened 2024-10-18
fix(install.sh): downloading the right binary for apple silicon chip

reply2future:
re #1673


reply2future:

PR #5082, opened 2024-10-21
core: chat ntf get conns

spaced4ndy:


src/Simplex/Chat/Controller.hs
```diff
@@ -1063,11 +1065,16 @@ instance FromJSON ComposedMessage where
...
+data NtfConn = NtfConn
   { user_ :: Maybe User,
     connEntity_ :: Maybe ConnectionEntity,
-    expectedMsg_ :: Maybe NtfMsgInfo,
-    receivedMsg_ :: Maybe NtfMsgInfo
+    expectedMsg_ :: Maybe NtfMsgInfo
+  }
+  deriving (Show)
+
+data NtfMessage = NtfMessage
```
epoberezkin, line 1075: this seems unnecessary

PR #5088, opened 2024-10-23
core: auto increment chat item ids

diogofcunha:


src/Simplex/Chat/Migrations/M20241023_chat_item_autoincrement_id.hs
```diff
@@ -0,0 +1,34 @@
+{-# LANGUAGE QuasiQuotes #-}
+
+module Simplex.Chat.Migrations.M20241023_chat_item_autoincrement_id where
+
+import Database.SQLite.Simple (Query)
+import Database.SQLite.Simple.QQ (sql)
+
+m20241023_chat_item_autoincrement_id :: Query
```
spaced4ndy, line 8: tests currently fail on schema mismatch - need to run them and commit updated schema

spaced4ndy (approved):

PR #5100, opened 2024-10-25
core: pagination API to load items around defined or the earliest unread item

diogofcunha:


src/Simplex/Chat/Store/Messages.hs
```diff
@@ -2650,3 +2837,10 @@ getGroupHistoryItems db user@User {userId} GroupInfo {groupId} count = do
             LIMIT ?
           |]
           (userId, groupId, rcvMsgContentTag, sndMsgContentTag, count)
+
+divideFetchCountAround_ :: Int -> (Int, Int)
```
epoberezkin, line 2841: this is overengineered and shouldn't be a function.
All it does is something like this:

before = count `div` 2
after = count - before

So it has to be called just in place

epoberezkin: to confuse things further, we pass count - 1 to this function

PR #5104, opened 2024-10-26
core: initial landing api for chat and gaps 

diogofcunha:


src/Simplex/Chat/Store/Messages.hs
```diff
@@ -1083,37 +1084,63 @@ getDirectChatAround_ db user ct@Contact {contactId} aroundItemId count search =
   let chatItems = beforeChatItems <> [middleChatItem] <> afterChatItems
   pure $ Chat (DirectChat ct) chatItems stats
 
-getGroupChat :: DB.Connection -> VersionRangeChat -> User -> Int64 -> ChatPagination -> Maybe String -> ExceptT StoreError IO (Chat 'CTGroup)
+getDirectChatInitial_ :: DB.Connection -> User -> Contact -> Int -> ExceptT StoreError IO (Chat 'CTDirect, ChatLandingSection)
+getDirectChatInitial_ db user@User {userId} ct@Contact {contactId} count = do
+  firstUnreadItem <- liftIO getDirectChatMinUnreadItemId_
+  case firstUnreadItem of
```
spaced4ndy, line 1090:
```suggestion
  firstUnreadItemId_ <- liftIO getDirectChatMinUnreadItemId_
  case firstUnreadItemId_ of
```

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -1083,37 +1084,63 @@ getDirectChatAround_ db user ct@Contact {contactId} aroundItemId count search =
...
+    Just firstUnreadItemId -> do
+      chat <- getDirectChatAround_ db user ct firstUnreadItemId count ""
```
spaced4ndy, line 1092: search not passed because chat is always opened without it? probably ok, but could be setting trap for future.. maybe at least worth throwing error in Chat on api call - validate that if it's initial search should be Nothing. or even simpler to pass it here, it shouldn't hurt since in current real scenario it'll be empty

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -1083,37 +1084,63 @@ getDirectChatAround_ db user ct@Contact {contactId} aroundItemId count search =
...
+      let Chat {chatItems} = chat
+      let chatItemIds = map (cchatItemId) chatItems
+      lastItemId <- liftIO $ getDirectChatItemIdsLast_ db user ct 1 ""
+      let lastItemIdInChat = head lastItemId `elem` chatItemIds
```
spaced4ndy, line 1096: safer to pattern match maybe.. though we're already in branch where we know there's at least 1 item.

also probably more efficient to use foldr for single pass over list (instead of map, elem) to find if last item id is in list

spaced4ndy: (we expect it last item id to match end of list)

spaced4ndy (dismissed):

PR #5111, opened 2024-10-28
core, ui: add support for server operators

spaced4ndy:


src/Simplex/Chat/Operators.hs
```diff
@@ -0,0 +1,202 @@
...
+
+data OperatorEnabled = OperatorEnabled
+  { operatorId :: OperatorId,
+    enabled :: Bool,
+    roles :: ServerRoles
+  }
+  deriving (Show)
+
+data UserServers = UserServers
+  { operator :: Maybe ServerOperator,
```
spaced4ndy, line 135:
```suggestion
  { serverOperator :: Maybe ServerOperator,
```
reserved word, simpler to rename

PR #5114, opened 2024-10-29
ios: server operators ui

spaced4ndy:


apps/ios/SimpleXChat/APITypes.swift
```diff
@@ -1194,8 +1194,136 @@ public struct UserProtoServers: Decodable {
...
+        largeLogo: "logo",
+        logoDarkMode: "decentralized-light",
+        largeLogoDarkMode: "logo-light"
+    )
+]
+
+public enum UsageConditionsAcceptance: Decodable, Hashable {
+    case accepted(date: Date)
+    case reviewAvailable(deadline: Date)
+    case reviewRequired
```
epoberezkin, line 1242: it should be case required(deadline: Date?) - you don't have 3 cases here, it's 2.

PR #5115, opened 2024-10-29
android, desktop: infinite loading

diogofcunha:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatView.kt
```diff
@@ -961,21 +989,104 @@ fun BoxScope.ChatItemsList(
     }
   }
 
-  PreloadItems(chatInfo.id, listState, ChatPagination.UNTIL_PRELOAD_COUNT, loadPrevMessages)
-
   Spacer(Modifier.size(8.dp))
   val reversedChatItems = remember { derivedStateOf { chatModel.chatItems.asReversed() } }
+  val revealedItems = rememberSaveable { mutableStateOf(setOf<Long>()) }
+  val sections by remember { derivedStateOf { reversedChatItems.value.putIntoSections(revealedItems.value) } }
```
avently, line 995: try to exclude from you practice using ```val something by ...```. Because when you do this, you cannot safely use that value inside of other Unit blocks because in that Unit blocks that value you're pointing at can be stale. In other words, you're pointing to a value that was already exist in time of creation of such block of code and will not change until next recomposition. And if you send such block of code to external function, the value will be old there. Instead, use this way:

```val something = remember {```

This means that in other Unit blocks you will get a value by calling ```something.value``` and it will be actual.

Also, another advantage of doing so is the fact that compose will not need to recompose containing function on change of value inside remember. So, for example, when you have:

```kotlin
val something = remember { derivedStateOf { chatModel.chats.value } }
```

and then you use ```something,value``` only inside NON-@Composable function, Compose will give you actual value but will not recompose function that contains this declaration which prevents unneeded recompositions.

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatView.kt
```diff
@@ -961,21 +989,104 @@ fun BoxScope.ChatItemsList(
...
+          val idx = sections.chatItemPosition(itemId)
+          scope.launch {
+            if (idx != null) {
+              listState.animateScrollToItem(scrollPosition(idx), -maxHeight.value)
+              withContext(Dispatchers.Main) {
+                if (!itemsToDrop.isNullOrEmpty()) {
+                  itemsToDrop.forEach {
+                    chatModel.chatItemsSectionArea.remove(it.id)
+                  }
+                  chatModel.chatItems.value.removeIf { chatModel.chatItemsSectionArea[it.id] == null }
```
avently: `removeif` is Java function. Kotlin equivalent is `.removeAll`.  In my practice I remember that when I used it on some old Android version i got crash. Maybe it's lower API when we support (like 21, for example), but anyway, better to use Kotlin functions that do the same as Java functions to be on a safe side

avently: I also noticed that I used it in different place, nice. So it will not make a crash on our min sdk level, but anyway

avently: Also, never modify chatModel.chats.value directly like you do with .removeIf. It should always be non-modifyable list, only items at index can change, not the elements order/size. Apply your changes to a new list and set this list to chatItems using the following block:

```kotlin
withChats {
    chats.removeAll {  }
}
```

Here `withChats` prevents concurrency and gives you chats instance you can use for modification using implemented helpers that will NOT modify the list but create a new one and re-set it to chatModel.chatItems.

So .removeAll here is a helper written in ChatModel.kt:2300, take a look.

The idea of why we do this strange thing is to prevent issues with concurrent modification of data that gave us a lot of random problems with LazyColumn in the past.

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ChatListNavLinkView.kt
```diff
@@ -204,43 +205,55 @@ suspend fun noteFolderChatAction(rhId: Long?, noteFolder: NoteFolder) {
...
+    openLoadedChat(chat.first, chatModel, chat.second)
   }
 }
 
-fun openLoadedChat(chat: Chat, chatModel: ChatModel) {
+fun openLoadedChat(chat: Chat, chatModel: ChatModel, landingSection: ChatLandingSection = ChatLandingSection.Latest) {
   chatModel.chatItemStatuses.clear()
   chatModel.chatItems.replaceAll(chat.chatItems)
   chatModel.chatId.value = chat.chatInfo.id
+  chatModel.chatItemsSectionArea = mutableMapOf<Long, ChatSectionArea>().also { map -> map.putAll(chatModel.chatItems.value.associate { it.id to landingSectionToArea(landingSection) }) }
```
avently, line 232: Better to use direct way like:
```kotlin
  val itemsArea = mutableMapOf<Long, ChatSectionArea>()
  chatModel.chatItems.value.associateTo(itemsArea) { it.id to landingSectionToArea(landingSection) }
  chatModel.chatItemsSectionArea = itemsArea
```

avently: are you sure it should be mutable instance? Why not immutable? Threading issues can be harder to diagnose with mutable instance

avently:
I see `Hide` button in menu when I right click on any text message in ChatView. It does nothing when I click on it

avently:
I can't reveal events (for example, when you establish connection with contact and see permissions in one line)

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatSections.kt
```diff
@@ -0,0 +1,254 @@
...
+          ChatSectionArea.Bottom -> ChatSectionArea.Bottom
+          ChatSectionArea.Current -> if (this.sectionArea == ChatSectionArea.Bottom) ChatSectionArea.Bottom else ChatSectionArea.Current
+          ChatSectionArea.Destination -> if (this.sectionArea == ChatSectionArea.Bottom) ChatSectionArea.Bottom else ChatSectionArea.Destination
+        }
+
+        chatItemsSectionArea.filter { it.value == itemSectionArea }
+          .forEach { chatItemsSectionArea[it.key] = targetSection }
+      }
+    }
+    chatItemsSectionArea.putAll(itemsToAdd.associate { it.id to sectionArea })
```
avently, line 60: Waaay too much loops. Can this be done in one loop?

- for
- filter
- forEach
- associate

avently: for every item you do two loops + top level for...

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatView.kt
```diff
@@ -993,23 +1103,14 @@ fun BoxScope.ChatItemsList(
...
-    contentPadding = PaddingValues(
-      top = topPaddingToContent(),
-      bottom = composeViewHeight.value
-    ),
-    additionalBarOffset = composeViewHeight
-  ) {
-    itemsIndexed(reversedChatItems.value, key = { _, item -> item.id to item.meta.createdAt.toEpochMilliseconds() }) { i, cItem ->
-      val itemScope = rememberCoroutineScope()
+  @Composable
+  fun ChatViewListItem(i: Int, sectionItems: SectionItems, cItem: ChatItem, prevItem: ChatItem?, nextItem: ChatItem?) {
```
avently, line 1107: Here you take sectionItems as a parameter which will cause Compose to make a lot of recompositions when data inside of it changes but in reality you use just a small part of the whole SectionItems. So just pass that small part you need as a parameter, not the whole SectionItems. This way Compose will be able to understand that values you pass are not changed

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
```diff
@@ -65,6 +66,7 @@ object ChatModel {
   // current chat
   val chatId = mutableStateOf<String?>(null)
   val chatItems = mutableStateOf(SnapshotStateList<ChatItem>())
+  val chatItemsSectionArea = mutableMapOf<Long, ChatSectionArea>()
```
avently, line 69: make a comment about what Long parameter means, its not obvious based on context

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatSections.kt
```diff
@@ -0,0 +1,290 @@
...
+data class ChatSectionAreaBoundary (
+  var minIndex: Int,
+  var maxIndex: Int,
+  val area: ChatSectionArea
+)
+
+data class ChatSection (
+  val items: MutableList<SectionItems>,
+  val boundary: ChatSectionAreaBoundary,
+  val itemPositions: MutableMap<Long, Int>
```
avently, line 28: make a comment about what Long and Int mean

PR #5142, opened 2024-11-04
core: preset operators and servers 

epoberezkin:


src/Simplex/Chat/Operators.hs
```diff
@@ -84,8 +94,9 @@ data ConditionsAcceptance
   deriving (Show)
 
 data ServerOperator = ServerOperator
-  { operatorId :: OperatorId,
+  { operatorId :: Maybe OperatorId,
```
spaced4ndy, line 97: I don't think it's a good idea to have Maybe here just for population purposes, it's UI facing types and we'll be dealing with this optionality everywhere. Can have separate type ServerOperatorTemplate and function to transform it

src/Simplex/Chat.hs
```diff
@@ -1549,10 +1621,12 @@ processChatCommand' vr = \case
     currentTs <- liftIO getCurrentTime
     withFastStore' $ \db -> setConditionsNotified db conditionsId currentTs
     ok_
-  APIAcceptConditions conditionsId operators -> do
+  -- TODO switch to IDs
```
spaced4ndy, line 1624: needs operator tags on insert, so if switched to ids it'd have to read operators/operator tags first, why not just pass operator objects

src/Simplex/Chat.hs
```diff
@@ -1806,7 +1880,9 @@ processChatCommand' vr = \case
         let ConnReqUriData {crSmpQueues = q :| _} = crData
             SMPQueueUri {queueAddress = SMPQueueAddress {smpServer}} = q
         cfg <- asks config
-        newUserServers <- L.map (\ServerCfg {server} -> protoServer server) . useServers cfg SPSMP <$> withFastStore' (`getProtocolServers` newUser)
+        liftIO $ putStrLn $ "smpServer " <> show smpServer
+        newUserServers <- userProtoServers cfg SPSMP <$> withFastStore' (`getProtocolServers` newUser)
+        liftIO $ putStrLn $ "newUserServers " <> show newUserServers
```
spaced4ndy, line 1885:
```suggestion
        newUserServers <- userProtoServers cfg SPSMP <$> withFastStore' (`getProtocolServers` newUser)
```

src/Simplex/Chat.hs
```diff
@@ -6659,15 +6737,17 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
         forwardMemberXGrpMemCon :: GroupMemberIntro -> CM ()
-        forwardMemberXGrpMemCon GroupMemberIntro {introId, introStatus}
-          | introStatus == GMIntroToConnected = updateStatus introId GMIntroConnected
-          | introStatus `elem` [GMIntroReConnected, GMIntroConnected] = pure ()
-          | otherwise = updateStatus introId GMIntroReConnected
+        forwardMemberXGrpMemCon GroupMemberIntro {introId, introStatus} = case introStatus of
+          GMIntroToConnected -> updateStatus introId GMIntroConnected
+          GMIntroReConnected -> pure ()
+          GMIntroConnected -> pure ()
+          _ -> updateStatus introId GMIntroReConnected
```
spaced4ndy, line 6750: why refactor?

src/Simplex/Chat/Store/Profiles.hs
```diff
@@ -524,177 +530,262 @@ updateUserAddressAutoAccept db user@User {userId} autoAccept = do
...
     currentTs <- getCurrentTime
-    DB.execute db "DELETE FROM protocol_servers WHERE user_id = ? AND protocol = ? " (userId, protocol)
-    forM_ servers $ \ServerCfg {server, preset, tested, enabled} -> do
-      let ProtoServerWithAuth ProtocolServer {host, port, keyHash} auth_ = server
+    DB.execute db "DELETE FROM protocol_servers WHERE user_id = ? AND protocol = ? " (userId, decodeLatin1 $ strEncode p)
+    forM_ servers $ \UserServer {serverId, server, preset, tested, enabled} -> do
       DB.execute
         db
         [sql|
           INSERT INTO protocol_servers
```
spaced4ndy, line 587: why does it need to preserve ids? could just re-read servers and avoid having 2 similar functions on insert?

src/Simplex/Chat/Store/Profiles.hs
```diff
@@ -524,177 +530,262 @@ updateUserAddressAutoAccept db user@User {userId} autoAccept = do
...
-      [sql|
-        INSERT INTO operator_usage_conditions
-          (server_operator_id, server_operator_tag, conditions_commit, accepted_at)
-        VALUES (?,?,?,?)
-      |]
-      (operatorId, operatorTag, conditionsCommit, acceptedAt)
-  getServerOperators db
+  let ts = Just acceptedAt
+  liftIO $ forM operators $ \op ->
+    acceptConditions_ db op conditionsCommit ts $> op {conditionsAcceptance = CAAccepted ts}
```
spaced4ndy, line 777: why not re-read? it's cheap and less error prone

src/Simplex/Chat/Store/Profiles.hs
```diff
@@ -524,177 +530,262 @@ updateUserAddressAutoAccept db user@User {userId} autoAccept = do
...
+              SELECT conditions_commit, accepted_at
+              FROM operator_usage_conditions
+              WHERE server_operator_id = ?
+              ORDER BY operator_usage_conditions_id DESC
+              LIMIT 1
+            |]
+            (Only operatorId)
+      pure $ case operatorAcceptedConds_ of
+        Just (operatorCommit, acceptedAt_)
+          | operatorCommit /= latestAcceptedCommit -> CARequired Nothing -- TODO should we consider this operator disabled?
```
spaced4ndy, line 730: it should be implementation error if it's in this state and not disabled

src/Simplex/Chat/Store/Profiles.hs
```diff
@@ -524,177 +530,262 @@ updateUserAddressAutoAccept db user@User {userId} autoAccept = do
...
-              -- no conditions were ever accepted for this operator
-              (_, Nothing) -> CARequired Nothing
-              (Just UsageConditions {conditionsCommit = latestAcceptedCommit}, Just operatorCommit)
-                | latestAcceptedCommit == currentCommit ->
-                    if operatorCommit == latestAcceptedCommit
-                      then -- current conditions were accepted for operator
-                        CAAccepted acceptedAt_
-                      else -- current conditions were NOT accepted for operator, but were accepted for other operator(s)
-                        CARequired Nothing
-                | otherwise ->
```
spaced4ndy, line 625: is this check/branch lost?

src/Simplex/Chat/Store/Profiles.hs
```diff
@@ -524,177 +530,262 @@ updateUserAddressAutoAccept db user@User {userId} autoAccept = do
...
+              WHERE server_operator_id = ?
+              ORDER BY operator_usage_conditions_id DESC
+              LIMIT 1
+            |]
+            (Only operatorId)
+      pure $ case operatorAcceptedConds_ of
+        Just (operatorCommit, acceptedAt_)
+          | operatorCommit /= latestAcceptedCommit -> CARequired Nothing -- TODO should we consider this operator disabled?
+          | currentCommit /= latestAcceptedCommit -> CARequired $ conditionsRequiredOrDeadline createdAt (fromMaybe now notifiedAt)
+          | otherwise -> CAAccepted acceptedAt_
```
spaced4ndy, line 732: this trying to save up on one extra branch and using guards instead of if else leads to overall less understandable code. instead of doing direct check

latestAcceptedCommit == currentCommit && operatorCommit == latestAcceptedCommit -> accepted

you now have to read all guards above and negate them to try to understand if this is valid logic:
"""otherwise is "!(operatorCommit /= latestAcceptedCommit) && !(currentCommit /= latestAcceptedCommit)" (and would be more conditions if other guards were above) so it's "(operatorCommit == latestAcceptedCommit) && (currentCommit == latestAcceptedCommit)" so it seems to be correct"""

so instead of going over logic branch by branch and understanding simple conditions one at a time, you now have to go over all cases and do some perverse logic manipulations. otherwise may be good for default (or at best primitive if-else), but here it's not default - if anything accepted is the furthest from default - and this construction simply doesn't fit this case well and forces to understand more complex logic

src/Simplex/Chat.hs
```diff
@@ -2539,13 +2615,15 @@ processChatCommand' vr = \case
...
+      users <- getUsers db
+      smpServers <- getServers db user cfg SPSMP
+      xftpServers <- getServers db user cfg SPXFTP
+      let presentedServersSummary = toPresentedServersSummary agentServersSummary users user smpServers xftpServers _defaultNtfServers
+      pure $ CRAgentServersSummary user presentedServersSummary
     where
-      getServers :: (ProtocolTypeI p, UserProtocol p) => DB.Connection -> ChatConfig -> User -> SProtocolType p -> IO (NonEmpty (ProtocolServer p))
-      getServers db cfg user p = L.map (\ServerCfg {server} -> protoServer server) . useServers cfg p <$> getProtocolServers db user
+      getServers :: (ProtocolTypeI p, UserProtocol p) => DB.Connection -> User -> ChatConfig -> SProtocolType p -> IO [ProtocolServer p]
+      getServers db user cfg p = userProtoServers cfg p <$> getProtocolServers db user
```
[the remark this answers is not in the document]
epoberezkin: what logic? it's already refactored

src/Simplex/Chat.hs
epoberezkin: to use overloaded lists

src/Simplex/Chat/Store/Profiles.hs
epoberezkin: Re-reading things seems a strange approach in simple cases.

epoberezkin: It's ok to use in complex cases, it doesn't mean it should be used in trivial cases - you have a tentency to use it when it's not needed.

src/Simplex/Chat/Store/Profiles.hs
epoberezkin: because I think it's an anti-pattern in such trivial cases.

src/Simplex/Chat/Store/Profiles.hs
epoberezkin: simpler logical expressions are simpler to understand. These branches define the "original" logic that should be applied here, not some simplified logic, and it is simpler than using the matrix.

The logic is a sequence of decision steps:
1) if operator commit is different from latest accepted, it has to be accepted now (Required nothing)
2) or if all operators commit is different from current, then it will be re-accepted at some deadline
3) otherwise it is accepted.

This is simpler than the originally proposed matrix, so it's not about simplifying code, it's about simplifying logic, and replacing decision tree with simpler linear decision sequence.

epoberezkin: so you don't need to negate anything, the code directly implements the required logic here

spaced4ndy (approved):

PR #5152, opened 2024-11-06
ios: server operators api, server views reorganisation; conditions and other view improvements

spaced4ndy:


apps/ios/SimpleXChat/APITypes.swift
```diff
@@ -1236,40 +1245,78 @@ public let operatorsInfo: Dictionary<OperatorTag, ServerOperatorInfo> = [
...
+    public var tradeName: String
+    public var legalName: String?
+    public var serverDomains: [String]
+    public var conditionsAcceptance: ConditionsAcceptance
     public var enabled: Bool
     public var roles: ServerRoles
 
     public var id: Int64 { operatorId }
 
+    public var conditionsName: String {
```
epoberezkin, line 1314:
```suggestion
    public var legalName_: String {
```

apps/ios/SimpleXChat/APITypes.swift
```diff
@@ -1290,104 +1337,154 @@ public struct ServerOperator: Identifiable, Decodable {
...
 
-public struct ServerRoles: Decodable {
+public struct ServerRoles: Equatable, Codable {
     public var storage: Bool
     public var proxy: Bool
+
+    enum CodingKeys: CodingKey {
+        case storage
+        case proxy
+    }
```
epoberezkin, line 1390:
```suggestion
```

PR #5153, opened 2024-11-06
core: fix one item missing from latest in initial and wrong check

diogofcunha:


src/Simplex/Chat/Store/Messages.hs
```diff
@@ -1118,8 +1118,8 @@ getDirectChatInitial_ db user@User {userId} ct@Contact {contactId} count = do
       (chat, gap) <- getDirectChatAround_ db user ct firstUnreadItemId count ""
       case gap of
         Just size -> do
-          if size > snd (divideFetchCountAround_ count)
-            then getLatestItems_ chat size
+          if size > 0
+            then getLatestItems_ chat (size + 1)
```
spaced4ndy, line 1122: don't subtract in getDirectChatItemGapToLatest_?
```suggestion
            then getLatestItems_ chat size
```

spaced4ndy (approved):

PR #5163, opened 2024-11-11
core: remove gaps and make sure page size is always the same

diogofcunha:


src/Simplex/Chat/Store/Messages.hs
```diff
@@ -1106,34 +1084,24 @@ getDirectChatAround_ db user ct@Contact {contactId} aroundItemId count search =
...
-            then getLatestItems_ chat size
-            else pure (chat, Nothing)
-        Nothing -> pure (chat, Nothing)
-    Nothing -> liftIO $ (,Nothing) <$> getDirectChatLast_ db user ct count ""
+      chat <- getDirectChatAround_ db user ct firstUnreadItemId count ""
+      let items = chatItems chat
+      if null items || length items == count
+        then pure chat
+        else do
+          let remainingCount = count - length items
```
spaced4ndy, line 1099: maybe move remainingCount to outside so that it guarantess negative count doesn't get into LIMIT
```suggestion
      let items = chatItems chat
           remainingCount = count - length items
      if null items || remainingCount <= 0
        then pure chat
        else do
```

spaced4ndy (approved):

PR #5183, opened 2024-11-14
build: use openssl 3.0

epoberezkin:


gitayam:
will this solve the Error: openssl@1.1 has been disabled because it is not supported upstream! It was disabled on 2024-10-24.

epoberezkin:
1.1 won't be used

PR #5192, opened 2024-11-16
core: use random servers for each operator

epoberezkin:


src/Simplex/Chat/Migrations/M20241027_server_operators.hs
```diff
@@ -53,4 +53,6 @@ DROP INDEX idx_operator_usage_conditions_server_operator_id;
 DROP TABLE operator_usage_conditions;
 DROP TABLE usage_conditions;
 DROP TABLE server_operators;
+
+DELETE FROM protocol_servers WHERE host LIKE "%.simplexonflux.com";
```
spaced4ndy, line 57:
```suggestion
DELETE FROM protocol_servers WHERE host LIKE "%.simplexonflux.com%";
```

Issue #5087, opened 2024-10-22
[Feature]: add `display-name` option to cli


reply2future:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Platform

Linux, Mac, Windows

### App version

6.1.1.0

### Feature

I want to deploy a websocket server using CLI, then I want to put the CLI to docker container for convenience, but it will need me to input the display name during the first time when running the container. Is there any way to set display name with non-interactive mode? Thank you for any helps
![image](https://github.com/user-attachments/assets/347d85af-a47d-4649-ad82-61eddbc8ed9e)

## ref
[A tutorial for using the websocket server](https://rickcarlino.com/2024/simplex-chat-bots.html)
https://github.com/simplex-chat/simplex-chat/issues/4844#issuecomment-2426749208

epoberezkin:
hey - how does it work without terminal?

reply2future:
> hey - how does it work without terminal?

It doesn't work without terminal.

reply2future:
I found it would use the simplex-chat core lib after reviewing the code of android, and I think, if I understand correctly, it wouldn't take any effort to android, ios and desktop app because the code will work as before if there is no `display-name` param.

reply2future:
![Image](https://github.com/user-attachments/assets/29ea022a-b5d4-44aa-9289-859ca24d3aca)

Just like what my [blog](https://reply2future.top/article/how-to-create-a-bot-for-simplex-chat-with-typescript) said, I want to dockerize CLI as a gateway for bot, although I could use CLI to create the user and generate the database and then mounting it to docker container, but it would be not a good idea for automation. So it would be nice if merged that PR https://github.com/simplex-chat/simplex-chat/pull/5285 :)

PR #5208, opened 2024-11-18
Feature/cli support non interactive mode

reply2future:
implement #5087 

I could deploy CLI as gateway with docker simply, and I think it would be great for anyone wants to build their own bots.


epoberezkin:
moved to https://github.com/simplex-chat/simplex-chat/pull/5285

This is cool, but I wonder how does it work for you without terminal?

PR #5216, opened 2024-11-20
ios: improve onboarding animations

spaced4ndy:


apps/ios/Shared/Views/Onboarding/CreateProfile.swift
```diff
@@ -78,6 +78,35 @@ struct CreateProfile: View {
             }
         }
     }
+
+    private func createProfile() {
```
spaced4ndy, line 82: original version - used on regular profile creation

PR #5230, opened 2024-11-22
desktop: unsaved changes popup for network and servers when clicking middle lane

diogofcunha:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/usersettings/networkAndServers/NetworkAndServers.kt
```diff
@@ -50,21 +50,34 @@ fun ModalData.NetworkAndServersView(close: () -> Unit) {
...
+    showUnsavedChangesAlert(
+      {
+        CoroutineScope(Dispatchers.Default).launch {
+          saveServers(currentRemoteHost?.remoteHostId, currUserServers, userServers)
+          chatModel.centerPanelBackgroundClickHandler = null
+          ModalManager.start.closeModals()
+        }
+      },
+      {
+        chatModel.centerPanelBackgroundClickHandler = null
```
spaced4ndy: easy to forget `chatModel.centerPanelBackgroundClickHandler = null` and leave it model, maybe always set it to null in `clickable(...` in App?

PR #5300, opened 2024-12-02
ios: support business addresses and chats

spaced4ndy:


apps/ios/SimpleXChat/ChatTypes.swift
```diff
@@ -1931,6 +1936,10 @@ public struct GroupInfo: Identifiable, Decodable, NamedChat, Hashable {
         createdAt: .now,
         updatedAt: .now
     )
+
+    public var businessType: BusinessChatType? {
```
spaced4ndy, line 1940: todo move up

PR #5324, opened 2024-12-05
core: take address lock before reading contact request data (to prevent possible race condition if user quickly accepts request several times in a row); android, desktop: show error context in agent CMD errors

spaced4ndy:


src/Simplex/Chat.hs
```diff
@@ -1438,11 +1439,12 @@ processChatCommand' vr = \case
         pure ct {contactUsed, activeConn = Just conn'}
       pure $ CRAcceptingContactRequest user ct'
   APIRejectContact connReqId -> withUser $ \user -> do
-    cReq@UserContactRequest {userContactLinkId, agentContactConnId = AgentConnId connId, agentInvitationId = AgentInvId invId} <-
-      withFastStore $ \db ->
-        getContactRequest db user connReqId
-          `storeFinally` liftIO (deleteContactRequest db user connReqId)
+    userContactLinkId <- withFastStore $ \db -> getUserContactLinkIdByCReq db connReqId
```
spaced4ndy, line 1442: reworked to take lock by getUserContactLinkIdByCReq because it's also taken on address's connection itself in receive loop by its id

PR #5327, opened 2024-12-05
ios: ask for confirmation of save on group preferences sheet dismiss

diogofcunha:


spaced4ndy (approved):

spaced4ndy:
there's the same issue for contact preferences, and likely it's also present on desktop both in groups and contacts (via clicking on middle pane)

PR #5337, opened 2024-12-06
ios: ask for confirmation of save on contact preferences sheet dismiss

diogofcunha:


apps/ios/Shared/Views/Chat/ChatInfoView.swift
```diff
@@ -306,6 +308,9 @@ struct ChatInfoView: View {
                 }
             }
         }
+        .onChange(of: contact.mergedPreferences) { pref in
+            currentFeaturesAllowed = contactUserPrefsToFeaturesAllowed(pref)
+        }
```
spaced4ndy, line 313: to clarify why it's needed, group change from #5327 didn't require it

spaced4ndy (dismissed):

PR #5344, opened 2024-12-06
rfc: chat tags

epoberezkin:


docs/rfcs/2024-12-06-chat-tags.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Solution
+
+This is a simple "quality of life" improvement that's long due, as the number of users and conversations grows, that would probably take just a week to ship.
+
+Proposed UX:
+- tags are per profile.
+- preset tags for favourites, contacts, groups and business chats.
+- user defined tags *must* have text and emoji.
+- if some conversation tag (preset or user defined) is not present in the conversation, the tag will not show in the bar, but would still allow selecting it when tagging.
```
spaced4ndy, line 19: > the tag will not show in the bar

chat view bar?

docs/rfcs/2024-12-06-chat-tags.md
```diff
@@ -0,0 +1,25 @@
...
+- there should be some way to delete tag completely, maybe on long-press on tab bar or maybe in the list of tags?
```
spaced4ndy, line 20: list of tags is to be somewhere in settings?

spaced4ndy: can preset tags be deleted? probably not

docs/rfcs/2024-12-06-chat-tags.md
```diff
@@ -0,0 +1,25 @@
...
+- tagging conversation will be done via the context menu that will open a second menu allowing to choose from existing tag or add a new type and select emoji (which means we need to add our embedded emoji keyboard on desktop and on android, and that should be used in compose too, as many desktop platforms and older android do not allow choosing emoji easily). We could also switch android to emoji font we use on desktop, if it's not too large, as default Android emoji are quite ugly.
```
spaced4ndy, line 21: on ios there's no context menu for conversations in chat list, will it be on swipe button? showing context menu on swipe would be unusual ux

docs/rfcs/2024-12-06-chat-tags.md
```diff
@@ -0,0 +1,25 @@
...
+- on iOS it would be a third "..." button on the trailing swipe area, which will consume Delete button for groups. If there is only Tag button, we don't need "...", we should use Tag directly.
```
spaced4ndy, line 22: showing context menu on swipe would be strange ux, dialogue (similar to confirmation dialogues) or sheet may be better

docs/rfcs/2024-12-06-chat-tags.md
epoberezkin: The list of tags will show when attaching tags to conversations. The tag probably should be automatically deleted once not used.

Preset tags they are not tags really, they are buttons for existing criteria. Not sure what you mean.

docs/rfcs/2024-12-06-chat-tags.md
epoberezkin: yes, dialogue.

epoberezkin: but it would only happen with groups you can delete - we can move delete and clear there, or delete and tag. Probably delete and  clear.

epoberezkin: alternatively, we could make delete button for group owners offer two options - delete for all / leave.

docs/rfcs/2024-12-06-chat-tags.md
```diff
@@ -0,0 +1,25 @@
...
+- it also needs some removable notice, once there are conversations, so there will be some preset tag(s), but no user-defined tags (tappable "Tag your chats" with info on how to tag with X to remove and maybe alert if info was not opened).
```
spaced4ndy, line 23: this feature is totally unnecessary until user has unmanageable number of chats, so it shouldn't be onboarding card or alert after 1-2 chats (like lock). Maybe instead of making it separate element to tag it should be simply hovering over some chat with "text swipe to tag" / "long press to tag", and dismissable without alert. It's not a very difficult / uncommon feature to understand to require some explanation texts and unread warning

docs/rfcs/2024-12-06-chat-tags.md
diogofcunha: Don't quite get this one too, I think it means that if a tag has 0 matches in the conversation list it shouldn't show, please confirm @epoberezkin

docs/rfcs/2024-12-06-chat-tags.md
```diff
@@ -0,0 +1,25 @@
...
+More fun as we solve groups problem.
+
+## Solution
+
+This is a simple "quality of life" improvement that's long due, as the number of users and conversations grows, that would probably take just a week to ship.
+
+Proposed UX:
+- tags are per profile.
+- preset tags for favourites, contacts, groups and business chats.
+- user defined tags *must* have text and emoji.
```
diogofcunha, line 18: if number of tags + preset tags is bigger than screen width it feels like it should scroll horizontally on mobile devices, but what should we do about desktop? Maybe something similar to user picker where we contain it to X lines and if then allow vertical scroll

docs/rfcs/2024-12-06-chat-tags.md
diogofcunha, line 21: Why second menu and not sheet (half-open or 1/3 open in ios)/modal?

docs/rfcs/2024-12-06-chat-tags.md
```diff
@@ -0,0 +1,25 @@
...
+Stop users from using chat profiles to organize chats.
+
+More fun as we solve groups problem.
+
+## Solution
+
+This is a simple "quality of life" improvement that's long due, as the number of users and conversations grows, that would probably take just a week to ship.
+
+Proposed UX:
+- tags are per profile.
```
diogofcunha, line 16: To add a new tag do I need to tag someone first? Given the example of "Family" tag I most likely have a few family members (20, 30) would I need to tag them 1 by 1? It would be better UX if we could create a new tag by click some + button and in that case be able to select a range of contacts to add to that tag

docs/rfcs/2024-12-06-chat-tags.md
epoberezkin: yes. It does scroll in the code I posted.

docs/rfcs/2024-12-06-chat-tags.md
epoberezkin: on iOS - yes, sheet. On android submenu seems better

docs/rfcs/2024-12-06-chat-tags.md
epoberezkin: let's not complicate. There is an argument for multi-select in the list, but seems a separate feature

PR #5356, opened 2024-12-08
core: fix (Invalid cross-device link) error during file transfers

### LINKED ISSUE #2660: [Bug]: cli downloads fail when /tmp is on a different file-system from ~/Downloads

### Platform

Linux

### OS version

Linux

### App version

5.1.2.0

### Current Behavior

When I save file transfers I get the following error: `simplex-chat: /tmp/20230705_093822_210404_rcv.xftp/xftp.decrypted: renameFile:renamePath:rename: unsupported operation (Invalid cross-device link)`.

### Expected Behavior

…

### Steps To Reproduce

…

### Relevant log output

_No response_

Max0nyshchenko:
Fixes #2660.

`rename(2)` fails with `EXDEV` when source and destination are on different
filesystems. For CLI file transfers this is the common case — the XFTP temp
directory is under `/tmp`, the destination is the user's Downloads folder:

```
exception: renameFile:renamePath:rename '/tmp/…/xftp.decrypted' to
'/home/…/Downloads/photo.jpg': unsupported operation (Invalid cross-device link)
```

Adds `crossDeviceRenameFile` to `Simplex.Chat.Util`: falls back to copy + delete
when `rename` fails with `EXDEV`, rethrows anything else unchanged. Applied at the
file-move call sites in `Subscriber.hs`, `Internal.hs`, `Archive.hs` and `Remote.hs`.

The `EXDEV` check reads `ioe_errno` from the caught `IOError` rather than calling
`getErrno` in the handler, so the value can't drift during exception unwinding.

### DIFF
--- a/src/Simplex/Chat/Archive.hs
+++ b/src/Simplex/Chat/Archive.hs
@@ -25,7 +25,7 @@ import Data.Text (Text)
 import qualified Data.Text as T
 import qualified Database.SQLite3 as SQL
 import Simplex.Chat.Controller
-import Simplex.Chat.Util ()
+import Simplex.Chat.Util (crossDeviceRenameFile)
 import Simplex.Messaging.Agent.Client (agentClientStore)
 import Simplex.Messaging.Agent.Store.SQLite (closeDBStore, keyString, sqlString, storeKey)
 import Simplex.Messaging.Agent.Store.SQLite.Common (DBStore (..))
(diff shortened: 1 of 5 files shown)


Max0nyshchenko:

Max0nyshchenko:
Rebased onto current master.

Still reproducing — #2660 is open with reports as recent as 30 Jan 2026, and master still calls `renameFile` directly at these sites.

Two changes vs the original version of this PR:

- `apps/simplex-directory-service/src/Directory/Store.hs` is no longer touched — `restoreDirectoryStore` was removed upstream, so that call site no longer exists. This is 5 files now instead of 6.
- The `EXDEV` check reads `ioe_errno` off the caught exception rather than calling `getErrno` in the handler, so the value can't drift between the failure and the handler running.

Happy to adjust anything here — including the helper's name or which module it belongs in.

PR #5364, opened 2024-12-10
desktop: offer to overwrite file on save

avently:


avently:
This one probably needs to be merged. Don't remember we had any issues with it

PR #5367, opened 2024-12-10
core, ios: chat tags

diogofcunha:


src/Simplex/Chat/Controller.hs
```diff
@@ -617,6 +621,8 @@ data ChatResponse
   | CRContactCode {user :: User, contact :: Contact, connectionCode :: Text}
   | CRGroupMemberCode {user :: User, groupInfo :: GroupInfo, member :: GroupMember, connectionCode :: Text}
   | CRConnectionVerified {user :: User, verified :: Bool, expectedCode :: Text}
+  | CRChatTagged {user :: User, chatTags :: [ChatTag]}
+  | CRChatUntagged {user :: User, chatTagId :: ChatTagId}
```
epoberezkin, line 625:
```suggestion
  | CRChatTagged {user :: User, userTags :: [ChatTag], chatTags :: [ChatTag]}
  | CRChatUntagged {user :: User, userTags :: [ChatTag], chatTags :: [ChatTag]}
```

src/Simplex/Chat/Types.hs
```diff
@@ -1637,6 +1637,21 @@ data CommandData = CommandData
...
+    chatTagText :: Text,
+    chatTagEmoji :: Text
+  }
+  deriving (Show)
+
+data ChatTagChat = ChatTagChat
+  { chatTagId :: Int64,
+    contactId :: Maybe ContactId,
+    groupId :: Maybe GroupId
+  }
```
epoberezkin, line 1652: you shouldn't need this type, it's the record

src/Simplex/Chat/Types.hs
```diff
@@ -1637,6 +1637,21 @@ data CommandData = CommandData
   }
   deriving (Show)
 
+data ChatTag = ChatTag
+  { chatTagId :: Int64,
+    userId :: UserId,
```
epoberezkin, line 1642:
```suggestion
```

src/Simplex/Chat/Controller.hs
```diff
@@ -587,6 +591,7 @@ data ChatResponse
   | CRApiChats {user :: User, chats :: [AChat]}
   | CRChats {chats :: [AChat]}
   | CRApiChat {user :: User, chat :: AChat, navInfo :: Maybe NavigationInfo}
+  | CRChatTags {user :: User, chatTags :: [ChatTag]}
```
diogofcunha, line 594:
```suggestion
  | CRChatTags {user :: User, userTags :: [ChatTag]}
```

epoberezkin: ok

src/Simplex/Chat/Migrations/M20241206_chat_tags.hs
```diff
@@ -0,0 +1,42 @@
...
+  chat_tag_text TEXT UNIQUE NOT NULL,
+  chat_tag_emoji TEXT UNIQUE NOT NULL
+);
+
+CREATE TABLE chat_tags_chats (
+  contact_id INTEGER REFERENCES contacts ON DELETE CASCADE, -- NULL for groups
+  group_id INTEGER REFERENCES groups ON DELETE CASCADE, -- NULL for contacts
+  chat_tag_id INTEGER NOT NULL REFERENCES chat_tags ON DELETE CASCADE,
+  UNIQUE(chat_tag_id, group_id),
+  UNIQUE(chat_tag_id, contact_id)
```
epoberezkin, line 23: maybe better define them outside so they can be changed

src/Simplex/Chat/Migrations/M20241206_chat_tags.hs
```diff
@@ -0,0 +1,42 @@
...
+);
+
+CREATE INDEX idx_chat_tags_user_id ON chat_tags(user_id);
+CREATE INDEX idx_chat_tags_chat_tag_id ON chat_tags(chat_tag_id);
+CREATE INDEX idx_chat_tags_chats_chat_tag_id ON chat_tags_chats(chat_tag_id);
+CREATE INDEX idx_chat_tags_user_id_chat_tag_id ON chat_tags(user_id, chat_tag_id);
```
epoberezkin, line 29: and it feels like some indices are missing to load tags of specific contact/group in joins

src/Simplex/Chat/Controller.hs
```diff
@@ -294,11 +294,15 @@ data ChatCommand
   | ExecChatStoreSQL Text
   | ExecAgentStoreSQL Text
   | SlowSQLQueries
+  | APIGetChatTags
```
epoberezkin, line 297: this needs UserId

src/Simplex/Chat/Controller.hs
```diff
@@ -1068,6 +1075,16 @@ instance FromJSON ComposedMessage where
...
 
+data ChatTagData = ChatTagData
+  { emoji :: Text,
+    text :: Text
+  }
+  deriving (Show)
+
+instance FromJSON ChatTagData where
+  parseJSON (J.Object v) = ChatTagData <$> v .: "emoji" <*> v .: "text"
+  parseJSON invalid = JT.prependFailure "bad ChatTagData, " (JT.typeMismatch "Object" invalid)
```
epoberezkin, line 1086: why do we need custom instance and not automatic with defaultOptions?

epoberezkin: we may need a different parser for CLI commands, but for API could be just JSON.

epoberezkin, line 1086:
```suggestion
```

src/Simplex/Chat/Controller.hs
```diff
@@ -1603,3 +1620,5 @@ $(JQ.deriveFromJSON defaultJSON ''ArchiveConfig)
 $(JQ.deriveFromJSON defaultJSON ''DBEncryptionConfig)
 
 $(JQ.deriveToJSON defaultJSON ''ComposedMessage)
+
+$(JQ.deriveToJSON defaultJSON ''ChatTagData)
```
epoberezkin, line 1624:
```suggestion
$(JQ.deriveJSON defaultJSON ''ChatTagData)
```

src/Simplex/Chat.hs
```diff
@@ -7460,6 +7487,35 @@ closeFileHandle fileId files = do
...
+tagDirectChat' db user cId ctId = do
+  liftIO $ tagDirectChat db cId ctId
+  (allTags, chatTags) <- updatedDirectChatTags db user cId
+  pure $ CRChatTagged user allTags chatTags
+
+updatedDirectChatTags :: DB.Connection -> User -> ContactId -> ExceptT StoreError IO ([ChatTag], [ChatTag])
+updatedDirectChatTags db user cId = do
+  allTags <- liftIO $ getUserChatTags db user
+  let (directTags, otherTags) = partition (\ChatTag {contactId} -> contactId == Just cId) allTags
+  pure (directTags, otherTags)
```
epoberezkin, line 7517: Put them in Store folder together with queries and return tuples from them. The current boundary (and the other functions) might be unnecessary or should be internal to store.

src/Simplex/Chat/Store/Profiles.hs
```diff
@@ -892,3 +900,84 @@ setUserUIThemes :: DB.Connection -> User -> Maybe UIThemeEntityOverrides -> IO (
...
+      [sql|
+        SELECT t.chat_tag_id, t.chat_tag_emoji, t.chat_tag_text, ct.contact_id, ct.group_id
+        FROM chat_tags_chats ct
+        JOIN chat_tags t ON ct.chat_tag_id = t.chat_tag_id
+        WHERE t.user_id = ?
+      |]
+      (Only userId)
+  where
+    toChatTag :: (ChatTagId, Text, Text, Maybe ContactId, Maybe GroupId) -> ChatTag
+    toChatTag (chatTagId, chatTagEmoji, chatTagText, contactId, groupId) = ChatTag {chatTagId, chatTagEmoji, chatTagText, contactId, groupId}
```
epoberezkin, line 979: some of these functions may be unnecessary once functions from chat are moved here.

Not sure that Profiles is the right place - we have Group/Contacts/Shared files, so probably them.

src/Simplex/Chat.hs
```diff
@@ -7460,6 +7487,35 @@ closeFileHandle fileId files = do
...
+tagGroupChat' db user gId ctId = do
+  liftIO $ tagGroupChat db gId ctId
+  (allTags, chatTags) <- updatedGroupChatTags db user gId
+  pure $ CRChatTagged user allTags chatTags
+
+updatedGroupChatTags :: DB.Connection -> User -> GroupId -> ExceptT StoreError IO ([ChatTag], [ChatTag])
+updatedGroupChatTags db user gId = do
+  allTags <- liftIO $ getUserChatTags db user
+  let (groupTags, otherTags) = partition (\ChatTag {groupId} -> groupId == Just gId) allTags
+  pure (groupTags, otherTags)
```
spaced4ndy, line 7500: I don't understand names:
updatedGroupChatTags returns (groupTags, otherTags), but when it's called it's (allTags, chatTags)
chatTags I'd expect to be group tags, but it's in different place, and allTags I'd expect to be all tags, not just remaining tags..

src/Simplex/Chat/Migrations/M20241206_chat_tags.hs
```diff
@@ -0,0 +1,42 @@
...
+down_m20241206_chat_tags =
+  [sql|
+DROP INDEX idx_chat_tags_user_id;
+DROP INDEX idx_chat_tags_chat_tag_id;
+DROP INDEX idx_chat_tags_chats_chat_tag_id;
+DROP INDEX idx_chat_tags_user_id_chat_tag_id;
+
+DROP TABLE chat_tags_chats
+DROP TABLE chat_tags;
+|]
```
spaced4ndy, line 42:
```suggestion
|]

```

src/Simplex/Chat/Types.hs
```diff
@@ -1637,6 +1637,15 @@ data CommandData = CommandData
   }
   deriving (Show)
 
+data ChatTag = ChatTag
+  { chatTagId :: Int64,
+    chatTagText :: Text,
+    chatTagEmoji :: Text,
+    contactId :: Maybe ContactId,
+    groupId :: Maybe GroupId
+  }
```
spaced4ndy, line 1646: I don't seem to understand the model / how it will work with ui - tags are not on chat type, but also tag has single chat and not list. So:
1. when / if tags are to be rendered for chat, all tags have to be scanned for including chat
2. when displaying tags, duplicate tags have to be filtered out in ui, but when tapping on them, ui has to find all tags with same text and emoji to filter corresponding conversations

Database has many-to-many, why at least ChatTag doesn't have list of chats instead of single chat? At least problem 2 wouldn't require additional logic in ui - tags (text/emoji) will be unique and have list of chats to filter. Since it's already in schema, why are tags loaded in such a strange way?
Problem 1 would still require searching in all tags but it's also present in current design and should be ok with small number of chats

epoberezkin: yes, it doesn't make any sense. contactId and groupId are simply not needed here.

src/Simplex/Chat/Controller.hs
spaced4ndy: or can just add list of tags to User and load with active user maybe

src/Simplex/Chat.hs
```diff
@@ -7460,6 +7487,35 @@ closeFileHandle fileId files = do
...
+tagGroupChat' :: DB.Connection -> User -> GroupId -> ChatTagId -> ExceptT StoreError IO ChatResponse
+tagGroupChat' db user gId ctId = do
+  liftIO $ tagGroupChat db gId ctId
+  (allTags, chatTags) <- updatedGroupChatTags db user gId
+  pure $ CRChatTagged user allTags chatTags
+
+updatedGroupChatTags :: DB.Connection -> User -> GroupId -> ExceptT StoreError IO ([ChatTag], [ChatTag])
+updatedGroupChatTags db user gId = do
+  allTags <- liftIO $ getUserChatTags db user
+  let (groupTags, otherTags) = partition (\ChatTag {groupId} -> groupId == Just gId) allTags
```
epoberezkin, line 7499: I don't understand what and why you are partitioning here.

epoberezkin: this approach can't scale if you are trying to extract tags of a given group from the full list of tag links.

You should not load the full list of links into memory, it never scales.

src/Simplex/Chat/Migrations/M20241206_chat_tags.hs
```diff
@@ -0,0 +1,47 @@
...
+  chat_tag_id INTEGER PRIMARY KEY AUTOINCREMENT,
+  user_id INTEGER REFERENCES users,
+  chat_tag_text TEXT NOT NULL,
+  chat_tag_emoji TEXT,
+  tag_order INTEGER NOT NULL
+);
+
+CREATE TABLE chat_tags_chats (
+  contact_id INTEGER REFERENCES contacts ON DELETE CASCADE, -- NULL for groups
+  group_id INTEGER REFERENCES groups ON DELETE CASCADE, -- NULL for contacts
```
spaced4ndy, line 21:
```suggestion
  contact_id INTEGER REFERENCES contacts ON DELETE CASCADE,
  group_id INTEGER REFERENCES groups ON DELETE CASCADE,
```

apps/ios/Shared/Views/ChatList/ChatListNavLink.swift
```diff
@@ -484,6 +560,389 @@ struct ChatListNavLink: View {
...
+                                    .foregroundColor(.secondary)
+                            }
+                        }
+                        TextField("List name...", text: $newName)
+                    }
+                    
+                    Button {
+                        saving = true
+                        if let tId = tagId {
+                            updateChatTag(tagId: tId, chatTagData: ChatTagData(emoji: newEmoji, text: trimmedName))
```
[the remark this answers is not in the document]
spaced4ndy: attempt to save duplicate emojis produces backend error

src/Simplex/Chat/Store/Connections.hs
```diff
@@ -111,7 +111,7 @@ getConnectionEntity db vr user@User {userId, userContactId} agentConnId = do
           chatSettings = ChatSettings {enableNtfs = fromMaybe MFAll enableNtfs_, sendRcpts, favorite}
           mergedPreferences = contactUserPreferences user userPreferences preferences $ connIncognito conn
           activeConn = Just conn
-       in Right Contact {contactId, localDisplayName, profile, activeConn, viaGroup, contactUsed, contactStatus, chatSettings, userPreferences, mergedPreferences, createdAt, updatedAt, chatTs, contactGroupMemberId, contactGrpInvSent, uiThemes, chatDeleted, customData}
+       in Right Contact {contactId, localDisplayName, profile, activeConn, viaGroup, contactUsed, contactStatus, chatSettings, userPreferences, mergedPreferences, createdAt, updatedAt, chatTs, contactGroupMemberId, contactGrpInvSent, chatTags = [], uiThemes, chatDeleted, customData}
```
spaced4ndy, line 114: can't it provide incorrect state to UI? if core sends toView event that triggers chat update.

spaced4ndy: there may be other similar state traps, for example toContact, toGroup - we'd have to exhaustively check that entities built from any usages of those are not passed anywhere in response to be then updated in ui

spaced4ndy: better check other functions too, and only set chatTags to empty list where the entity is new (for example createDirectContact, createAcceptedContact should be ok)

src/Simplex/Chat/Store/Direct.hs
```diff
@@ -842,7 +846,10 @@ getContactIdByName db User {userId} cName =
     DB.query db "SELECT contact_id FROM contacts WHERE user_id = ? AND local_display_name = ? AND deleted = 0" (userId, cName)
 
 getContact :: DB.Connection -> VersionRangeChat -> User -> Int64 -> ExceptT StoreError IO Contact
-getContact db vr user contactId = getContact_ db vr user contactId False
+getContact db vr user contactId = do
+  ct <- getContact_ db vr user contactId False
+  chatTags <- liftIO $ getDirectChatTags db contactId
+  pure (ct :: Contact) {chatTags}
```
spaced4ndy, line 852: revert, move inside toContact
```suggestion
getContact db vr user contactId = getContact_ db vr user contactId False
```

src/Simplex/Chat/Store/Shared.hs
```diff
@@ -398,7 +398,7 @@ toContact vr user ((Only contactId :. (profileId, localDisplayName, viaGroup, di
       chatSettings = ChatSettings {enableNtfs = fromMaybe MFAll enableNtfs_, sendRcpts, favorite}
       incognito = maybe False connIncognito activeConn
       mergedPreferences = contactUserPreferences user userPreferences preferences incognito
-   in Contact {contactId, localDisplayName, profile, activeConn, viaGroup, contactUsed, contactStatus, chatSettings, userPreferences, mergedPreferences, createdAt, updatedAt, chatTs, contactGroupMemberId, contactGrpInvSent, uiThemes, chatDeleted, customData}
+   in Contact {contactId, localDisplayName, profile, activeConn, viaGroup, contactUsed, contactStatus, chatSettings, userPreferences, mergedPreferences, createdAt, updatedAt, chatTs, contactGroupMemberId, contactGrpInvSent, chatTags = [], uiThemes, chatDeleted, customData}
```
spaced4ndy, line 401: toContact should take tags as parameter and set, all places calling it should pass list (either read it before or pass empty list if new entity)

spaced4ndy: for example getContact_ would be doing this and passing tags inside toContact
chatTags <- liftIO $ getDirectChatTags db contactId

spaced4ndy: same for groups / toGroupInfo - read and pass everywhere it's not new entity

PR #5392, opened 2024-12-18
ios: open chat on first unread, "scroll" to quoted items that were not loaded

avently:


apps/ios/Shared/Model/ChatModel.swift
```diff
@@ -82,19 +86,19 @@ class ItemsModel: ObservableObject {
                 await MainActor.run { showLoadingProgress = true }
             } catch {}
         }
+        let type = ChatType(rawValue: String(chatId.prefix(1)))!
+        let id = Int64(chatId.suffix(from: chatId.index(chatId.startIndex, offsetBy: 1)))!
```
epoberezkin, line 90: this is hacky - you should never need to parse IDs - change loadChat and downstream API types instead

apps/ios/Shared/Model/ChatModel.swift
```diff
@@ -82,19 +86,19 @@ class ItemsModel: ObservableObject {
...
+            await MainActor.run { self.isLoading = true }
+            //                try? await Task.sleep(nanoseconds: 5000_000000)
+            await loadChat(type: type, id: id)
+            navigationTimeout.cancel()
+            progressTimeout.cancel()
+            await MainActor.run {
+                self.isLoading = false
+                self.showLoadingProgress = false
+                willNavigate()
+//                ChatModel.shared.chatId = id
```
epoberezkin, line 101: why is this commented?

avently: because the id is set in loadChat -> apiLoadMessages

epoberezkin: aren't there cases when it doesn't change after loading chat?

PR #5403, opened 2024-12-20
core: support postgres backend

spaced4ndy:


src/Simplex/Chat/Store/Postgres/Migrations/M20241220_initial.hs
```diff
@@ -0,0 +1,999 @@
...
+  enabled SMALLINT NOT NULL DEFAULT 1,
+  smp_role_storage SMALLINT NOT NULL DEFAULT 1,
+  smp_role_proxy SMALLINT NOT NULL DEFAULT 1,
+  xftp_role_storage SMALLINT NOT NULL DEFAULT 1,
+  xftp_role_proxy SMALLINT NOT NULL DEFAULT 1,
+  created_at TIMESTAMPTZ NOT NULL DEFAULT (now()),
+  updated_at TIMESTAMPTZ NOT NULL DEFAULT (now())
+);
+CREATE TABLE usage_conditions(
+  usage_conditions_id BIGINT PRIMARY KEY,
```
spaced4ndy, line 648: note: not auto-incremented, code inserts ids manually

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -479,17 +486,17 @@ getChatItemQuote_ db User {userId, userContactId} chatDirection QuotedMsg {msgRe
...
           (userId, groupId, msgId, MDSnd)
-    getGroupChatItemId_ :: Int64 -> MemberId -> IO (Maybe ChatItemId)
-    getGroupChatItemId_ groupId mId =
+    getGroupChatItemId_ :: Int64 -> GroupMemberId -> IO (Maybe ChatItemId)
+    getGroupChatItemId_ groupId groupMemberId =
       maybeFirstRow fromOnly $
         DB.query
           db
           "SELECT chat_item_id FROM chat_items WHERE user_id = ? AND group_id = ? AND shared_msg_id = ? AND item_sent = ? AND group_member_id = ?"
-          (userId, groupId, msgId, MDRcv, mId)
```
spaced4ndy, line 488: I don't fully understand how this even worked at all, may be dead code (in a sense its result is never used (?), because it does get executed). Sqlite must've ignored type incompatibility here and always returned nothing

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -2419,12 +2428,14 @@ processChatCommand' vr = \case
           | name == "" -> withFastStore (`getUserNoteFolderId` user)
           | otherwise -> throwChatError $ CECommandError "not supported"
         _ -> throwChatError $ CECommandError "not supported"
+#if !defined(dbPostgres)
```
epoberezkin, line 2431: why conditional?

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -582,58 +596,53 @@ findDirectChatPreviews_ db User {userId} pagination clq =
...
+        queryWithPagination q p
       CLQFilters {favorite = True, unread = True} -> do
         let q =
               baseQuery
-                <> [sql|
-                      WHERE ct.user_id = ? AND ct.is_user = 0 AND ct.deleted = 0 AND ct.contact_used
-                        AND (ct.favorite = 1
-                          OR ct.unread_chat = 1 OR ChatStats.UnreadCount > 0)
-                  |]
+                <> ( " WHERE ct.user_id = ? AND ct.is_user = 0 AND ct.deleted = 0 AND ct.contact_used = 1"
```
epoberezkin, line 621: this and other similar changes don't seem to be necessary, baseQuery <> [sql||] compiles as is

src/Simplex/Chat/Library/Commands.hs
spaced4ndy: it's only used for sqlite commands

PR #5414, opened 2024-12-23
opt-in to MTE async mode

imgalaxyhunter:
currently only benefits GrapheneOS users on Pixel 8 and 9 who don't turn on MTE for all apps or SimpleX manually. hopefully will be extended to all Pixel users and then more devices


imgalaxyhunter:

PR #5415, opened 2024-12-23
core, ui: support chat item TTL per chat and group aliases

epoberezkin:


src/Simplex/Chat/Library/Commands.hs
```diff
@@ -3247,7 +3282,11 @@ startExpireCIThread user@User {userId} = do
           atomically $ TM.lookup userId expireFlags >>= \b -> unless (b == Just True) retry
           lift waitChatStartedAndActivated
           ttl <- withStore' (`getChatItemTTL` user)
-          forM_ ttl $ \t -> expireChatItems user t False
+          if ttl > 0
```
epoberezkin, line 3285: we already do all these checks outside, why do we repeat them here?

apps/ios/SimpleXChat/ChatTypes.swift
```diff
@@ -4334,52 +4355,97 @@ public enum ChatItemTTL: Identifiable, Comparable, Hashable {
...
+            return self.seconds
+        }
     }
 
     public static func < (lhs: Self, rhs: Self) -> Bool {
         return lhs.comparisonValue < rhs.comparisonValue
     }
 }
 
+public enum ChatTTL: Identifiable, Hashable {
```
epoberezkin, line 4412: This type is structurally equivalent to "ChatItemTTL?", so probably unnecessary.

The UI has to show which global default is set, but there is no need to store it in type, it should be just a view concern.

So at a point where you pass the list to ForEach, you would pass [ChatItemTTL?] and render nil as global default.

Then this type can be removed and all its function be local to view(s).

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -1401,27 +1406,57 @@ processChatCommand' vr = \case
...
+        _ <- if expireChats
+          then do
             lift $ setExpireCIFlag user False
-          Just newTTL -> do
-            oldTTL <- withFastStore' (`getChatItemTTL` user)
-            when (maybe True (newTTL <) oldTTL) $ do
-              lift $ setExpireCIFlag user False
-              expireChatItems user newTTL True
-            withFastStore' $ \db -> setChatItemTTL db user newTTL_
+            expireChatItems user globalTTL True
```
epoberezkin, line 1425: why should it expire all chat items, and not just the chat where TTL changed?

apps/ios/Shared/Views/Chat/ChatInfoView.swift
```diff
@@ -191,97 +191,109 @@ struct ChatInfoView: View {
...
+                        NavigationLink {
+                            ChatWallpaperEditorSheet(chat: chat)
+                        } label: {
+                            Label("Chat theme", systemImage: "photo")
+                        }
+                        //                    } else if developerTools {
+                        //                        synchronizeConnectionButtonForce()
+                        //                    }
                     }
                     .disabled(!contact.ready || !contact.active)
```
spaced4ndy, line 195: why is this removed? updating preferences would not send message at this point and preferences would diverge. same for ratchet if button is shown. verifying code also doesn't make much sense at this point if contact is not even online. receipts is the only one that could be useful but also limited use case

apps/ios/SimpleXChat/ChatTypes.swift
```diff
@@ -4334,52 +4355,97 @@ public enum ChatItemTTL: Identifiable, Comparable, Hashable {
...
-    public var seconds: Int64? {
+    public var seconds: Int64 {
         switch self {
         case .day: return 86400
         case .week: return 7 * 86400
         case .month: return 30 * 86400
+        case .year: return 365 * 86400
         case let .seconds(seconds): return seconds
-        case .none: return nil
+        case .none: return 0
```
spaced4ndy, line 4395: why null didn't work? it could be used as default too

apps/ios/Shared/Views/Chat/ChatInfoView.swift
epoberezkin: it duplicates "disabled" on parent

PR #5426, opened 2024-12-25
ios: open file as preview on click

avently:


avently:
This one is useful and maybe should be reconsidered for inclusion.

PR #5451, opened 2024-12-29
core: rfc, protocol and types for user reports

epoberezkin:


docs/rfcs/2024-12-28-reports.md
```diff
@@ -0,0 +1,80 @@
...
+- moderate the original message,
+- navigate to the original message (requires infinite scrolling, so initially will be only supported on Android and desktop),
+- connect to the user who sent the report - it should be possible even if the group prohibits direct messages. There are two options how this communication can be handled - either by creating a new connection, and shown as normal contacts, or as comments to the report, and sent in the same group connection. The latter approach has the advantage that the interface would not be clutter the interace. The former is much simpler, so should probably be offered as MVP.
+
+This additional chat is necessary, as without it it would be very hard to notice the reports, particularly for the people who moderate multiple groups, and even more so - in our group directory and future super peers.
```
spaced4ndy, line 34: another middle option is to show reports indicator on chat list level per conversation for moderators (e.g. some red flag on each group row that has reports, maybe in the same place where unread counter is shown), and inside groups. Inside groups for example like this:
- bar below chat info bar showing "you have x unprocessed reports"
- inside chat info menu itself

the main downside of the first approach seems to be that messages scrolling up would cause reports being missed, that this "you have reports" bar would be visible only to moderators and up seems expected.

this probably would be simpler approach than the latter. another upside is that you instantly see whether you need to react to some important group, or can ignore. also filtering/tags would apply. if it's a single folder without priority you always need to scan all reports

epoberezkin: I like the idea of having groups with unprocessed reports marked, but it doesn't seem to negate the need for dedicated reports folder...

The scope can easily explode, we need MVP...

epoberezkin: If you continue on this path you need:
- mark groups with unprocessed reports, inside and outside.
- show reports just for this group (yet another view + API).
- show archived reports for a group.
- engage with report from the group itself.

Given that we are trying to achieve two things:
- give people opportunity to easily submit reports to group moderators (including our directory bot),
- give owners the ability not to miss reports.

Having a single "chat" seems a better compromise, if I have to choose.

The positioning is not a problem - we can always show Reports with unprocessed reports first, and let it bump to its chronological position once all are processed.

I don't even want to have a record for this "chat" and have it virtual instead - returned by backend when the list is loaded only if reports are present.

The main gap in the UI is that we need a view that mixes messages from multiple groups, that is - somehow show both group and member, but it may be the foundation for my dream come true :)

Another question is how and if we show submitted reports to the user who sent it. It probably should be shown somewhere? probably in the conversation itself, but with the marker "only you can see it". Then once we implement comments, the conversation with admin will be under this message. Another option is to show report folder for submitting users as well, but without priority - in which case sent reports will be shown as sent messages (again, they need to show in which group they were sent).

diogofcunha: I like the chat idea better, I think it has the biggest potential to solve what we are trying to achieve, and I agree that the group by group approach can be very challenging to manage. A few things that aren't clear about the UX and we should probably at least have on the back of our mind:

- As far as I understand it's rare that groups will have a single moderator, I'm guessing at the time of reporting everyone over a certain privilege gets a notification and a new message pushed to this chat, what happens if 1 moderator acts on that report (could be ignore too)? Do we remove the message from the other moderators? Do we show them another moderator already took action? How simple is that?
- On the back of the previous point, what if moderators disagree, shouldn't they be allowed to have a discussion easily?
- We may be focusing too much on the moderators here and too little on the person doing the report, if I do 3 or 4 reports (can be in same or different groups) how would this experience look like without a mechanism similar to "threads" in most chat apps? It isn't clear to me at all how a helpful conversation should be kept and then how we root each message to the right moderators and make the best effort to avoid mistakes done by reporters

epoberezkin: > Do we remove the message from the other moderators?

Yes, it will be marked as processed based on moderation or blocking events

epoberezkin: > a mechanism similar to "threads" in most chat apps

Yes

src/Simplex/Chat/Migrations/M20250105_reports_msg_content.hs
```diff
@@ -0,0 +1,20 @@
+{-# LANGUAGE QuasiQuotes #-}
+
+module Simplex.Chat.Migrations.M20250105_reports_msg_content where
```
spaced4ndy, line 3:
```suggestion
module Simplex.Chat.Migrations.M20250105_indexes where
```

src/Simplex/Chat/Migrations/M20250105_reports_msg_content.hs
```diff
@@ -0,0 +1,20 @@
...
+
+import Database.SQLite.Simple (Query)
+import Database.SQLite.Simple.QQ (sql)
+
+m20250105_reports_msg_content :: Query
+m20250105_reports_msg_content =
```
spaced4ndy, line 9:
```suggestion
m20250105_indexes :: Query
m20250105_indexes =
```

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -548,16 +550,11 @@ data ChatPreviewData (c :: ChatType) where
...
-paginationByTimeFilter :: PaginationByTime -> (Query, [NamedParam])
-paginationByTimeFilter = \case
-  PTLast count -> ("\nORDER BY ts DESC LIMIT :count", [":count" := count])
-  PTAfter ts count -> ("\nAND ts > :ts ORDER BY ts ASC LIMIT :count", [":ts" := ts, ":count" := count])
-  PTBefore ts count -> ("\nAND ts < :ts ORDER BY ts DESC LIMIT :count", [":ts" := ts, ":count" := count])
-
-type ChatStatsRow = (Int, ChatItemId, Bool)
+type ChatStatsRow = (Int, Int, ChatItemId, BoolInt)
 
+-- TODO [reports]
```
spaced4ndy, line 555:
```suggestion
```

spaced4ndy: or is it not done? comments in Messages.hs imply it's read / not read by design at this point

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -664,84 +653,77 @@ findGroupChatPreviews_ db User {userId} pagination clq =
...
+          WHERE user_id = ? AND group_id IS NOT NULL AND item_status = ?
           GROUP BY group_id
         ) ChatStats ON ChatStats.group_id = g.group_id
+        LEFT JOIN (
+          SELECT group_id, COUNT(1) AS Count
+          FROM chat_items
+          WHERE user_id = ? AND group_id IS NOT NULL
+            AND msg_content_tag = ? AND item_deleted = ?
+          GROUP BY group_id
+        ) ReportCount ON ReportCount.group_id = g.group_id
```
spaced4ndy, line 676: some diverging naming
```suggestion
        ) UnreadChatStats ON UnreadChatStats.group_id = g.group_id
        LEFT JOIN (
          SELECT group_id, COUNT(1) AS ReportCount
          FROM chat_items
          WHERE user_id = ? AND group_id IS NOT NULL
            AND msg_content_tag = ? AND item_deleted = ?
          GROUP BY group_id
        ) ReportChatStats ON ReportChatStats.group_id = g.group_id
```

PR #5455, opened 2024-12-30
android, desktop: chat tags UX improvements

diogofcunha:
- [x]  Show all in menu when any active filter is enabled (including search, and unread and user tags)
- [x]  Show active preset filter as primary color
- [x]  Rename list to add to list in android menu?
- [x]  Rename the second “create list” button to “save list” or “add to list” (in the form depending on where it’s open from
- [x]  Add context menus directly to items,  with third section being change
- [x]  Preset tag for notes
- [x]  Current chat shouldn’t be filtered out and simplify logic like in ios
- [x] Glitch in all inclusion exclusion animation on preset filters dropdown


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ChatListView.kt
```diff
@@ -963,69 +951,79 @@ private fun TagsView() {
...
-      Row(
-        rowSizeModifier
-          .clip(shape = CircleShape)
-          .combinedClickable(
-            onClick = {
-              if (chatModel.activeChatTagFilter.value == ActiveFilter.UserTag(tag)) {
-                chatModel.activeChatTagFilter.value = null
-              } else {
-                chatModel.activeChatTagFilter.value = ActiveFilter.UserTag(tag)
+      val showMenu = remember { mutableStateOf(false) }
```
avently, line 954: should be rememberSaveable

avently:
why `editMode` still in the code? Looks like it is in edit mode by default so what purpose of having this boolean again? If it's not needed, remove it from everywhere. And remove BackHandler that changes editMode. And remove a button to add new list in the list of tags if it's not used anywhere now

avently:
There is still annoying distortion animation when you choose All in menu and then you choose Contacts, for example. And vice versa. Element removes faster than dropdownmenu closes

diogofcunha:
> why `editMode` still in the code? Looks like it is in edit mode by default so what purpose of having this boolean again? If it's not needed, remove it from everywhere. And remove BackHandler that changes editMode. And remove a button to add new list in the list of tags if it's not used anywhere now
> 
> Ah, I see. The edit mode is still in use when choosing a tag for a chat... Sad. Ok, just remove BackHandler then. But still there is no reason to keep editMode as mutableState. Now it's just a boolean

We should probably rename it reorderMode for clarity, what do you think?

avently:
> probably rename it reorderMode for clarity

Sounds good

PR #5466, opened 2025-01-02
ios: reports inline

diogofcunha:


apps/ios/Shared/Views/Chat/ChatItem/FramedItemView.swift
```diff
@@ -147,6 +157,8 @@ struct FramedItemView: View {
             case let .link(_, preview):
                 CILinkView(linkPreview: preview)
                 ciMsgContentView(chatItem)
+            case .report(_, _):
```
diogofcunha, line 160: not needed, default here does this

apps/ios/Shared/Views/Chat/ComposeMessage/ComposeView.swift
```diff
@@ -116,13 +117,24 @@ struct ComposeState {
         default: return false
         }
     }
-
+    
+    var submitingValidReport: Bool {
```
diogofcunha, line 121:
```suggestion
    var submittingValidReport: Bool {
```

apps/ios/Shared/Views/Chat/ComposeMessage/ComposeView.swift
```diff
@@ -686,6 +712,31 @@ struct ComposeView: View {
         .frame(maxWidth: .infinity, alignment: .leading)
         .background(.thinMaterial)
     }
+    
+    
+    private func reportReasonView(_ reason: ReportReason) -> some View {
+        HStack {
```
diogofcunha, line 718: probably better to do the full string on each case, in some languages this may not even translate well

PR #5469, opened 2025-01-02
core: api and protocol support for reporting messages to group moderators

epoberezkin:


src/Simplex/Chat/Protocol.hs
```diff
@@ -121,6 +123,10 @@ businessChatsVersion = VersionChat 10
 businessChatPrefsVersion :: VersionChat
 businessChatPrefsVersion = VersionChat 11
 
+-- support updating preferences in business chats (MCReport message content)
```
spaced4ndy, line 126:
```suggestion
-- support receiving content reports
```

src/Simplex/Chat/Store/Groups.hs
```diff
@@ -743,6 +744,14 @@ getGroupMembers db vr user@User {userId, userContactId} GroupInfo {groupId} = do
       (groupMemberQuery <> " WHERE m.group_id = ? AND m.user_id = ? AND (m.contact_id IS NULL OR m.contact_id != ?)")
       (userId, groupId, userId, userContactId)
 
+getGroupModerators :: DB.Connection -> VersionRangeChat -> User -> GroupInfo -> IO [GroupMember]
+getGroupModerators db vr user@User {userId, userContactId} GroupInfo {groupId} = do
+  map (toContactMember vr user)
+    <$> DB.query
+      db
+      (groupMemberQuery <> " WHERE m.group_id = ? AND m.user_id = ? AND (m.contact_id IS NULL OR m.contact_id != ?) AND member_role IN (?,?,?)")
+      (userId, groupId, userId, userContactId, GRModerator, GRAdmin, GROwner)
```
spaced4ndy, line 753:
```suggestion
      (groupMemberQuery <> " WHERE m.user_id = ? AND m.group_id = ? AND (m.contact_id IS NULL OR m.contact_id != ?) AND member_role IN (?,?,?)")
      (userId, groupId, userId, userContactId, GRModerator, GRAdmin, GROwner)
```

spaced4ndy: there's index
CREATE INDEX idx_group_members_group_id ON group_members(user_id, group_id);
without checking plan this order of filters is more likely to use it

spaced4ndy: and most or all queries don't inverse these, it's usually user_id, then entity id

PR #5485, opened 2025-01-07
android, desktop: inline reports

diogofcunha:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/item/ChatItemView.kt
```diff
@@ -1202,8 +1411,59 @@ fun moderateMessagesAlertDialog(itemIds: List<Long>, questionText: String, delet
...
+        }
+      }
+    },
+    destructive = true,
+  )
+}
+
 expect fun copyItemToClipboard(cItem: ChatItem, clipboard: ClipboardManager)
 
+private suspend fun getLocalIdForReportedMessage(
```
diogofcunha, line 1446: @avently there is a similar piece of code to this in the Loader for chat items, but it deals with a lot of state that I don't have as state in this view + scrolls when done, didn't seem worth the complication of trying to wire up everything for something so small, but you may know an easy way to do it

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
```diff
@@ -1642,7 +1642,7 @@ data class GroupMember (
   fun canChangeRoleTo(groupInfo: GroupInfo): List<GroupMemberRole>? =
     if (!canBeRemoved(groupInfo)) null
     else groupInfo.membership.memberRole.let { userRole ->
-      GroupMemberRole.values().filter { it <= userRole && it != GroupMemberRole.Author }
+      GroupMemberRole.supportedRoles.filter { it <= userRole }
```
avently, line 1645: what does it mean "supported role"? Name suggests that any other role except from supportedRoles is not supported. Please, use naming that describes what it actually means

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
```diff
@@ -2120,8 +2126,14 @@ data class ChatItem (
...
 
+  val isReport: Boolean get() = when (content) {
+    is CIContent.SndMsgContent, is CIContent.RcvMsgContent ->
+      content.msgContent is MsgContent.MCReport
+    else -> false
+  }
+
   val canBeDeletedForSelf: Boolean
-    get() = (content.msgContent != null && !meta.isLive) || meta.itemDeleted != null || isDeletedContent || mergeCategory != null || showLocalDelete
+    get() = !isReport && (content.msgContent != null && !meta.isLive) || meta.itemDeleted != null || isDeletedContent || mergeCategory != null || showLocalDelete
```
avently, line 2136: add more braces as it may now be incorrect

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatView.kt
```diff
@@ -1917,7 +1917,7 @@ private fun scrollToItem(
   }
 }
 
-private fun findQuotedItemFromItem(
+fun findQuotedItemFromItem(
```
avently, line 1920: it still can be private

avently:
Is it expected that I can report my own message?

avently:
There is a padding between the line "Report other: only group moderators can see it" and quoted content in ComposeView. I know we have it in other cases too and if you know how to fix it, please, fix it. If not, it's ok, someday we will find time on it

avently:
It looks like the moderate button is visible even if the message already moderated and once repeated, there is an error in the console. If there is no easy way (without loading the possibly moderated item) to determine is it moderated already or not, forget, not a big deal

avently:
Overall, apart from the comments I mentioned, the code looks good and UI feels comfortable to use, very good.

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/item/ChatItemView.kt
avently: it's fine like this

PR #5503, opened 2025-01-10
ios: report tags and icon on ChatList

avently:


apps/ios/Shared/Model/ChatModel.swift
```diff
@@ -106,6 +106,8 @@ class ChatTagsModel: ObservableObject {
     @Published var userTags: [ChatTag] = []
     @Published var activeFilter: ActiveFilter? = nil
     @Published var presetTags: [PresetTag:Int] = [:]
+    @Published var collapsiblePresetTags: [PresetTag:Int] = [:]
```
diogofcunha, line 109: are there any downside of instead of making 2 new properties in the model just move the this information as a new var of preset tag?

https://github.com/simplex-chat/simplex-chat/pull/5503/files#diff-3cd419442d11b07e971c230143e3064834892aecbcc1e31c851114d06c8a3800R871

same you have in that function but inside the preset tag type as a var, maybe i'm missing something but seems simpler than keeping 2 new vars in the model to hold this info

epoberezkin: neither is needed

PR #5539, opened 2025-01-16
desktop, android: use timestamp as file name for videos

diogofcunha:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ComposeView.kt
```diff
@@ -610,7 +610,17 @@ fun ComposeView(
                 if (remoteHost == null) saveAnimImage(it.uri)
                 else CryptoFile.desktopPlain(it.uri)
               is UploadContent.Video ->
-                if (remoteHost == null) saveFileFromUri(it.uri)
+                if (remoteHost == null) saveFileFromUri(
+                  it.uri,
+                  getDestFileName = { filename ->
```
avently, line 615: I don't like indirection here. Would be better to pass some parameter to saveFileFromUri like:
`fileNamePrefix: String?` which then be transferred to generateNewFileName (if not null) or uniqueCombine (if null)

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/helpers/Utils.kt
```diff
@@ -247,13 +247,26 @@ fun saveAnimImage(uri: URI): CryptoFile? {
...
     val inputStream = uri.inputStream()
     val fileToSave = getFileName(uri)
     return if (inputStream != null && fileToSave != null) {
-      val destFileName = uniqueCombine(fileToSave, File(getAppFilePath("")))
+      val destFileName = if (hiddenFileNamePrefix == null) {
+        uniqueCombine(fileToSave, File(getAppFilePath("")))
+      } else {
+        val ext = when {
+          // remove everything but extension
+          fileToSave.contains(".") -> fileToSave.replaceBeforeLast('.', "").replace(".", "")
```
avently, line 265: .substringAfterLast(".")

PR #5546, opened 2025-01-17
android, desktop: support chat item ttl per chat

diogofcunha:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatInfoView.kt
```diff
@@ -1308,6 +1343,77 @@ fun queueInfoText(info: Pair<RcvMsgInfo?, ServerQueueInfo>): String {
   return generalGetString(MR.strings.message_queue_info_server_info).format(json.encodeToString(qInfo), msgInfo)
 }
 
+fun setChatTTLAlert(
+  m: ChatModel,
```
avently, line 1347: this pattern transferring chatModel by param is gone couple of years ago. Just declare it inside a function if you need it named `m` like this:
val m = chatModel

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatInfoView.kt
```diff
@@ -1308,6 +1343,77 @@ fun queueInfoText(info: Pair<RcvMsgInfo?, ServerQueueInfo>): String {
...
+    confirmText = generalGetString(if (newTTLToUse.neverExpires) MR.strings.disable_automatic_deletion else MR.strings.delete_messages),
+    onConfirm = { setChatTTL(m, rhId, chatInfo, selectedChatTTL, progressIndicator, previousChatTTL) },
+    onDismiss = { selectedChatTTL.value = previousChatTTL },
+    onDismissRequest = { selectedChatTTL.value = previousChatTTL },
+    destructive = true,
+  )
+}
+
+private fun setChatTTL(
+  m: ChatModel,
```
avently, line 1375: same

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatInfoView.kt
```diff
@@ -1308,6 +1343,77 @@ fun queueInfoText(info: Pair<RcvMsgInfo?, ServerQueueInfo>): String {
...
+      afterSetChatTTL(m, chatInfo, progressIndicator)
+    } catch (e: Exception) {
+      chatTTL.value = previousChatTTL
+      afterSetChatTTL(m, chatInfo, progressIndicator)
+      AlertManager.shared.showAlertMsg(generalGetString(MR.strings.error_changing_message_deletion), e.stackTraceToString())
+    }
+  }
+}
+
+private fun afterSetChatTTL(m: ChatModel, chatInfo: ChatInfo, progressIndicator: MutableState<Boolean>) {
```
avently, line 1395: same

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatItemsLoader.kt
```diff
@@ -47,13 +49,19 @@ suspend fun apiLoadMessages(
...
+          } else if (replaceChat) {
+            replaceChat(chat.remoteHostId, chat.id, chat)
           } else {
             updateChatInfo(chat.remoteHostId, chat.chatInfo)
             updateChatStats(chat.remoteHostId, chat.id, chat.chatStats)
           }
         }
       }
       withChats(contentTag) {
+        if (redirectDisabled && chatModel.chatId.value != chat.id) {
```
avently, line 61: why do you need to have a check on chatId here if it's already checked before?
https://github.com/simplex-chat/simplex-chat/pull/5546/files#diff-c73fbf26887eb073849e49f529bfa56f8592ac068157bd63d7e963f2e5a7f73bR37

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatInfoView.kt
```diff
@@ -1308,6 +1343,77 @@ fun queueInfoText(info: Pair<RcvMsgInfo?, ServerQueueInfo>): String {
...
+        redirectDisabled = true
+      )
+    } catch (e: Exception) {
+      Log.e(TAG, "apiGetChat error: ${e.message}")
+    } finally {
+      progressIndicator.value = false
+    }
+  }
+}
+
```
avently, line 1416: can this be replaced with `openChat` funciton from ChatListNavLinkView? What's so special in it so you have to introduce new call and parameters to the apiLoadMessages?

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/group/GroupChatInfoView.kt
```diff
@@ -314,7 +337,7 @@ fun ModalData.GroupChatInfoLayout(
       if (s.isEmpty()) members else members.filter { m -> m.anyNameContains(s) }
     }
   }
-  Box {
+  Box(Modifier.alpha(if (disabled.value) 0.6f else 1f)) {
```
avently, line 340: what purpose of introducing another way of showing progress here? There was no such thing in the app before

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/database/DatabaseView.kt
```diff
@@ -300,21 +304,27 @@ private fun setChatItemTTLAlert(
...
     }
-    all.map {
-      when (it) {
-        is ChatItemTTL.None -> it to generalGetString(MR.strings.chat_item_ttl_none)
-        is ChatItemTTL.Day -> it to generalGetString(MR.strings.chat_item_ttl_day)
-        is ChatItemTTL.Week -> it to generalGetString(MR.strings.chat_item_ttl_week)
-        is ChatItemTTL.Month -> it to generalGetString(MR.strings.chat_item_ttl_month)
-        is ChatItemTTL.Seconds -> it to String.format(generalGetString(MR.strings.chat_item_ttl_seconds), it.secs)
-      }
+    val options = mutableListOf<Pair<ChatItemTTL?, String>>().apply {
```
avently, line 319:
```kotlin
val options: MutableList<Pair<ChatItemTTL?, String>> = all.map { it to it.text }.toMutableList()
```

is this what you wanted?

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatInfoView.kt
```diff
@@ -520,15 +543,16 @@ fun ChatInfoLayout(
...
+  disabled: State<Boolean>
 ) {
   val cStats = connStats.value
   val scrollState = rememberScrollState()
   val scope = rememberCoroutineScope()
   KeyChangeEffect(chat.id) {
     scope.launch { scrollState.scrollTo(0) }
   }
-  ColumnWithScrollBar {
+  ColumnWithScrollBar(Modifier.alpha(if (disabled.value) 0.6f else 1f)) {
```
avently, line 555: same question about introducing new way of progress

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/group/GroupChatInfoView.kt
```diff
@@ -394,7 +417,18 @@ fun ModalData.GroupChatInfoLayout(
...
+      SectionView {
+        TtlOptions(
+          chatItemTTL,
+          enabled = remember { mutableStateOf(true) },
+          onSelected = setChatItemTTL,
+          default = chatModel.chatItemTTL
+        )
+      }
+      SectionTextFooter(stringResource(MR.strings.chat_ttl_options_footer))
+      SectionDividerSpaced(maxTopPadding = true, maxBottomPadding = false)
```
avently, line 431: there is a smaller padding on bottom than on top. Probably, `maxBottomPadding` should be `true`?
On direct chat the padding looks ok

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatInfoView.kt
diogofcunha: In here there are 2 situations, this seemed to be the best way to handle it.

1. The api replies fast with deletion, in the case open chat would be fine and was my initial tough (not sure if it would also update the chat list preview tough)
2. API to delete is slow if you have many items, user can click in another chat, we still want to update the chat in chat list but we don't want to redirect, correct me if i'm wrong but in that cases this would randomly redirect you to the previous chat, that's not a good user experience, looks more like a bug from user perspective

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatItemsLoader.kt
diogofcunha: not the same if you call this function with initial right? what was happening when I added artificial delay was this would redirect to the previous chat

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ChatInfoView.kt
avently: 1. If you're unsure, check it:)
2. So before calling openChat, just make this check for chatId. Right before openChat. No need to put another params to the main loading function

PR #5552, opened 2025-01-20
core: don't subscribe to deleted/left groups, read less data for groups on subscription

spaced4ndy:


src/Simplex/Chat/Controller.hs
```diff
@@ -736,7 +736,7 @@ data ChatResponse
   | CRNetworkStatuses {user_ :: Maybe User, networkStatuses :: [ConnNetworkStatus]}
   | CRHostConnected {protocol :: AProtocolType, transportHost :: TransportHost}
   | CRHostDisconnected {protocol :: AProtocolType, transportHost :: TransportHost}
-  | CRGroupInvitation {user :: User, groupInfo :: GroupInfo}
+  | CRGroupInvitation {user :: User, groupId :: GroupId, groupName :: GroupName}
```
epoberezkin, line 739: I think this particular event is used in other contexts? so it may need full group info?

src/Simplex/Chat/Controller.hs
```diff
@@ -752,8 +752,7 @@ data ChatResponse
   | CRUnknownMemberCreated {user :: User, groupInfo :: GroupInfo, forwardedByMember :: GroupMember, member :: GroupMember}
   | CRUnknownMemberBlocked {user :: User, groupInfo :: GroupInfo, blockedByMember :: GroupMember, member :: GroupMember}
   | CRUnknownMemberAnnounced {user :: User, groupInfo :: GroupInfo, announcingMember :: GroupMember, unknownMember :: GroupMember, announcedMember :: GroupMember}
-  | CRGroupEmpty {user :: User, groupInfo :: GroupInfo}
-  | CRGroupRemoved {user :: User, groupInfo :: GroupInfo}
+  | CRGroupEmpty {user :: User, groupId :: GroupId, groupName :: GroupName}
```
epoberezkin, line 755: also probably better to make a type for group reference

src/Simplex/Chat/Types.hs
```diff
@@ -372,6 +372,21 @@ optionalFullName displayName fullName
   | T.null fullName || displayName == fullName = ""
   | otherwise = " (" <> fullName <> ")"
 
+data GroupToSubscribe = GroupToSubscribe
```
epoberezkin, line 375:
```suggestion
data ShortGroupInfo = ShortGroupInfo
```

I think it should be more generic, then we can use it wider

src/Simplex/Chat/Types.hs
```diff
@@ -372,6 +372,21 @@ optionalFullName displayName fullName
...
+  { groupId :: GroupId,
+    groupName :: GroupName,
+    membershipStatus :: GroupMemberStatus,
+    membersToSubscribe :: [GroupMemberToSubscribe]
+  }
+
+data GroupMemberToSubscribe = GroupMemberToSubscribe
```
epoberezkin, line 382: same here

src/Simplex/Chat/Controller.hs
epoberezkin: ShortGroupInfo (see below)

spaced4ndy: Then it should mirror Group/GroupInfo so in response it's without list of members. ShortGroup/ShortGroupInfo - latter in response

src/Simplex/Chat/Controller.hs
spaced4ndy: no

PR #5555, opened 2025-01-20
core: member mentions, types and rfc

epoberezkin:


docs/rfcs/2025-01-20-member-mentions.md
```diff
@@ -0,0 +1,39 @@
...
+
+```
+Hello @'abcd1234'
+```
+
+In addition, the message JSON will include the array of mentions, as objects with property `memberId`. This is to ensure the intent and that the fragments of text are treated as mentions.
+
+Using an immutable `memberId` would prevent any race conditions and duplicate display names. The receiving client would show a local view name (display name or an alias), and would open a correct member card when mention is tapped.
+
+There should be a reasonable limit on the number of mentions in a group, e.g. 3. This is to prevent abuse, expensive processing both in the client and in super-peers that would have to forward member profiles if they were not forwarded before. This limit has to be enforced both on sending and receiving ends.
```
spaced4ndy, line 19: how will mention look in UI in chat item's text? possibly it's not necessary to render avatar (as tapping it already opens member info page), so sending/forwarding mentions with name included should be sufficient without additionally sending profiles, even if receiving client doesn't have that member's profile.

docs/rfcs/2025-01-20-member-mentions.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Forwarding and saving to local items
+
+When forwarding a message with mentions to another conversation the app should forward:
+- display names instead of member IDs in the text.
+- remove mentions from the message data.
+
+When saving a message to local chat items the app retains the reference to the original group, so it could include member IDs. It means that member data should be loaded from the correct group. Alternatively, the reference could include one of:
+- both member ID and display name at the time of saving to the local item.
+- only display name, and no mentions.
```
spaced4ndy, line 39: so mention would also have to store reference to group - probably overengineering for an unclear use case. why would you need to open member info page from notes - to manage from notes? instead we could improve navigation to group / chat item inside it

src/Simplex/Chat/Markdown.hs
```diff
@@ -50,6 +50,7 @@ data Format
   | Colored {color :: FormatColor}
   | Uri
   | SimplexLink {linkType :: SimplexLinkType, simplexUri :: Text, smpHosts :: NonEmpty Text}
+  | Mention {memberId :: MemberId}
```
spaced4ndy, line 53: \+ name?

src/Simplex/Chat/Messages.hs
```diff
@@ -150,13 +151,21 @@ data ChatItem (c :: ChatType) (d :: MsgDirection) = ChatItem
...
     formattedText :: Maybe MarkdownList,
     quotedItem :: Maybe (CIQuote c),
     reactions :: [CIReactionCount],
     file :: Maybe (CIFile d)
   }
   deriving (Show)
 
+data MentionedMember = MentionedMember
+  { memberViewName :: Text, -- display name or alias
+    memberRole :: GroupMemberRole
```
spaced4ndy, line 164: why role is needed here? also why memberId is not in this type and only in markdown?

src/Simplex/Chat/Protocol.hs
```diff
@@ -589,9 +589,20 @@ msgContentTag = \case
...
+data ExtMsgContent = ExtMsgContent
+  { content :: MsgContent,
+    mentions :: [MemberMention],
+    file :: Maybe FileInvitation,
+    ttl :: Maybe Int,
+    live :: Maybe Bool
+  }
+  deriving (Eq, Show)
+
+data MemberMention = MemberMention {memberId :: MemberId}
```
spaced4ndy, line 601: \+ name?

docs/rfcs/2025-01-20-member-mentions.md
```diff
@@ -0,0 +1,39 @@
...
+## Problem
+
+Mention members in the messages.
+
+## Solution
+
+Message text can include references to the shared member IDs in base64 format:
+
+```
+Hello @'abcd1234'
```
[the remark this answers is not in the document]
epoberezkin: name can contain <, so if it's included it's better to use @'abcd/name'

But I think we can just ignore it with two stage-release anyway.

docs/rfcs/2025-01-20-member-mentions.md
epoberezkin: Just a highlighted name. For forwards - not even highlighted, as they don't refer to the current members.

docs/rfcs/2025-01-20-member-mentions.md
epoberezkin: local member ID is enough.

src/Simplex/Chat/Messages.hs
epoberezkin: to show it as **ep** (*admin*)

epoberezkin: memberId should be included, I initially made it a Map.

src/Simplex/Chat/Protocol.hs
epoberezkin: Don't think it's needed here.

docs/rfcs/2025-01-20-member-mentions.md
```diff
@@ -4,36 +4,57 @@
...
 
-There should be a reasonable limit on the number of mentions in a group, e.g. 3. This is to prevent abuse, expensive processing both in the client and in super-peers that would have to forward member profiles if they were not forwarded before. This limit has to be enforced both on sending and receiving ends.
+As display names are not unique in the group, we should convert them to locally-unique names (per message), by appending _1, _2, as necessary, and the same locally unique names should be used in the mapping to member IDs. These locally unique names must NOT use local user aliases, and must NOT use localDisplayName, as otherwise it may leak information that is known only to the user's client.
+
+There should be a reasonable limit on the number of mentions per message, e.g. 3. This is to prevent abuse, expensive processing both in the client and in super-peers that would have to forward member profiles if they were not forwarded before. This limit has to be enforced both on sending and receiving ends.
```
spaced4ndy, line 34: it's not much harder to send multiple messages with different mentions, so what's the point in this limitation? also why does profile need to be forwarded if mention now includes name?

epoberezkin: not profile, mapping from mentioned name to member_id

src/Simplex/Chat/Messages.hs
```diff
@@ -150,17 +151,28 @@ data ChatItem (c :: ChatType) (d :: MsgDirection) = ChatItem
...
+data MentionedMember = MentionedMember
+  { mentionName :: ContactName, -- name used in the message text, used to look up this object
+    memberId :: MemberId,
+    groupMemberId :: Maybe GroupMemberId,
+    memberViewName :: Text, -- current member display name or alias, shown in the message
+    memberRole :: GroupMemberRole -- used for admins/owners in the message
+  }
+  deriving (Show)
+
+isUserMention :: ChatInfo c -> ChatItem c d -> Bool
```
epoberezkin, line 171: this function can be removed, and instead userMention property used.

src/Simplex/Chat/View.hs
```diff
@@ -499,7 +499,7 @@ responseToView hu@(currentRH, user_) ChatConfig {logLevel, showReactions, showRe
     contactList :: [ContactRef] -> String
     contactList cs = T.unpack . T.intercalate ", " $ map (\ContactRef {localDisplayName = n} -> "@" <> n) cs
     unmuted :: User -> ChatInfo c -> ChatItem c d -> [StyledString] -> [StyledString]
-    unmuted u chat ci@ChatItem {chatDir} = unmuted' u chat chatDir $ isMention ci
+    unmuted u chat ci@ChatItem {chatDir} = unmuted' u chat chatDir $ isUserMention chat ci
```
epoberezkin, line 502: use userMention

src/Simplex/Chat/Controller.hs
```diff
@@ -1635,4 +1623,18 @@ $(JQ.deriveFromJSON defaultJSON ''DBEncryptionConfig)
 
 $(JQ.deriveToJSON defaultJSON ''ComposedMessage)
 
+-- This instance is needed for backward compatibility, can be removed in v6.0
```
spaced4ndy, line 1626: now is 6.3 - in 7.0?

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -1786,10 +1788,11 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
             -- received an update from the sender, so that it can be referenced later (e.g. by broadcast delete).
             -- Chat item and update message which created it will have different sharedMsgId in this case...
             let timed_ = rcvGroupCITimed gInfo ttl_
-            ci <- saveRcvChatItem' user (CDGroupRcv gInfo m) msg (Just sharedMsgId) brokerTs content Nothing timed_ live
+            -- TODO [mentions] compare and remove old mentions if necessary
```
spaced4ndy, line 1791: this would be in updateRcvChatItem

src/Simplex/Chat/Protocol.hs
```diff
@@ -589,9 +594,23 @@ msgContentTag = \case
...
+  { content :: MsgContent,
+    mentions :: [MemberMention],
+    file :: Maybe FileInvitation,
+    ttl :: Maybe Int,
+    live :: Maybe Bool
+  }
+  deriving (Eq, Show)
+
+-- memberName is a locally (per message) unique display name of member.
+-- Suffixes _1, _2 should be appended to make names locally unique.
```
spaced4ndy, line 607: it seems for each added mention we'd need to scan whole list of members (not just already added mentions), to see if `<name>_<suffix>` is not already taken as just name by another member. but also what's the point of uniqueness here? mentioned clients will "know" if mention refers to them by member id, other members still won't be able to understand who's mentioned even with suffixes. for example, if group has multiple members with name "alice", naming mentions as "alice_1", "alice_2" etc. doesn't really clarify which alices are mentioned. moreover, it even makes things worse when in different message different set of alices is mentioned with same suffixes. (what I propose is to not bother with suffixes at all and just mention by display name)

src/Simplex/Chat/Store/SQLite/Migrations/M20250126_mentions.hs
```diff
@@ -0,0 +1,29 @@
...
+import Database.SQLite.Simple.QQ (sql)
+
+m20250126_mentions :: Query
+m20250126_mentions =
+  [sql|
+ALTER TABLE chat_items ADD COLUMN user_mention INTEGER NOT NULL DEFAULT 0;
+
+CREATE TABLE chat_item_mentions (
+  chat_item_mention_id INTEGER PRIMARY KEY AUTOINCREMENT,
+  chat_item_id INTEGER NOT NULL REFERENCES chat_items,
```
spaced4ndy, line 15:
```suggestion
  chat_item_id INTEGER NOT NULL REFERENCES chat_items ON DELETE CASCADE,
```

src/Simplex/Chat/Protocol.hs
epoberezkin: The name should be locally (per message) unique, not per member-list. The suffix will only be used if two different Alices are mentioned in the same message.

src/Simplex/Chat/Store/SQLite/Migrations/M20250126_mentions.hs
```diff
@@ -0,0 +1,34 @@
+{-# LANGUAGE QuasiQuotes #-}
+
+module Simplex.Chat.Store.SQLite.Migrations.M20250126_mentions where
+
+import Database.SQLite.Simple (Query)
+import Database.SQLite.Simple.QQ (sql)
+
+m20250126_mentions :: Query
```
spaced4ndy, line 8: add to postgres initial migration

PR #5562, opened 2025-01-22
core: get group history faster

spaced4ndy:


src/Simplex/Chat/Store/Messages.hs
```diff
@@ -3070,12 +3074,11 @@ getGroupHistoryItems db user@User {userId} GroupInfo {groupId} m count = do
             SELECT i.chat_item_id
             FROM chat_items i
             LEFT JOIN group_snd_item_statuses s ON s.chat_item_id = i.chat_item_id AND s.group_member_id = ?
```
spaced4ndy, line 3076: had to keep this join, without it it does lead to unnecessary duplicates in some cases (I thought it'd be only theoretic problem). with it and new index for it, plan is following:

```
sqlite> EXPLAIN QUERY PLAN
   ...>   SELECT i.chat_item_id
   ...>   FROM chat_items i
   ...>   LEFT JOIN group_snd_item_statuses s ON s.chat_item_id = i.chat_item_id AND s.group_member_id = ?
   ...>   WHERE s.group_snd_item_status_id IS NULL
   ...>     AND i.user_id = ? AND i.group_id = ?
   ...>     AND i.include_in_history = 1
   ...>     AND i.item_deleted = 0
   ...>   ORDER BY i.item_ts DESC, i.chat_item_id DESC
   ...>   LIMIT ?;
QUERY PLAN
|--SEARCH i USING COVERING INDEX idx_chat_items_groups_history (user_id=? AND group_id=? AND include_in_history=? AND item_deleted=?)
`--SEARCH s USING COVERING INDEX idx_group_snd_item_statuses_chat_item_id_group_member_id (chat_item_id=? AND group_member_id=?) LEFT-JOIN
```

without it plan is:

```
sqlite> EXPLAIN QUERY PLAN
   ...>   SELECT chat_item_id
   ...>   FROM chat_items i
   ...>   WHERE user_id = ? AND group_id = ?
   ...>     AND include_in_history = 1
   ...>     AND item_deleted = 0
   ...>   ORDER BY item_ts DESC, chat_item_id DESC
   ...>   LIMIT ?;
QUERY PLAN
`--SEARCH i USING COVERING INDEX idx_chat_items_groups_history (user_id=? AND group_id=? AND include_in_history=? AND item_deleted=?)
```

so search over the table is the same

PR #5569, opened 2025-01-24
core: optimize group delete (delayed members cleanup)

spaced4ndy:


src/Simplex/Chat/Store/SQLite/Migrations/M20250123_indexes_group_members_cleanup.hs
```diff
@@ -0,0 +1,42 @@
...
+
+CREATE TABLE group_members_cleanup(
+  group_member_cleanup_id INTEGER PRIMARY KEY,
+  user_id INTEGER NOT NULL REFERENCES users ON DELETE CASCADE,
+  local_display_name TEXT NOT NULL,
+  contact_profile_id INTEGER NOT NULL,
+  created_at TEXT NOT NULL DEFAULT(datetime('now'))
+);
+
+CREATE INDEX idx_group_members_cleanup_user_id ON group_members_cleanup(user_id);
```
spaced4ndy, line 26: table and index are added compared to first pr

PR #5574, opened 2025-01-24
android, desktop: group member mentions

diogofcunha:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/group/GroupChatInfoView.kt
```diff
@@ -557,7 +558,7 @@ private fun AddMembersButton(titleId: StringResource, tint: Color = MaterialThem
 }
 
 @Composable
-private fun MemberRow(member: GroupMember, user: Boolean = false, onClick: (() -> Unit)? = null) {
+fun MemberRow(member: GroupMember, user: Boolean = false, infoPage: Boolean = true) {
```
diogofcunha, line 561: onClick was not used

apps/multiplatform/common/src/desktopMain/kotlin/chat/simplex/common/platform/ScrollableColumn.desktop.kt
```diff
@@ -135,9 +137,11 @@ actual fun LazyColumnWithScrollBarNoAppBar(
   // When scroll bar is dragging, there is no scroll event in nested scroll modifier. So, listen for changes on lazy column state
   // (only first visible row is useful because LazyColumn doesn't have absolute scroll position, only relative to row)
   val scrollBarDraggingState = remember { mutableStateOf(false) }
-  Box {
+  Box(contentAlignment = containerAlignment) {
```
diogofcunha, line 140: this component was no longer used, I did some changes to make it work as expected when we don't want bars to influence a simple lazy column with scrollbars on desktop

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
```diff
@@ -2255,6 +2321,20 @@ data class MemberReaction(
...
+@Serializable
+data class MentionedMemberInfo(
+  val groupMemberId: Long,
+  val displayName: String,
+  val localAlias: String?,
+  val memberRole: GroupMemberRole
+)
+
+@Serializable
+data class MentionedMember(
```
epoberezkin, line 2333:
```suggestion
data class CIMention(
```

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
```diff
@@ -2255,6 +2321,20 @@ data class MemberReaction(
   val reactionTs: Instant
 )
 
+@Serializable
+data class MentionedMemberInfo(
```
epoberezkin, line 2325:
```suggestion
data class CIMentionMember(
```

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/SimpleXAPI.kt
```diff
@@ -7460,3 +7469,14 @@ enum class MsgType {
...
+data class ChatNtfs(
+  val mode: MsgFilter,
+  val text: StringResource,
+  val icon: ImageResource,
+  val iconFilled: ImageResource
+) {
+  companion object {
+    val sampleData = ChatNtfs(MsgFilter.All, MR.strings.unmute_chat, MR.images.ic_notifications, MR.images.ic_notifications)
+  }
+}
```
epoberezkin, line 7482: eol

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
```diff
@@ -1762,6 +1811,31 @@ data class GroupInfo (
...
+      Mentions -> MR.strings.mute_chat
+      None -> MR.strings.mute_all_chat
+    }
+  }
+
+  private fun notificationModeIcon(m: MsgFilter, filled: Boolean): ImageResource {
+    return when (m) {
+      None -> (if (filled) MR.images.ic_notifications_off_filled else MR.images.ic_notifications_off)
+      All -> MR.images.ic_notifications
+      Mentions -> if (filled) MR.images.ic_notifications_mentions_filled else MR.images.ic_notifications_mentions
```
avently, line 1835: please name the icon file the same as on Material Icons website. We always preserve original name in order to know the source icon to be able to modify it if needed or to find a similar one

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/SimpleXAPI.kt
```diff
@@ -1018,12 +1020,13 @@ object ChatController {
     }
   }
 
-  suspend fun apiUpdateChatItem(rh: Long?, type: ChatType, id: Long, itemId: Long, mc: MsgContent, live: Boolean = false): AChatItem? {
-    val r = sendCmd(rh, CC.ApiUpdateChatItem(type, id, itemId, mc, live))
+  suspend fun apiUpdateChatItem(rh: Long?, type: ChatType, id: Long, itemId: Long, um: UpdatedMessage, live: Boolean = false): AChatItem? {
```
avently, line 1023: what are you trying to save by writting um versus updatedMessage? Characters? No need, just use names of variables where you understand the meaning without doing mind gymnastics.
um: UpdatedMessage -> updatedMessage: UpdatedMessage

epoberezkin: no need

epoberezkin: It's debatable whether you want names to repeat the meaning already conveyed in types.

It's less cognitive load to have short names and descriptive types, especially with short functions.

avently: type name doesn't represent variable meanting. You have hundreds of ChatItem but you don't know what's inside if you don't call it `chatItem: ChatItem, updatedChatItem: ChatItem`

avently: and when you read the code, you will only see `um = something` from the calling side without knowing what this `um` means because you will not see type name

avently: shorteing variable name is useful inside a local function when you need to call it often but not inside a parameter name. All globals should have clear name always

PR #5578, opened 2025-01-27
doc: delete for everyone: mention how it affects moderated messages

UnkwUsr:
I've not touched ios strings, because there localization depends on whole string, so need to edit all localization files. Hope someone know how to handle this better.


UnkwUsr:

PR #5579, opened 2025-01-27
core: optimize group delete - delayed group cleanup, delete unused contacts before deleting group

spaced4ndy:


src/Simplex/Chat/Library/Commands.hs
```diff
@@ -3536,6 +3517,41 @@ cleanupManager = do
...
+          withFastStore' $ \db -> deleteGroupMembers db user gInfo members'
+          withFastStore' $ \db -> deleteGroup db user gInfo
+          where
+            deleteUnusedContact :: DB.Connection -> ContactId -> IO (Either ChatError (Maybe StoreError, Maybe ContactId, [ConnId]))
+            deleteUnusedContact db contactId = runExceptT . withExceptT ChatErrorStore $ do
+              ct <- getContact db vr user contactId
+              ifM
+                ((directOrUsed ct ||) . isJust <$> liftIO (checkContactHasGroups db user ct gInfo))
+                (pure (Nothing, Nothing, []))
+                (deleteCt ct)
```
spaced4ndy, line 3544: Compared to old code, contacts are deleted before group and members records are removed. So if app is killed mid cleanup, group would still be checked for unused contacts.

Also instead of marking contact for deletion, contact is deleted immediately, without deleting its profile and ldn, as they will be deleted in next step when deleting members.

PR #5620, opened 2025-02-09
core, ui: report preference

epoberezkin:


tests/ChatTests/Groups.hs
```diff
@@ -1226,7 +1226,7 @@ testGroupMessageDelete =
   testChat3 aliceProfile bobProfile cathProfile $
     \alice bob cath -> do
       createGroup3 "team" alice bob cath
-      disableFullDeletion3 "team" alice bob cath
+      -- disableFullDeletion3 "team" alice bob cath
```
spaced4ndy, line 1229: is it commented because role is set to Nothing now? maybe remove altogether from tests then?

spaced4ndy (dismissed):

tests/ChatTests/Groups.hs
epoberezkin: it has to be set back to moderators only, but it was rushed - the role has to roll out first.

PR #5641, opened 2025-02-18
core: superpeers data model first draft

spaced4ndy:


src/Simplex/Chat/Store/SQLite/Migrations/M20250217_superpeers.hs
```diff
@@ -0,0 +1,42 @@
...
+import Database.SQLite.Simple (Query)
+import Database.SQLite.Simple.QQ (sql)
+
+m20250217_superpeers :: Query
+m20250217_superpeers =
+  [sql|
+CREATE TABLE superpeers(
+  super_peer_id INTEGER PRIMARY KEY,
+  address TEXT NOT NULL,
+  name TEXT NOT NULL, -- domain to match with operators to be part of name/address, or separate field?
```
spaced4ndy, line 14: should we save domains as separate field?

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -3251,18 +3260,23 @@ processChatCommand' vr = \case
       msgInfo <- withFastStore' (`getLastRcvMsgInfo` connId)
       CRQueueInfo user msgInfo <$> withAgent (`getConnectionQueueInfo` acId)
 
-protocolServers :: UserProtocol p => SProtocolType p -> ([Maybe ServerOperator], [UserServer 'PSMP], [UserServer 'PXFTP]) -> ([Maybe ServerOperator], [UserServer 'PSMP], [UserServer 'PXFTP])
-protocolServers p (operators, smpServers, xftpServers) = case p of
-  SPSMP -> (operators, smpServers, [])
-  SPXFTP -> (operators, [], xftpServers)
+-- TODO [superpeers] used for CLI specific APIs (also updatedServers below)
+-- TODO              - rework/add similar APIs for superpeers?
+-- TODO              - new "ServerType" type to replace ProtocolType that would unify servers and superpeers?
```
spaced4ndy, line 3265: Probably unnecessary to add another level of abstraction on chat level, and superpeers can be specially treated where necessary. Adding new protocol type is definitely wrong as it's defined in agent, and superpeers are chat level abstraction

epoberezkin: yes, better to keep them separate, not as servers

src/Simplex/Chat/Store/SQLite/Migrations/M20250217_superpeers.hs
epoberezkin: you mean extract? yes, probably

epoberezkin: Should they have profile too?

spaced4ndy: not sure they need it as a preset relay, owner/members will receive profile in group

spaced4ndy: > you mean extract? yes, probably

for servers we don't extract it - it's in host. depends on whether we want names with or without domains, e.g. superpeer1.simplex.im

src/Simplex/Chat/Operators.hs
```diff
@@ -306,6 +340,16 @@ newUserServer_ :: Bool -> Bool -> ProtoServerWithAuth p -> NewUserServer p
...
   UserServer {serverId = DBNewEntity, server, preset, tested = Nothing, enabled, deleted = False}
 
+presetSuperpeer :: Bool -> Text -> [Text] -> ConnReqContact -> NewUserSuperpeer
+presetSuperpeer = newSuperpeer_ True
+
+newSuperpeer :: Text -> [Text] -> ConnReqContact -> NewUserSuperpeer
+newSuperpeer = newSuperpeer_ False True
+
+newSuperpeer_ :: Bool -> Bool -> Text -> [Text] -> ConnReqContact -> NewUserSuperpeer
+newSuperpeer_ preset enabled name domains address =
```
epoberezkin, line 350:
```suggestion
newSuperpeer_ preset enabled name domains !address =
```

src/Simplex/Chat/Operators.hs
```diff
@@ -306,6 +340,16 @@ newUserServer_ :: Bool -> Bool -> ProtoServerWithAuth p -> NewUserServer p
 newUserServer_ preset enabled server =
   UserServer {serverId = DBNewEntity, server, preset, tested = Nothing, enabled, deleted = False}
 
+presetSuperpeer :: Bool -> Text -> [Text] -> ConnReqContact -> NewUserSuperpeer
+presetSuperpeer = newSuperpeer_ True
+
+newSuperpeer :: Text -> [Text] -> ConnReqContact -> NewUserSuperpeer
+newSuperpeer = newSuperpeer_ False True
```
epoberezkin, line 347: inline

src/Simplex/Chat/Operators.hs
epoberezkin: the point is to make it crash on invalid address - check that it does.

src/Simplex/Chat/Store/Postgres/Migrations/M20250217_superpeers.hs
```diff
@@ -0,0 +1,46 @@
...
+
+m20250217_superpeers :: Text
+m20250217_superpeers =
+  T.pack
+    [r|
+CREATE TABLE superpeers(
+  superpeer_id BIGINT PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  address TEXT NOT NULL,
+  name TEXT NOT NULL,
+  domains TEXT NOT NULL,
```
epoberezkin, line 17: why plural?

epoberezkin: and probably not needed - attaching to operators happens on the fly, based on pattern match anyway.

src/Simplex/Chat/Types.hs
```diff
@@ -129,13 +129,15 @@ data User = User
     sendRcptsContacts :: Bool,
     sendRcptsSmallGroups :: Bool,
     userMemberProfileUpdatedAt :: Maybe UTCTime,
-    uiThemes :: Maybe UIThemeEntityOverrides
+    uiThemes :: Maybe UIThemeEntityOverrides,
+    userSuperpeer :: Bool
```
epoberezkin, line 133: I don't think we need it in type - it's not really a "user" is it? or how do you plan to use it?

If the point of having it as a user to uniformly subscribe, then maybe, but then it's a separate query to get all users as opposed to other users for the UI.

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -308,19 +308,20 @@ processChatCommand cmd =
 processChatCommand' :: VersionRangeChat -> ChatCommand -> CM ChatResponse
 processChatCommand' vr = \case
   ShowActiveUser -> withUser' $ pure . CRActiveUser
-  CreateActiveUser NewUser {profile, pastTimestamp} -> do
+  CreateActiveUser NewUser {profile, pastTimestamp, userSuperpeer} -> do
```
epoberezkin, line 311: I don't think NewUser/CreateActiveUser should allow super peer - there is another api to create local super peer.

src/Simplex/Chat/Operators.hs
spaced4ndy: it did produce error on invalid addresses even without it, it's created as `either error id $ strDecode ...`. but adding it

src/Simplex/Chat/Operators.hs
```diff
@@ -306,6 +340,16 @@ newUserServer_ :: Bool -> Bool -> ProtoServerWithAuth p -> NewUserServer p
 newUserServer_ preset enabled server =
   UserServer {serverId = DBNewEntity, server, preset, tested = Nothing, enabled, deleted = False}
 
+presetSuperpeer :: Bool -> Text -> [Text] -> ConnReqContact -> NewUserSuperpeer
+presetSuperpeer = newSuperpeer_ True
```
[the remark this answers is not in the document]
spaced4ndy: copied from server, it's also not inline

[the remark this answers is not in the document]
spaced4ndy: adding for both

src/Simplex/Chat/Store/Postgres/Migrations/M20250217_superpeers.hs
spaced4ndy: why can't it have plural - same as servers?

> and probably not needed - attaching to operators happens on the fly, based on pattern match anyway.

I thought we agreed above to make it a separate field

src/Simplex/Chat/Store/Postgres/Migrations/M20250217_superpeers.hs
```diff
@@ -0,0 +1,46 @@
...
+import Text.RawString.QQ (r)
+
+m20250217_superpeers :: Text
+m20250217_superpeers =
+  T.pack
+    [r|
+CREATE TABLE superpeers(
+  superpeer_id BIGINT PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  address TEXT NOT NULL,
+  name TEXT NOT NULL,
```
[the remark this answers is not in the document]
spaced4ndy: right

> will there be a profile?

also answered above - as "preset relay" entity they don't need it (at least now there doesn't seem to be a reason for it), owner/members will receive profile in group

src/Simplex/Chat/Types.hs
spaced4ndy: it's for uniform subscription, yes, maybe other reasons too (I checked through code when added it as field - subscription was a reason, don't remember if there were others)

there's probably more work to do to make client work as supeerpeer, was going to filter out users then, will add TODO for now. Or maybe I thought even not filtering it out and instead filter it out in UI

src/Simplex/Chat/Library/Commands.hs
spaced4ndy: I created separate api type initially but almost all code could be reused so I changed it to parameter. There's still separate api string for CLI

src/Simplex/Chat/Store/Postgres/Migrations/M20250217_superpeers.hs
spaced4ndy: here it's used as just a mechanism to link to operators - for server its host directly implies its domain, but for superpeer domain of server's host in address doesn't necessarily belong to same operator (simply speaking, technically operator can host link to superpeer not on their own smp server)

spaced4ndy: so for example each superpeer can be preconfigured with all the same domains that a preconfigured operator has

PR #5698, opened 2025-03-03
rfc: user per-service per-device identities

ajsjsjdn:


epoberezkin:
what was the need for this PR? It's an old doc abandoned ~2y ago.

PR #123, opened 2021-11-07
add version command

epoberezkin:

PR #5726, opened 2025-03-07
core: group knocking

spaced4ndy:


docs/rfcs/2025-03-07-group-knocking.md
```diff
@@ -0,0 +1,38 @@
...
+- and group owners set up other arbitrary automated challenges or entry rules, while still being able to advertise groups in directory service.
+
+## Solution
+
+Group link host (further host), knowing group requires admin approval, would initially only introduce member pending approval to admins. Admins can connect with member for screening, meanwhile host would be forwarding messages as usual between connecting members. As a result of screening, pending member can either be removed, or approve by admins.
+
+Upon acceptance, for further member connections to not depend on availability of admins, host should not only forward acceptance message, but also introduce remaining members to now accepted joining member. Respectively, admins' clients should not introduce members for approved members who are not their invitees.
+
+For group owners to be able to set up alternative automated challenges, these are some possible alternatives:
+- We could add a new role `Approver`, so that instead of adding all \[possibly human\] admins, host would initially introduce only approvers.
```
epoberezkin, line 18: I think it should be a flag, not role.

epoberezkin: that is, orthogonal to a role

docs/rfcs/2025-03-07-group-knocking.md
```diff
@@ -0,0 +1,38 @@
...
+- It could be an orthogonal to role member setting (would require protocol extension).
+- List of "approver" member IDs could be communicated to host client.
```
epoberezkin, line 20: Possibly, we need to think about "member profile" where this flag would reside.

docs/rfcs/2025-03-07-group-knocking.md
```diff
@@ -0,0 +1,38 @@
...
+
+### Implementation details draft
+
+Host needs to have knowledge whether to automatically accept, or only introduce admins/approvers.
+
+```sql
+ALTER TABLE group_profiles ADD COLUMN approval TEXT; -- comma separated member IDs; null - automatic introduction
```
epoberezkin, line 27: This feels wrong. Group profile is accessible to members, and we should not advertise which admins manage knocking.

docs/rfcs/2025-03-07-group-knocking.md
```diff
@@ -0,0 +1,38 @@
...
+
+-- or
+
+ALTER TABLE group_profiles ADD COLUMN approval INTEGER; -- if based on `Approver` role
+```
+
+Alternatively, a different extension of protocol could be done in order to communicate group approval rule from owner to host outside of group profile (special messages).
```
epoberezkin, line 34: That also feels wrong, but maybe it's the least bad option.

docs/rfcs/2025-03-07-group-knocking.md
```diff
@@ -0,0 +1,38 @@
...
+
+Admins/approvers need to have separate conversation per pending member, requires adding scope to chat items.
```
epoberezkin, line 36: Model that.

docs/rfcs/2025-03-07-group-knocking.md
```diff
@@ -0,0 +1,38 @@
...
+
+Host to have specific processing of forwarded `XGrpLinkAcpt` - continue introduction of remaining members.
```
epoberezkin, line 38: don't we already have it? Why is it forwarded - you mean "forwarding"?

spaced4ndy: I meant it has to be forwarded, not issued directly from host - and host has to process it in a special way by starting introductions

src/Simplex/Chat/Protocol.hs
```diff
@@ -243,13 +243,22 @@ instance ToJSON SharedMsgId where
   toJSON = strToJSON
   toEncoding = strToJEncoding
 
+data MsgScope
+  = MSGroup
+  | MSMember {memberId :: MemberId} -- Admins can use any member id; members can use only their own id
```
spaced4ndy, line 248: how is it possible to understand from this name that scope is conversation with admins

src/Simplex/Chat/Messages.hs
```diff
@@ -75,44 +83,68 @@ chatTypeStr = \case
...
+        unanswered :: Bool
+      }
+  deriving (Show)
+
+toGCS :: GroupChatScopeInfo -> GroupChatScope
+toGCS = \case
+  GCSIGroup -> GCSGroup
+  GCSIMemberSupport {groupMember_} -> GCSMemberSupport $ groupMemberId' <$> groupMember_
+
+toMsgScope :: GroupInfo -> GroupChatScopeInfo -> MsgScope
```
epoberezkin, line 120: not used anywhere

src/Simplex/Chat/Messages.hs
```diff
@@ -186,22 +216,33 @@ isUserMention ChatItem {meta = CIMeta {userMention}} = userMention
...
 data CCIDirection c = forall d. MsgDirectionI d => CCID (SMsgDirection d) (CIDirection c d)
 
 data ACIDirection = forall c d. (ChatTypeI c, MsgDirectionI d) => ACID (SChatType c) (SMsgDirection d) (CIDirection c d)
 
 data JSONCIDirection
   = JCIDirectSnd
   | JCIDirectRcv
-  | JCIGroupSnd
-  | JCIGroupRcv {groupMember :: GroupMember}
+  | JCIGroupSnd {groupChatScopeInfo :: GroupChatScopeInfo}
```
epoberezkin, line 244:
```suggestion
  | JCIGroupSnd {chatScope :: GroupChatScopeInfo}
```

src/Simplex/Chat/Protocol.hs
```diff
@@ -1086,8 +1100,8 @@ appJsonToCM AppMessageJson {v, msgId, event, params} = do
       XGrpLinkInv_ -> XGrpLinkInv <$> p "groupLinkInvitation"
       XGrpLinkReject_ -> XGrpLinkReject <$> p "groupLinkRejection"
       XGrpLinkMem_ -> XGrpLinkMem <$> p "profile"
-      XGrpLinkAcpt_ -> XGrpLinkAcpt <$> p "role"
-      XGrpMemNew_ -> XGrpMemNew <$> p "memberInfo"
+      XGrpLinkAcpt_ -> XGrpLinkAcpt <$> p "role" <*> opt "memberId"
+      XGrpMemNew_ -> XGrpMemNew <$> p "memberInfo" <*> opt "msgScope"
```
epoberezkin, line 1104:
```suggestion
      XGrpMemNew_ -> XGrpMemNew <$> p "memberInfo" <*> opt "scope"
```

src/Simplex/Chat/Protocol.hs
```diff
@@ -1150,8 +1164,8 @@ chatToAppMessage ChatMessage {chatVRange, msgId, chatMsgEvent} = case encoding @
       XGrpLinkInv groupLinkInv -> o ["groupLinkInvitation" .= groupLinkInv]
       XGrpLinkReject groupLinkRjct -> o ["groupLinkRejection" .= groupLinkRjct]
       XGrpLinkMem profile -> o ["profile" .= profile]
-      XGrpLinkAcpt role -> o ["role" .= role]
-      XGrpMemNew memInfo -> o ["memberInfo" .= memInfo]
+      XGrpLinkAcpt role memberId -> o $ ("memberId" .=? memberId) ["role" .= role]
+      XGrpMemNew memInfo msgScope -> o $ ("msgScope" .=? msgScope) ["memberInfo" .= memInfo]
```
epoberezkin, line 1168:
```suggestion
      XGrpMemNew memInfo scope -> o $ ("msgScope" .=? scope) ["memberInfo" .= memInfo]
```

src/Simplex/Chat/Protocol.hs
```diff
@@ -335,8 +347,8 @@ data ChatMsgEvent (e :: MsgEncoding) where
   XGrpLinkInv :: GroupLinkInvitation -> ChatMsgEvent 'Json
   XGrpLinkReject :: GroupLinkRejection -> ChatMsgEvent 'Json
   XGrpLinkMem :: Profile -> ChatMsgEvent 'Json
-  XGrpLinkAcpt :: GroupMemberRole -> ChatMsgEvent 'Json
-  XGrpMemNew :: MemberInfo -> ChatMsgEvent 'Json
+  XGrpLinkAcpt :: GroupMemberRole -> Maybe MemberId -> ChatMsgEvent 'Json -- Maybe MemberId for compatibility
```
epoberezkin, line 350: compatibility with what? wasn't it just added for directory approvals?

src/Simplex/Chat/Protocol.hs
```diff
@@ -243,6 +248,13 @@ instance ToJSON SharedMsgId where
   toJSON = strToJSON
   toEncoding = strToJEncoding
 
+data MsgScope
+  = MSGroup
+  | MSMember {memberId :: MemberId} -- Admins can use any member id; members can use only their own id
+  deriving (Eq, Show)
+
+$(JQ.deriveJSON (sumTypeJSON $ dropPrefix "MS") ''MsgScope)
```
epoberezkin, line 256: you can't use platform-dependent encoding in protocol... tsk tsk.

```suggestion
$(JQ.deriveJSON (taggedObjectJSON $ dropPrefix "MS") ''MsgScope)
```

epoberezkin: which means that the instance has to be manually derived in swift

src/Simplex/Chat/Controller.hs
```diff
@@ -301,7 +301,7 @@ data ChatCommand
   | APIGetAppSettings (Maybe AppSettings)
   | APIGetChatTags UserId
   | APIGetChats {userId :: UserId, pendingConnections :: Bool, pagination :: PaginationByTime, query :: ChatListQuery}
-  | APIGetChat ChatRef (Maybe MsgContentTag) ChatPagination (Maybe String)
+  | APIGetChat ChatRef (Maybe GroupChatFilter) ChatPagination (Maybe String)
```
epoberezkin, line 304: I don't understand why it is needed. If you don't want to support MsgContentTag filter in support chats, then you simply pass nothing here, but chatscope is already in chat ref, so why it's here too?

src/Simplex/Chat/Controller.hs
```diff
@@ -363,6 +363,9 @@ data ChatCommand
   | APIRemoveMembers {groupId :: GroupId, groupMemberIds :: Set GroupMemberId, withMessages :: Bool}
   | APILeaveGroup GroupId
   | APIListMembers GroupId
+  | APIMemberSupportChats GroupId
```
epoberezkin, line 366: I disagree with this being a separate API to list these chats and not just return the list of members, where the presence of support chat and any statistics is indicated by some object.

The argument that there are fewer supports chats than members is invalid as eventually we'd need to support pagination with search, sorting and filtering anyway, so the presence of support chat is just another filter.

src/Simplex/Chat/Controller.hs
```diff
@@ -749,6 +755,8 @@ data ChatResponse
   | CRUserJoinedGroup {user :: User, groupInfo :: GroupInfo, hostMember :: GroupMember}
   | CRJoinedGroupMember {user :: User, groupInfo :: GroupInfo, member :: GroupMember}
   | CRJoinedGroupMemberConnecting {user :: User, groupInfo :: GroupInfo, hostMember :: GroupMember, member :: GroupMember}
+  | CRMemberAccepted {user :: User, groupInfo :: GroupInfo, member :: GroupMember}
+  | CRMemberAcceptedByOther {user :: User, groupInfo :: GroupInfo, acceptingMember :: GroupMember, member :: GroupMember}
```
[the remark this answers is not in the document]
epoberezkin: Most likely you want CRMemberAccepted {user :: User, groupInfo :: GroupInfo, acceptedBy :: Maybe GroupMember, member :: GroupMember} with Nothing for current user accepting? The meaning is not clear here

src/Simplex/Chat/Controller.hs
```diff
@@ -658,6 +661,9 @@ data ChatResponse
   | CRWelcome {user :: User}
   | CRGroupCreated {user :: User, groupInfo :: GroupInfo}
   | CRGroupMembers {user :: User, group :: Group}
+  | CRMemberSupportChats {user :: User, groupInfo :: GroupInfo, members :: [GroupMember]}
```
epoberezkin, line 664: same here - it's the same response as CRGroupMembers

src/Simplex/Chat/Controller.hs
```diff
@@ -898,16 +906,20 @@ logResponseToFile = \case
...
   deriving (Eq, Show)
 
 sendToChatRef :: SendRef -> ChatRef
 sendToChatRef = \case
-  SRDirect cId -> ChatRef CTDirect cId
-  SRGroup gId _ -> ChatRef CTGroup gId
+  SRDirect cId -> ChatRef CRTDirect cId
+  SRGroup gId gcs -> ChatRef (CRTGroup gcs) gId
+
+data GroupChatFilter
```
epoberezkin, line 919: not needed?

src/Simplex/Chat/Store/SQLite/Migrations/M20250310_group_scope.hs
```diff
@@ -0,0 +1,45 @@
...
+-- TODO [knocking] TBC schema
+-- TODO            - group_members.support_chat_unanswered - don't persist, calculate on the fly?
+-- TODO            - review indexes (drop idx_chat_items_groups_item_ts?)
+m20250310_group_scope :: Query
+m20250310_group_scope =
+  [sql|
+ALTER TABLE group_members ADD COLUMN support_chat_ts TEXT;
+ALTER TABLE group_members ADD COLUMN support_chat_unanswered INTEGER;
+
+ALTER TABLE chat_items ADD COLUMN group_scope_group_member_id INTEGER REFERENCES group_members(group_member_id) ON DELETE CASCADE;
```
epoberezkin, line 22: the name is confusing - what suggests that it is a support chat? Also, for the chat of the current member with admins it would be Nothing?

It probably needs two fields:

group_chat_scope TEXT -- NULL - group, "support" for member support
group_chat_scope_member_id INTEGER -- NULL for member chat with admins, ID for admins' chat with member. You cannot have just one field, as the members can have both categories - their chat with admins, and their participation in support chat as admins (at a later point once promoted to admins).

src/Simplex/Chat/Types.hs
```diff
@@ -825,7 +825,14 @@ data GroupMember = GroupMember
...
     createdAt :: UTCTime,
-    updatedAt :: UTCTime
+    updatedAt :: UTCTime,
+    supportChat :: Maybe GroupMemberSupportChat
+  }
+  deriving (Eq, Show)
+
+data GroupMemberSupportChat = GroupMemberSupportChat
+  { chatTs :: UTCTime,
+    unanswered :: Bool
```
epoberezkin, line 835: based on the discussion, also needs mentions and unread counts

epoberezkin: Mentions are actually important, as one admin can comment on the question and effectively ask that another admin answers.

src/Simplex/Chat/Types/Preferences.hs
```diff
@@ -152,6 +152,7 @@ data GroupFeature
   | GFSimplexLinks
   | GFReports
   | GFHistory
+  | GFNewMemberReview
```
epoberezkin, line 155: something feels wrong about it being a preference

epoberezkin: I am thinking about the time when member screening moves to chat relays, and while it has to be part of group profile, it feels like a separate space from preferences and requiring multiple toggles. At the same time it is unlikely to have role associated.

So we probably need profile property:

```
admission :: GroupMemberAdmission 

data GroupMemberAdmission = GroupMemberAdmission 
  { -- names :: AdmissionFilter,
    -- captcha :: AdmissionFilter,
    review :: AdmissionFilter
  }
  
 data AdmissionCriteria = ACAll | ACNoProfileImage -- etc. e.g. unverified profiles
```

effectively borrow the type that is now present in directory, and the whole admission process deserves a separate page, that initially would have one toggle.

src/Simplex/Chat/Types/Shared.hs
```diff
@@ -51,11 +51,6 @@ instance ToJSON GroupMemberRole where
 
 data GroupAcceptance = GAAccepted | GAPending deriving (Eq, Show)
```
epoberezkin, line 52: not clear what happened here

tests/ChatTests/Groups.hs
```diff
@@ -3023,8 +3034,301 @@ testGLinkManualAcceptMember =
...
+      alice ##> "/create link #team"
+      gLink <- getGroupLink alice "team" GRMember True
+      eve ##> ("/c " <> gLink)
+      eve <## "connection request sent!"
+      alice <## "eve (Eve): accepting request to join group #team..."
+      concurrentlyN_
+        [ alice <## "#team: eve connected and pending review",
+          eve
+            <### [ "#team: joining the group...",
+                   "#team: you joined the group, pending review",
```
epoberezkin, line 3070: We need some other language here, something like this:

#team: connecting to group moderators for admission to group

tests/ChatTests/Groups.hs
```diff
@@ -6610,36 +6914,78 @@ testUpdatedMentionNames = do
...
-    cath <# "#team alice> 5"
+    bob #> "#team 2"
+    [alice, cath, dan] *<# "#team bob> 2"
+
+    threadDelay 1000000
+
+    alice ##> "/_send #1 @support-2 text 3"
+    alice <# "#team <<support: bob>> 3"
+    bob <# "#team alice> <<support>> 3"
+    dan <# "#team alice> <<support: bob>> 3"
```
epoberezkin, line 6959:
```suggestion
    alice <# "#team (support: bob) 3"
    bob <# "#team (support) alice> 3"
    dan <# "#team (support: bob) alice> 3"
```

otherwise it's inconsistent, you are putting message sender in the middle of chat "name"

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -4256,10 +4349,22 @@ chatCommandP =
...
+    groupChatRefP = do
+      gId <- A.char '#' *> A.decimal
+      gcs <- A.space *> gcScopeP <|> pure GCSGroup
+      pure $ ChatRef (CRTGroup gcs) gId
     sendRefP =
       (A.char '@' $> SRDirect <*> A.decimal)
-        <|> (A.char '#' $> SRGroup <*> A.decimal <*> optional (" @" *> A.decimal))
+        <|> (A.char '#' $> SRGroup <*> A.decimal <*> (A.space *> gcScopeP <|> pure GCSGroup))
+    gcScopeP =
+      ("@group" $> GCSGroup)
```
epoberezkin, line 4365: This is unnecessarily complex syntax. Default scope can remain as is - `#123`

For support scope I suggest `#123(456)` and `#123()`.

The proposed syntax does not keep it together.

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -3942,6 +4028,9 @@ chatCommandP =
       "/_remove #" *> (APIRemoveMembers <$> A.decimal <*> _strP <*> (" messages=" *> onOffP <|> pure False)),
       "/_leave #" *> (APILeaveGroup <$> A.decimal),
       "/_members #" *> (APIListMembers <$> A.decimal),
+      "/_member support chats #" *> (APIMemberSupportChats <$> A.decimal),
```
epoberezkin, line 4031: see comment in controller

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -3400,10 +3457,39 @@ processChatCommand' vr = \case
       CRQueueInfo user msgInfo <$> withAgent (`getConnectionQueueInfo` acId)
     withSendRef :: ChatRef -> (SendRef -> CM ChatResponse) -> CM ChatResponse
     withSendRef chatRef a = case chatRef of
-      ChatRef CTDirect cId -> a $ SRDirect cId
-      ChatRef CTGroup gId -> a $ SRGroup gId Nothing
+      ChatRef CRTDirect cId -> a $ SRDirect cId
+      ChatRef (CRTGroup gcs) gId -> a $ SRGroup gId gcs
       _ -> throwChatError $ CECommandError "not supported"
 
+getGCSDependencies :: VersionRangeChat -> User -> GroupInfo -> GroupChatScope -> VersionChat -> CM (GroupChatScopeInfo, MsgScope, [GroupMember], Int)
```
epoberezkin, line 3464: the name is cryptic - what does it do?

epoberezkin: it seems like it returns a bunch of things that must be derived locally and not in function - e.g., you should never return array and its length - length can be taken locally at a point of use.

Same for MsgScope - it seems like it can be trivially derived from GroupChatScopeInfo locally.

You should not be creating functions to simply do trivial conversions.

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -3169,46 +3228,44 @@ processChatCommand' vr = \case
...
+        where
+          hasReport = any (\(ComposedMessage {msgContent}, _, _, _) -> isReport msgContent) cmrs
+          modsCompatVersion = if hasReport then contentReportsVersion else groupKnockingVersion
+    sendGroupContentMessages_ :: User -> GroupInfo -> GroupChatScope -> GroupChatScopeInfo -> MsgScope -> [GroupMember] -> Int -> Bool -> Maybe Int -> NonEmpty ComposedMessageReq -> CM ChatResponse
+    sendGroupContentMessages_ user gInfo@GroupInfo {groupId, membership} gcs gcsi msgScope recipients numFileInvs live itemTTL cmrs = do
       forM_ allowedRole $ assertUserGroupRole gInfo
       assertGroupContentAllowed
       processComposedMessages
       where
+        allowedRole :: Maybe GroupMemberRole
```
epoberezkin, line 3262: why is it moved to the bottom? return it back.

epoberezkin: also, I think we should allow support chats to observers. And not check role for host, as before

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -2152,34 +2177,48 @@ processChatCommand' vr = \case
...
       assertUserGroupRole gInfo $ max GRAdmin maxRole
       (errs1, deleted1) <- deleteInvitedMems user invitedMems
-      (errs2, deleted2, acis2) <- deleteMemsSend user gInfo members currentMems
-      rs <- forM pendingMems $ \m -> deleteMemsSend user gInfo [m] [m] -- TODO [knocking]
-      let (errs3, deleted3, acis3) = concatTuples rs
-          acis = acis2 <> acis3
-          errs = errs1 <> errs2 <> errs3
+      let recipients = filter memberCurrent members
+      (errs2, deleted2, acis2) <- deleteMemsSend user gInfo GCSIGroup recipients currentMems
+      rs3 <- forM pendingApprvMems $ \m -> do
```
epoberezkin, line 2188: replace with foldM to avoid concatTuples

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -2152,34 +2177,48 @@ processChatCommand' vr = \case
...
+        gcsi <- liftIO $ memberSupportGCSI m (Just m)
+        deleteMemsSend user gInfo gcsi [m] [m]
+      rs4 <-
+        if not (null pendingRvwMems)
+          then do
+            let moderators = filter (\GroupMember {memberRole} -> memberRole >= GRModerator) members
+            forM pendingRvwMems $ \m -> do
```
epoberezkin, line 2195: replace with foldM to avoid concatTuples

src/Simplex/Chat/Library/Commands.hs
epoberezkin: probably best to make it a function with some name - what does it return

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -2025,18 +2027,39 @@ processChatCommand' vr = \case
...
-      Just mConn -> do
-        let msg = XGrpLinkAcpt role
-        void $ sendDirectMemberMessage mConn msg groupId
-        m' <- withFastStore' $ \db -> updateGroupMemberAccepted db user m role
-        introduceToGroup vr user gInfo m'
-        pure $ CRJoinedGroupMember user gInfo m'
-      _ -> throwChatError CEGroupMemberNotActive
+    assertUserGroupRole gInfo GRModerator
+    case memberStatus m of
+      GSMemPendingApproval | memberCategory m == GCInviteeMember -> do -- only host can approve
```
epoberezkin, line 2034: it may be better to have different APIs - approve/accept

epoberezkin: where approve can trigger admission (maybe admit is a better term than accept).

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -575,25 +577,16 @@ processChatCommand' vr = \case
...
+      gInfo <- withFastStore $ \db -> getGroupInfo db vr user gId
+      let mc = MCReport reportText reportReason
           cm = ComposedMessage {fileSource = Nothing, quotedItemId = Just reportedItemId, msgContent = mc, mentions = M.empty}
-      when (null ms') $ throwChatError $ CECommandError "no moderators support receiving reports"
-      let numFileInvs = length $ filter memberCurrent ms'
-      sendGroupContentMessages_ user gInfo Nothing ms' numFileInvs False Nothing [composedMessageReq cm]
-    where
-      compatibleModerator GroupMember {activeConn, memberChatVRange} =
-        maxVersion (maybe memberChatVRange peerChatVRange activeConn) >= contentReportsVersion
+      sendGroupContentMessages user gInfo (GCSMemberSupport Nothing) False Nothing [composedMessageReq cm]
```
epoberezkin, line 583: Does it mean that report won't be visible to admins and to members in the group itself? Not sure it is a good idea - they would be more likely to be missed, and they are higher priority than support chats.

Btw, what I don't see at all that support chats are allowed in the group - this should be in the preferences, unlike admission criteria.

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -762,54 +762,68 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
-          _ -> do
-            withStore' $ \db -> do
-              updateGroupMemberStatus db userId m GSMemConnected
-              unless (memberActive membership) $
-                updateGroupMemberStatus db userId membership GSMemConnected
-            -- possible improvement: check for each pending message, requires keeping track of connection state
-            unless (connDisabled conn) $ sendPendingGroupMessages user m conn
-            pure GSMemConnected
+      CON _pqEnc -> unless (memberStatus m == GSMemRejected || memberStatus membership == GSMemRejected) $ do
+        -- TODO [knocking] send pending messages after accepting?
```
epoberezkin, line 766: why removed?

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -762,54 +762,68 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+            memberConnectedChatItem gInfo gcsi m
+            toView $ CRJoinedGroupMember user gInfo m {memberStatus = mStatus}
             let Connection {viaUserContactLink} = conn
             when (isJust viaUserContactLink && isNothing (memberContactId m)) sendXGrpLinkMem
             when (connChatVersion < batchSend2Version) sendGroupAutoReply
-            unless (status' == GSMemPendingApproval) $ introduceToGroup vr user gInfo m
+            case mStatus of
+              GSMemPendingApproval -> pure ()
+              -- edge case: reviews were turned off mid connection;
+              -- options: proceed to review as declared, or introduce to group and send XGrpLinkAcpt;
```
epoberezkin, line 800: it is also more sensible - that review is disabled, does not need to auto-admit all members. We have multi-select, it could be a bulk action.

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -1353,18 +1372,15 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
         throwChatError $ CEFileRcvChunk err
 
-    memberConnectedChatItem :: GroupInfo -> GroupMember -> CM ()
-    memberConnectedChatItem gInfo m =
+    memberConnectedChatItem :: GroupInfo -> GroupChatScopeInfo -> GroupMember -> CM ()
+    memberConnectedChatItem gInfo gcsi m =
       -- ts should be broker ts but we don't have it for CON
-      createInternalChatItem user (CDGroupRcv gInfo m) (CIRcvGroupEvent RGEMemberConnected) Nothing
-
-    groupDescriptionChatItem :: GroupInfo -> GroupMember -> Text -> CM ()
```
epoberezkin, line 1361: why removed

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -1667,13 +1688,25 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
           ci' <- blockedMember m ci $ withStore' $ \db -> markGroupChatItemBlocked db user gInfo ci
           reactions <- maybe (pure []) (\sharedMsgId -> withStore' $ \db -> getGroupCIReactions db gInfo memberId sharedMsgId) sharedMsgId_
           groupMsgToView gInfo ci' {reactions}
 
-    groupMessageUpdate :: GroupInfo -> GroupMember -> SharedMsgId -> MsgContent -> Map MemberName MsgMention -> RcvMessage -> UTCTime -> Maybe Int -> Maybe Bool -> CM ()
-    groupMessageUpdate gInfo@GroupInfo {groupId} m@GroupMember {groupMemberId, memberId} sharedMsgId mc mentions msg@RcvMessage {msgId} brokerTs ttl_ live_
+    getMsgScopeGCSI :: GroupInfo -> GroupMember -> Maybe MsgScope -> CM GroupChatScopeInfo
+    getMsgScopeGCSI gInfo@GroupInfo {membership} m = \case
+      -- first two cases check scope if member/membership is pending
+      Nothing -> liftIO $ getNonMsgGCSI gInfo m -- correct scope if member is of old version or incorrectly sends no scope
```
epoberezkin, line 1700: `getNonMsgGCSI` name is wrong, it does not explain what it does, it only says what it does not.

And the whole logic is upside down. You are not determining the scope based on the message in most cases. You first determine scope based on the member status, and if the member scope is undeterminable, only then you look at message scope.

In general case when member may be allowed in multiple scopes this construction may change, but currently we only have two scenarios - member is only allowed support scope, or member is allowed any scope.

So the function should be getMemberChatScope :: GroupMember -> Maybe GroupChatScopeInfo

and getGroupMsgScope :: GroupInfo -> GroupMember -> Maybe MsgScope -> CM GroupChatScopeInfo
(and not getMsgScopeGCSI which is unclear and cryptic).

And inside it will first check member's scope, and if it's nothing then it'll look at MsgScope.

epoberezkin: as getNonMsgGCSI is used all over the place, maybe it has to be 

getMemberChatScope_ :: GroupMember -> Maybe GroupChatScopeInfo

and

getMemberChatScope :: GroupMember -> GroupChatScopeInfo (same as the previous but defaults to group)

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -2064,15 +2094,50 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
-      membership' <- withStore' $ \db -> do
-        updateGroupMemberStatus db userId m GSMemConnected
-        updateGroupMemberAccepted db user membership role
-      let m' = m {memberStatus = GSMemConnected}
-      toView $ CRUserJoinedGroup user gInfo {membership = membership'} m'
-      let connectedIncognito = memberIncognito membership
-      probeMatchingMemberContact m' connectedIncognito
+    xGrpLinkAcpt :: GroupInfo -> GroupMember -> GroupMemberRole -> Maybe MemberId -> CM ()
+    xGrpLinkAcpt gInfo@GroupInfo {groupId, membership} m role memberId_ = case memberId_ of
+      Nothing -> processUserAccepted -- backwards compatibility
```
epoberezkin, line 2099: maybe we don't need it

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -1259,6 +1288,27 @@ parseChatMessage conn s = do
     errType = CEInvalidChatMessage conn Nothing (safeDecodeUtf8 s)
 {-# INLINE parseChatMessage #-}
 
+getLocalGCSI :: GroupInfo -> IO GroupChatScopeInfo
```
epoberezkin, line 1291: Let's just call it chatScope in names? GCSI is cryptic

src/Simplex/Chat/Protocol.hs
spaced4ndy: yes - but host (directory) sends it to invitee without MemberId now. It could be non optional otherwise

src/Simplex/Chat/Protocol.hs
spaced4ndy: this type is not needed in swift

src/Simplex/Chat/Controller.hs
spaced4ndy: yes, it's redundant, there is a comment - didn't get to it (I added scope to chat ref after this was already working)

src/Simplex/Chat/Controller.hs
spaced4ndy: maybe

src/Simplex/Chat/Controller.hs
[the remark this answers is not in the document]
spaced4ndy: one is response to api, another is event on receiving protocol message. could be united into one event

src/Simplex/Chat/Store/SQLite/Migrations/M20250310_group_scope.hs
spaced4ndy: it's differentiated by joined member records (whether it's membership or member), where necessary. and when reading items list we don't need to differentiate because scope is known (and so group member id). main reason was to add single new index and not two, but as discussed proposed fields can also work with single index

src/Simplex/Chat/Types.hs
spaced4ndy: tbc maintain in store or calculate on read

src/Simplex/Chat/Types/Shared.hs
spaced4ndy: if this comment refers to deletion of instances - they were not used (initially this field were added to db, then removed from db in same pr, and also their encoding was wrong for db - mismatched text/blob)

tests/ChatTests/Groups.hs
spaced4ndy: it's quite messy there IIRC with they way from/toGroup are defined.. will see if it's easy fix

src/Simplex/Chat/Library/Commands.hs
spaced4ndy: I think we should use "support" for support here, to easier add new scope if required (e.g. admins to each other, or direct message to other member), and not have to change other apis or leave their syntax as "more general". also it's syntax for ui so why does it matter that it's a bit longer? also `#123()` is completely unintelligible

src/Simplex/Chat/Library/Commands.hs
spaced4ndy: because it became bigger and unnecessarily breaks the flow of function

src/Simplex/Chat/Library/Commands.hs
spaced4ndy: we do send the same protocol message though. but other logic is completely different, yes

spaced4ndy: maybe we should add protocol message too, then, XGrpLinkAdmit, and remove optional memberId from XGrpLinkAcpt (it will return to being only for invitee)

src/Simplex/Chat/Library/Commands.hs
spaced4ndy: yes, and they wouldn't be missable because there's also panel in top of chat, and also they wouldn't create confusion for both members and admins that other members don't see it, and the need for this header on chat items;

there's no such preference

src/Simplex/Chat/Library/Subscriber.hs
spaced4ndy: it's reverted to how it was before - I think it was a bug from when we added acceptance that this also got inside the condition, because it's then never sent anywhere.

src/Simplex/Chat/Library/Subscriber.hs
spaced4ndy: it had a single call site and only created indirection, it's moved to maybeCreateGroupDescrLocal instead with is reused

src/Simplex/Chat/Library/Subscriber.hs
spaced4ndy: then members with previous version wouldn't be able to join groups. alternatively we could add XGrpLinkAdmit as separate event from moderators to host which does make sense (see api comment), and keep this as it was without optional member id

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -1237,27 +1242,27 @@ getContactNavInfo_ db User {userId} Contact {contactId} afterCI = do
...
 getGroupChatScopeInfo' db vr user g = \case
-  Just (GCFChatScope gcScope) -> getGroupChatScopeInfo_ db vr user g gcScope
-  _ -> pure Nothing
+  Just scope -> Just <$> getGroupChatScopeInfo_ db vr user g scope
+  Nothing -> pure Nothing
 
-getGroupChatScopeInfo_ :: DB.Connection -> VersionRangeChat -> User -> GroupInfo -> Maybe GroupChatScope -> ExceptT StoreError IO (Maybe GroupChatScopeInfo)
-getGroupChatScopeInfo_ db vr user GroupInfo {membership} = mapM $ \case
+getGroupChatScopeInfo_ :: DB.Connection -> VersionRangeChat -> User -> GroupInfo -> GroupChatScope -> ExceptT StoreError IO GroupChatScopeInfo
+getGroupChatScopeInfo_ db vr user GroupInfo {membership} = \case
```
epoberezkin, line 1265: why we are not patching here?

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -1269,23 +1274,63 @@ getGroupChatScopeInfo_ db vr user GroupInfo {membership} = mapM $ \case
...
-getGroupChatItemIDs db User {userId} GroupInfo {groupId, membership} groupChatFilter range count search = case groupChatFilter of
-  Nothing -> idsQuery (baseCond <> " AND group_scope_group_member_id IS NULL ") (userId, groupId)
-  Just (GCFChatScope Nothing) -> idsQuery (baseCond <> " AND group_scope_group_member_id IS NULL ") (userId, groupId)
-  Just (GCFChatScope (Just (GCSMemberSupport gmId_))) ->
-    let gmId = fromMaybe (groupMemberId' membership) gmId_
-     in idsQuery (baseCond <> " AND group_scope_group_member_id = ? ") (userId, groupId, gmId)
-  Just (GCFMsgContentTag mcTag) -> idsQuery (baseCond <> " AND msg_content_tag = ? ") (userId, groupId, mcTag)
+getGroupChatItemIDs :: DB.Connection -> User -> GroupInfo -> Maybe GroupChatScopeInfo -> Maybe MsgContentTag -> GroupItemIDsRange -> Int -> String -> ExceptT StoreError IO [ChatItemId]
+getGroupChatItemIDs db User {userId} GroupInfo {groupId} scopeInfo_ contentFilter range count search = case (scopeInfo_, contentFilter) of
+  (Nothing, Nothing) ->
```
epoberezkin, line 1312: Looks like it should be a layer in this query constructor, the condition on scope is independent of the condition on content (and then you don't need to fail).

tests/ChatTests/Groups.hs
```diff
@@ -3262,24 +3262,24 @@ testGLinkApproveThenReviewMember =
       eve <## "bad chat command: not current member"
 
       -- pending member and moderators can send messages to each other
-      alice ##> "/_send #1 @support-5 text 11"
-      alice <# "#team <<support: eve>> 11"
-      [cath, dan] *<# "#team alice> <<support: eve>> 11"
-      eve <# "#team alice> <<support>> 11"
+      alice ##> "/_send #1(_support:5) text 11"
```
epoberezkin, line 3265: why "_support" and not "support"?

spaced4ndy: to differentiate from cli api that will use `(support: bob)`

spaced4ndy: (to avoid ambiguity for names consisting of digits)

Issue #4922, opened 2024-09-21
[Feature]: linux arm64 support


coll4ctg43:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Platform

Linux

### Feature

Hi, I couldn't fork it and I don't know if this is the right place to communicate it here but I fixed the install.sh file that doesn't distinguish between the various linux architectures and mistakenly installs an x86_64 binary even on other architectures and I modified the Dockerfile to compile a binary for arm64 from other architectures. Also but I guess not a safe way to release binaries I wanted to share the binary I created. It can all be found in my simplex-chat-arm64 repository.

daic11:
Yes please add arm support

coll4ctg43:
close for related #3182

Masked-Witch:
Also modifying the GUI to adjust to smaller screens automatically would make it usable on linux phones

Anderhar:
The [Flatpak](https://flathub.org/apps/chat.simplex.simplex) version is rated as aarch64-compatible to date.

Issue #1274, opened 2022-10-31
ERROR: - libcrypto.so.1.1: cannot open shared object file: No such file or directory


driftywinds:
This is what it says when I try to open it up in the terminal

avently:
Looks like you:
- don't have openssl installed -> install it via a package manager of your OS (on Ubuntu it's `sudo apt install libssl-dev`)
- don't have specific version of openssl installed (1.1) -> probably, nothing you can do if your OS doesn't have a 1.1 version
- don't have a symlink (`libcrypto.so.1.1` to `libcrypto.so`) -> you can try to create a symlink yourself, like (in Ubuntu x64) `sudo ln -s /usr/lib/x86_64-linux-gnu/libcrypto.so /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1`.
And then try to launch it again.

epoberezkin:
1. use cabal, it won't compile with stack - I'm updating the doc (there is an open PR, but not complete)
2. you need to add cabal.project.local pointing to the location of libcrypto - see CONTRIBUTING.md in docs folder.

snakyjake1:
In the future I hope for Linux AppImage, Flatpak, or similar.

westhighr:
I fixed the problem with the command `sudo dnf install openssl1.1 `on fedora with the package manager dnf.

tomquas:
failed to build on macos-arm64, too.

```
$ docker build .
...
#7 47.67 [ Info  ] Unpacking: ghc-8.10.7-aarch64-deb10-linux.tar.xz to /root/.ghcup/tmp/ghcup-ab8065eeee4453d0
#7 70.45 [ Info  ] Installing GHC (this may take a while)

#7 71.97
#7 71.97
#7 71.97
#7 71.97
#7 71.97
#7 74.76 [ Error ] [GHCup-00841] Process "make" with arguments ["DESTDIR=/root/.ghcup/tmp/ghcup-cb24028a5d988ddd",
#7 74.76 [ ...   ]                                "install"] failed with exit code 2.
#7 74.76 [ Error ] Also check the logs in /root/.ghcup/logs
------
executor failed running [/bin/sh -c ghcup install ghc 8.10.7]: exit code: 3
```

the cli would be really handy for some testing...

JakeTrock:
#1762

ata-star:
Doesnt work on ubuntu or anything based on it

WilhelmFreiheitsberg:
[...]

PR #5773, opened 2025-03-21
Detect linux details to use the correct build during installation of the CLI

ShapeOfMatter:
This is sibling to, and a little more involved than,  #5070. 

Also likely addresses, mitigates, or otherwise affects #4922 and #1274.


ShapeOfMatter:

PR #5799, opened 2025-04-02
core, ui: optionally use short links

epoberezkin:


src/Simplex/Chat/Library/Commands.hs
```diff
@@ -1744,23 +1746,25 @@ processChatCommand' vr = \case
             )
   APIConnect userId incognito (Just (ACR SCMContact cReq)) -> withUserId userId $ \user -> connectViaContact user incognito cReq
   APIConnect _ _ Nothing -> throwChatError CEInvalidConnReq
-  Connect incognito aCReqUri@(Just cReqUri) -> withUser $ \user@User {userId} -> do
-    plan <- connectPlan user cReqUri `catchChatError` const (pure $ CPInvitationLink ILPOk)
+  Connect incognito (Just cLink@(ACL m cLink')) -> withUser $ \user@User {userId} -> do
+    (cReq, plan) <- connectPlan user cLink `catchChatError` \e -> case cLink' of CLFull cReq -> pure (ACR m cReq, CPInvitationLink ILPOk); _ -> throwError e
```
spaced4ndy, line 1750: why throw error if short?

apps/ios/Shared/Views/UserSettings/PrivacySettings.swift
```diff
@@ -20,6 +20,8 @@ struct PrivacySettings: View {
     @AppStorage(GROUP_DEFAULT_PRIVACY_ENCRYPT_LOCAL_FILES, store: groupDefaults) private var encryptLocalFiles = true
     @AppStorage(GROUP_DEFAULT_PRIVACY_ASK_TO_APPROVE_RELAYS, store: groupDefaults) private var askToApproveRelays = true
     @State private var simplexLinkMode = privacySimplexLinkModeDefault.get()
+    @AppStorage(DEFAULT_DEVELOPER_TOOLS) private var developerTools = false
+    @AppStorage(DEFAULT_PRIVACY_SHORT_LINKS) private var shortSimpleLinks = false
```
spaced4ndy, line 24:
```suggestion
    @AppStorage(DEFAULT_PRIVACY_SHORT_LINKS) private var shortSimplexLinks = false
```

apps/ios/Shared/Views/NewChat/NewChatView.swift
```diff
@@ -322,6 +325,8 @@ private struct InviteView: View {
                 .listRowBackground(Color.clear)
                 .listRowSeparator(.hidden)
                 .listRowInsets(EdgeInsets(top: 0, leading: 0, bottom: 0, trailing: 0))
+        } header: {
+            ToggleShortLinkHeader(text: Text("Or show this code"), link: connLinkInvitation, short: $showShortLink)
```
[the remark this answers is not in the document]
spaced4ndy: also seems like bad UX that there's two versions of UI, one via header another via button (groups, pending connections)

PR #5800, opened 2025-04-02
core: fix types for PostgreSQL database

epoberezkin:


tests/ChatClient.hs
```diff
@@ -117,7 +117,9 @@ testCoreOpts =
         { dbConnstr = testDBConnstr,
           -- dbSchemaPrefix is not used in tests (except bot tests where it's redefined),
           -- instead different schema prefix is passed per client so that single test database is used
-          dbSchemaPrefix = ""
+          dbSchemaPrefix = "",
+          dbPoolSize = 3,
```
spaced4ndy, line 121: why make it different from one used in chat?

src/Simplex/Chat/Options/Postgres.hs
```diff
@@ -33,16 +36,32 @@ chatDbOptsP _appDir defaultDbName = do
...
+      ( long "pool-size"
+          <> metavar "DB_POOL_SIZE"
+          <> help "Database connection pool size"
+          <> value 1
+          <> showDefault
+      )
+  dbCreateSchema <-
+    switch
+      ( long "create-schema"
+          <> help "Create database schema when it does not exist"
```
spaced4ndy, line 51: should be default true, or is it so already?

epoberezkin: why? I think it should be false by default

tests/ChatClient.hs
epoberezkin: different how? In chat it's configured via option. Here we test that pool works.

PR #5825, opened 2025-04-10
core: improve connection plan for short links

epoberezkin:


src/Simplex/Chat/Library/Commands.hs
```diff
@@ -3045,37 +3039,76 @@ processChatCommand' vr = \case
...
+      CLFull cReq -> contactReqAndPlan cReq Nothing
+      CLShort sLnk@(CSLContact _ ct _ _) -> case ct of
+        CCTContact ->
+          withFastStore' (\db -> getUserContactLinkViaShortLink db user sLnk) >>= \case
+            Just (UserContactLink (CCLink cReq _) _) -> pure (ACCL SCMContact $ CCLink cReq (Just sLnk), CPContactAddress CAPOwnLink)
+            Nothing -> getShortLinkConnReq user sLnk >>= (`contactReqAndPlan` Just sLnk)
+        CCTGroup ->
+          withFastStore' (\db -> getGroupInfoViaUserShortLink db vr user sLnk) >>= \case
+            Just (cReq, g) -> pure (ACCL SCMContact $ CCLink cReq (Just sLnk), CPGroupLink (GLPOwnLink g))
+            Nothing -> getShortLinkConnReq user sLnk >>= (`contactReqAndPlan` Just sLnk)
```
spaced4ndy, line 3064: could be smell that it first differentiates contact/group here, then both do contactReqAndPlan, which in turn again differentiates contact/group but differently.

shouldn't it be streamlined to all contact related separately, all group related separately?

contactReqAndPlan is used elsewhere too, so maybe it should be split into "contactReqAndPlanAddr", "contactReqAndPlanGroup", these two reused here and united contactReqAndPlan used where it's needed?

PR #5830, opened 2025-04-14
kotlin: refactor chat contexts 2 (null secondary context, pass context instead of content tag, straighten chat state code)

spaced4ndy:


apps/multiplatform/common/src/androidMain/kotlin/chat/simplex/common/platform/UI.android.kt
```diff
@@ -80,9 +80,6 @@ actual class GlobalExceptionsHandler: Thread.UncaughtExceptionHandler {
           chatModel.chatId.value = null
           chatModel.chatsContext.chatItems.clearAndNotify()
         }
-        withContext(Dispatchers.Main) {
```
spaced4ndy, line 83: this was already present in original code for unclear reason, it seems to simply duplicate the line above

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
```diff
@@ -395,6 +394,26 @@ object ChatModel {
         addChat(chat)
       }
     }
+
+    fun addToChatItems(index: Int, elem: ChatItem) {
```
spaced4ndy, line 398: these are moved inside Context class to avoid round trip in code figuring out current context

PR #5852, opened 2025-04-25
core: correctly communicate group acceptance state

spaced4ndy:


src/Simplex/Chat/Types.hs
```diff
@@ -1050,7 +1050,8 @@ instance ToJSON GroupMemberStatus where
 
 acceptanceToStatus :: Maybe GroupMemberAdmission -> GroupAcceptance -> GroupMemberStatus
 acceptanceToStatus memberAdmission groupAcceptance
-  | groupAcceptance == GAPending = GSMemPendingApproval
+  | groupAcceptance == GAPendingApproval = GSMemPendingApproval
+  | groupAcceptance == GAPendingReview = GSMemPendingReview
   | (memberAdmission >>= review) == Just MCAll = GSMemPendingReview
```
spaced4ndy, line 1055: gives option for host to choose to introduce member for review, even if member admission is not configured

PR #5862, opened 2025-04-30
ios: secondary ItemsModel

spaced4ndy:


apps/ios/Shared/Model/ChatModel.swift
```diff
@@ -115,6 +129,34 @@ class ItemsModel: ObservableObject {
...
+    public var groupScopeInfo: GroupChatScopeInfo? {
+        switch secondaryIMFilter {
+        case nil: nil
+        case let .groupChatScopeContext(scopeInfo): scopeInfo
+        case .msgContentTagContext: nil
+        }
+    }
+}
+
+class PreloadState {
```
spaced4ndy, line 150: moved from ChatScrollHelpers, static shared replaced with instance in ItemsModel

apps/ios/Shared/Views/Chat/ChatView.swift
```diff
@@ -653,79 +654,8 @@ struct ChatView: View {
         }
     }
 
-    class FloatingButtonModel: ObservableObject {
```
spaced4ndy, line 656: moved outside

PR #5881, opened 2025-05-09
core: member requires attention counter

spaced4ndy:


src/Simplex/Chat/Store/Groups.hs
```diff
@@ -1224,6 +1230,34 @@ updateGroupMemberAccepted db User {userId} m@GroupMember {groupMemberId} status
     (status, role, currentTs, userId, groupMemberId)
   pure m {memberStatus = status, memberRole = role, updatedAt = currentTs}
 
+updateGroupMembersRequireAttention :: DB.Connection -> User -> GroupInfo -> GroupMember -> GroupMember -> IO GroupInfo
+updateGroupMembersRequireAttention db user@User {userId} g@GroupInfo {groupId, membersRequireAttention} member member'
+  | gmRequiresAttention member' && not (gmRequiresAttention member) =
+      increaseGroupMembersRequireAttention db user g
+  | not (gmRequiresAttention member') && gmRequiresAttention member = do
```
epoberezkin, line 1237: better put them into names

src/Simplex/Chat/Store/Groups.hs
```diff
@@ -1224,6 +1230,34 @@ updateGroupMemberAccepted db User {userId} m@GroupMember {groupMemberId} status
...
+      DB.execute
+        db
+        [sql|
+          UPDATE groups
+          SET members_require_attention = members_require_attention - 1
+          WHERE user_id = ? AND group_id = ?
```
epoberezkin, line 1243: this and below can be one query with parameter 1/-1

src/Simplex/Chat/Store/Groups.hs
epoberezkin: gmRequiresAttention can be made to return 1/0, then it's just 

| was == now -> pure g
| otherwise -> add (now - was)

PR #5910, opened 2025-05-14
ios: open support chat with any member

spaced4ndy:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/group/GroupMemberInfoView.kt
```diff
@@ -441,7 +441,6 @@ fun GroupMemberInfoLayout(
       SectionView {
         if (
           groupInfo.membership.memberRole >= GroupMemberRole.Moderator &&
-          member.memberActive &&
```
spaced4ndy, line 444: same condition is above

PR #5921, opened 2025-05-17
core: support observers send messages to admins, CLI commands

epoberezkin:


src/Simplex/Chat/Library/Commands.hs
```diff
@@ -3369,7 +3372,7 @@ processChatCommand' vr = \case
           Nothing -> Just GRAuthor
           Just (GCSMemberSupport Nothing)
             | memberPending membership -> Nothing
```
epoberezkin, line 3374: why this returns Nothing and not GRObserver?

spaced4ndy: if it works observer should be ok

epoberezkin: I don't know if it works

PR #5947, opened 2025-05-26
core: short link connection plans wip; short links implementation comments

spaced4ndy:


src/Simplex/Chat/Controller.hs
```diff
@@ -448,6 +448,14 @@ data ChatCommand
   | APISetConnectionIncognito Int64 IncognitoEnabled
   | APIChangeConnectionUser Int64 UserId -- new user id to switch connection to
   | APIConnectPlan UserId AConnectionLink
+  | APIPrepareInvitationContact UserId InvitationLinkData ACreatedConnLink
+  | APIPrepareAddressContact UserId AddressLinkData ACreatedConnLink
```
epoberezkin, line 452: looking at the types, it seems like invitation/address link data, and all relevant APIs can be merged into one, and the rest managed via contact status.

Otherwise it creates too much taxonomy that would be largely duplicated.

Welcome message would still remain Maybe, and we would still allow sending only one message prior to connecting... Maybe we do need to have some additional flags though to differentiate.

src/Simplex/Chat/Controller.hs
```diff
@@ -448,6 +448,14 @@ data ChatCommand
...
+  | APIPrepareGroup UserId GroupLinkData ACreatedConnLink
+  | APIChangeContactUser ContactId UserId
+  | APIChangeGroupUser GroupId UserId
+  | APIConnectPreparedInvitationContact {contactId :: ContactId, msgContent_ :: Maybe MsgContent}
+  | APIConnectPreparedAddressContact {contactId :: ContactId, msgContent_ :: Maybe MsgContent}
```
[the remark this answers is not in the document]
epoberezkin: and what they do would depend on the status

src/Simplex/Chat/Store/SQLite/Migrations/M20250526_short_links.hs
```diff
@@ -0,0 +1,22 @@
+{-# LANGUAGE QuasiQuotes #-}
+
+module Simplex.Chat.Store.SQLite.Migrations.M20250526_short_links where
+
+import Database.SQLite.Simple (Query)
+import Database.SQLite.Simple.QQ (sql)
+
+m20250526_short_links :: Query
+m20250526_short_links =
```
epoberezkin, line 9: we already had migration with that name I think... Maybe short_link_profiles

src/Simplex/Chat/Store/SQLite/Migrations/M20250526_short_links.hs
```diff
@@ -0,0 +1,22 @@
...
+  [sql|
+ALTER TABLE contacts ADD COLUMN inv_conn_req_to_connect BLOB;
+ALTER TABLE contacts ADD COLUMN addr_conn_req_to_connect BLOB;
```
epoberezkin, line 12: I don't think this is necessary to have two fields here, we have type that can accomodate both, and you can build logic off subtypes - that is what subtype of link is stored here would determine what happens. Some additional field may still be necessary if we need to differentiate something in UI.

src/Simplex/Chat/Types.hs
```diff
@@ -649,6 +649,23 @@ instance ToField ImageData where toField (ImageData t) = toField t
...
+-- TODO [short links] StrEncoding instances?
+data InvitationLinkData = InvitationLinkData
+  { profile :: Profile
+  }
+  deriving (Show)
+
+data AddressLinkData = AddressLinkData
+  { profile :: Profile,
+    welcomeMessage :: Maybe Text
+  }
```
epoberezkin, line 661: For now we could merge these types. The argument not to merge would be the flag to prohibit/allow connection request messages that would only be needed for addresses. "Allow" would fit with legacy clients that all prohibit connection request messages by default, and we could enable for existing addresses via migration.

src/Simplex/Chat/Store/SQLite/Migrations/M20250526_short_links.hs
spaced4ndy: it's of different date, so name is different

PR #5950, opened 2025-05-27
short links better UX

spaced4ndy:
- [x] set short link data on link creation
- [x] return short link data in connection plans
- [x] using short link data prepare contact, group; connect to prepared entities; send initial message
- [x] create contact request with attached contact, create message
- [x] track "short link data set" state for user address; based on the flag:
  - allow setting data for short link when no data is set _(UI)_
  - prohibit to connect incognito if data is set, ignore incognito on accepting _(core, UI)_
  - don't send auto reply message if peer supports _(core)_
 - [x] allow to choose incognito when connecting to prepared entity _(UI)_
 - [x] allow to change user when conneting to prepared entity _(core, UI)_
 - [x] update short link data on profile updates, address auto-reply, etc. _(core)_
 - [x] business address - declare in short link data, create group when preparing entity _(core, UI) (?)_
 - [x] fix short link data encoding (images don't decode) _(core)_
   - also on updating invitation short link data - AGENT A_MESSAGE error
 - [x] short link data compression _(core)_
 - [x] bug: connecting incognito to prepared group doesn't display that membership is incognito _(core)_
 - [x] change default to short links, remove toggle _(UI)_
 - [x] show beautiful error on profile update when network is off _(UI)_
 - [x] review, test connection plan logic (comments, scanning short link of prepared entity should open it) _(core)_
   - bug: scanning short link of known prepared contact offers to prepare new contact, instead of opening known one
 - [x] create feature and e2ee items in the beginning of conversation, prior to connection - review case by case:
   - prepared entities - contact via invitation, contact via address, group
   - contact via member
   - contact request
 - [x] fix auto-reply when large short link data is disabled (should send auto-reply) _(core)_
 - [x] postgres migration

\+ Kotlin
 
 After beta
 - [ ] progress view when getting connection plan _(UI)_
 - [x] fix address deletion _(core)_
   - either alert user that N chats will be deleted and delete contact request contacts and business chats
   - or allow to accept contact requests for deleted address (remove cascade deletes, rework agent)
 - [ ] ~~negotiate auto-reply message _(core)_~~
 - [ ] ~~asynchronous connect to prepared contact, group _(core) (?)_~~
 - [ ] automatically focus keyboard when opening prepared contact _(UI)_

### DIFF
--- a/apps/ios/Shared/Views/Chat/ComposeMessage/ContextContactRequestActionsView.swift
+++ b/apps/ios/Shared/Views/Chat/ComposeMessage/ContextContactRequestActionsView.swift
@@ -69,7 +69,7 @@ struct ContextContactRequestActionsView: View {
     private func acceptRequest(incognito: Bool = false) {
         inProgress = true
         Task {
-            await acceptContactRequest(incognito: false, contactRequestId: contactRequestId)
+            await acceptContactRequest(incognito: incognito, contactRequestId: contactRequestId)
             await MainActor.run { inProgress = false }
         }
     }
--- a/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
+++ b/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
@@ -171,7 +171,7 @@ object ChatModel {
   // return true if you handled the click
   var centerPanelBackgroundClickHandler: (() -> Boolean)? = null
 
-  var addressShortLinkDataSet: Boolean = userAddress.value?.shortLinkDataSet ?: true
+  fun addressShortLinkDataSet(): Boolean = userAddress.value?.shortLinkDataSet ?: true
 
   fun getUser(userId: Long): User? = if (currentUser.value?.userId == userId) {
     currentUser.value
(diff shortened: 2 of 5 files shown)


apps/ios/Shared/Views/Chat/ComposeMessage/ContextProfilePickerView.swift
```diff
@@ -0,0 +1,284 @@
...
+    private func profilerPickerUserOption(_ user: User) -> some View {
+        Button {
+            if selectedUser == user {
+                if !incognitoDefault {
+                    listExpanded.toggle()
+                } else {
+                    incognitoDefault = false
+                    listExpanded = false
+                }
+            } else if selectedUser != user {
```
spaced4ndy, line 150: seems redundant

apps/ios/Shared/Views/Chat/ComposeMessage/ContextProfilePickerView.swift
```diff
@@ -0,0 +1,284 @@
...
+    }
+    
+    private func profilerPickerUserOption(_ user: User) -> some View {
+        Button {
+            if selectedUser == user {
+                if !incognitoDefault {
+                    listExpanded.toggle()
+                } else {
+                    incognitoDefault = false
+                    listExpanded = false
```
spaced4ndy, line 148: also redundant
```suggestion
                listExpanded.toggle()
```

Issue #4810, opened 2024-09-01
[Solved?]: Simplex crashes with storage DCL disabled


ghost:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Platform

Android

### OS version

AP2A.240805.005/2024083100

### App version

6.0.4

### Current Behavior

App crashes after a while of use, enabling storage DCL fixes it. Disabling memory DCL doesn't seem cause crashes.

### Expected Behavior

App doesn't crash.

### Steps To Reproduce

1. update to latest GrapheneOS 2024083100 release
2. disable storage DCL for simplex
3. open simplex and wait for a minute 
4. app will crash, GrapheneOS will send a notification about attempted storage DCL use

### Relevant log output

```shell
type: storage_DCL
osVersion: google/husky/husky:14/AP2A.240805.005/2024083100:user/release-keys
package: chat.simplex.app:237

DCL denial type: DENY_EXECUTE_APP_DATA_FILE
thread: ghc_worker
targetPath: /data/data/chat.simplex.app/cache/#68973 (deleted)
```

ayaen:
I think, this deserves further explanation and context since we aren't dealing with standard android restrictions. Simplex essentially seems to be making use of executable memory and shared libraries and toggles on GrapheneOS cause it to crash. Simplex itself is only running with standard android environment assumptions but it would be nice if we can get better insight into whats going on.

ayaen:
The app uses a Haskell Runtime since the core is in Haskell, thus the Dynamic code loading observed. Fixing it I think will require the haskell base to be completely precompiled and I am not sure how far that is possible or even optimal

epoberezkin:
We are looking into it, but to me indiscriminately disabling dynamic code loading seems counterproductive, for several reasons:
1) if the app is malicious, preventing dynamic code loading is not going to prevent any app from attacking the users.
2) if the app does not load code from untrusted sources, and simply dynamically loads its own libraries, then avoiding DCL would result in longer start times and, potentially, larger memory usage.

As far as I understand it, correct me if I am wrong, DCL in itself does not allow attacker without elevated privileges any advantages. And if the attacker managed to elevate privileges, then they don't need DCL, they can simply modify the app code before it is started.

[...]

PR #5957, opened 2025-06-01
core: update simplexmq (use static RNG function)

epoberezkin:


epoberezkin:
Possbly related to https://github.com/simplex-chat/simplex-chat/issues/4810

The actual change is here: https://github.com/simplex-chat/simplexmq/pull/1556/files

PR #5978, opened 2025-06-10
core, ios: track "short link data set" state, interact with address accordingly

spaced4ndy:


apps/ios/Shared/Model/NtfManager.swift
```diff
@@ -12,7 +12,6 @@ import UIKit
 import SimpleXChat
 
 let ntfActionAcceptContact = "NTF_ACT_ACCEPT_CONTACT"
-let ntfActionAcceptContactIncognito = "NTF_ACT_ACCEPT_CONTACT_INCOGNITO"
```
spaced4ndy, line 15: removed this notification action, as it seems we can't simply rely on chatModel.userAddress for shortLinkDataSet to decide whether we can show it, because this action can also allow accepting contact request for another user (see below), while chatModel would have this data point for current user

apps/ios/Shared/Model/NtfManager.swift
```diff
@@ -59,13 +58,12 @@ class NtfManager: NSObject, UNUserNotificationCenterDelegate, ObservableObject {
...
         }
-        if content.categoryIdentifier == ntfCategoryContactRequest && (action == ntfActionAcceptContact || action == ntfActionAcceptContactIncognito),
+        if content.categoryIdentifier == ntfCategoryContactRequest && action == ntfActionAcceptContact,
            let chatId = content.userInfo["chatId"] as? String {
-            let incognito = action == ntfActionAcceptContactIncognito
             if case let .contactRequest(contactRequest) = chatModel.getChat(chatId)?.chatInfo {
-                Task { await acceptContactRequest(incognito: incognito, contactRequestId: contactRequest.apiId) }
+                Task { await acceptContactRequest(incognito: false, contactRequestId: contactRequest.apiId) }
             } else {
-                chatModel.ntfContactRequest = NTFContactRequest(incognito: incognito, chatId: chatId)
```
spaced4ndy, line 68: here action is remembered in chatModel for when chats are updated (this user's userAddress.shortLinkDataSet will be irrelevant when user is changed)

PR #5991, opened 2025-06-16
core, ui: create all links with short links, config parameter to use large link data, use short link as address in user profile

epoberezkin:


tests/ChatTests/Profiles.hs
```diff
@@ -48,22 +48,27 @@ chatProfileTests = do
     it "deduplicate contact requests with profile change" testDeduplicateContactRequestsProfileChange
     it "reject contact and delete contact link" testRejectContactAndDeleteUserContact
     it "delete connection requests when contact link deleted" testDeleteConnectionRequests
-    it "auto-reply message" testAutoReplyMessage
-    it "auto-reply message in incognito" testAutoReplyMessageInIncognito
+    -- TODO [short links] order of events is incorrect when short link exists, and auto-reply message is not sent/not shown
```
spaced4ndy: auto-reply should be created for prepared contact via short-link data, so it's valid that it's earlier

tests/ChatTests/Profiles.hs
```diff
@@ -48,22 +48,27 @@ chatProfileTests = do
...
     it "connecting via contact address" testPlanAddressConnecting
     it "connecting via contact address (slow handshake)" testPlanAddressConnectingSlow
     it "re-connect with deleted contact" testPlanAddressContactDeletedReconnected
     it "contact via address" testPlanAddressContactViaAddress
+    it "contact via short address" testPlanAddressContactViaShortAddress
   describe "incognito" $ do
     it "connect incognito via invitation link" testConnectIncognitoInvitationLink
     it "connect incognito via contact address" testConnectIncognitoContactAddress
-    it "accept contact request incognito" testAcceptContactRequestIncognito
+    -- TODO [short links] accepting incognito is not accepted with short links
```
spaced4ndy: it's also valid that incognito is prohibited with short link data set, we either should find a way to create address without it, or remove test

PR #6011, opened 2025-06-25
core: rework contact requests so that they are always created with entity

spaced4ndy:


src/Simplex/Chat/Store/ContactRequest.hs
```diff
@@ -0,0 +1,333 @@
...
+import UnliftIO.STM
+#if defined(dbPostgres)
+import Database.PostgreSQL.Simple ((:.) (..))
+import Database.PostgreSQL.Simple.SqlQQ (sql)
+#else
+import Database.SQLite.Simple ((:.) (..))
+import Database.SQLite.Simple.QQ (sql)
+#endif
+
+createOrUpdateContactRequest ::
```
spaced4ndy, line 47: moved to new module because it required functions from multiple store modules, which either created cyclic dependencies, or required to move multiple other functions to shared module

src/Simplex/Chat/Store/ContactRequest.hs
```diff
@@ -0,0 +1,299 @@
...
+                  "UPDATE contact_requests SET contact_id = ? WHERE contact_request_id = ?"
+                  (contactId, contactRequestId)
+              ucr <- getContactRequest db user contactRequestId
+              ct <- getContact db vr user contactId
+              pure $ RSCurrentRequest ucr (Just $ REContact ct) False
+            createBusinessChat = do
+              let Profile {preferences = userPreferences} = profileToSendOnAccept user Nothing True
+                  groupPreferences = maybe defaultBusinessGroupPrefs businessGroupPrefs userPreferences
+              (gInfo@GroupInfo {groupId}, clientMember) <-
+                createBusinessRequestGroup db vr gVar user cReqChatVRange profile profileId ldn groupPreferences
```
spaced4ndy, line 204: Here we're passing contact request's profile id (to be used for client member, to avoid duplicating profile) and ldn (for group, to have avoid ldn_1)

src/Simplex/Chat/Store/ContactRequest.hs
```diff
@@ -0,0 +1,299 @@
...
+                liftIO $ do
+                  DB.execute
+                    db
+                    [sql|
+                      UPDATE contact_requests
+                      SET agent_invitation_id = ?, pq_support = ?, peer_chat_min_version = ?, peer_chat_max_version = ?, local_display_name = ?, updated_at = ?
+                      WHERE user_id = ? AND contact_request_id = ?
+                    |]
+                    (Binary invId, pqSup, minV, maxV, ldn, currentTs, userId, contactRequestId)
+                  -- Here we could also update business chat, but is always synchronously auto-accepted so it's less of an issue
```
spaced4ndy, line 262: this could be improved but currently ignored, as it's synchronously auto-accepted and profile update would be received via regular channel (unlike message change that wouldn't be sent as an update)

PR #6012, opened 2025-06-25
ios: better error handling

epoberezkin:


apps/ios/SimpleXChat/ChatTypes.swift
```diff
@@ -1378,6 +1378,7 @@ public enum ChatInfo: Identifiable, Decodable, NamedChat, Hashable {
                     case .memGroupDeleted: return ("group is deleted", nil)
                     case .memRemoved: return ("removed from group", nil)
                     case .memLeft: return ("you left", nil)
+                    case .memUnknown: return groupInfo.businessChat == nil ? ("can't send messages", nil) : nil
```
spaced4ndy: this probably could instead be more specific `groupInfo.nextConnectPrepared`

PR #6017, opened 2025-06-27
core: add/send contact address and request messages, ios: use auto-reply as welcome message

epoberezkin:


src/Simplex/Chat/Store/Shared.hs
```diff
@@ -431,7 +431,22 @@ deleteUnusedIncognitoProfileById_ db User {userId} profileId =
...
+    DB.execute
+      db
+      [sql|
+        DELETE FROM contact_profiles
+        WHERE user_id = ? AND contact_profile_id = ?
+          AND NOT EXISTS (SELECT 1 FROM contacts WHERE contact_profile_id = ?)
+          AND NOT EXISTS (SELECT 1 FROM contact_requests WHERE contact_profile_id = ?)
+          AND NOT EXISTS (SELECT 1 FROM contact_requests WHERE contact_profile_id = ? OR member_profile_id = ?)
+      |]
+      (userId, profileId, profileId, profileId, profileId, profileId)
```
spaced4ndy: why 2x
AND NOT EXISTS (SELECT 1 FROM contact_requests WHERE contact_profile_id = ? ...

spaced4ndy: also there're already functions `deleteUnusedIncognitoProfileById_`, `deleteUnusedProfile_` - why can't they be used? `deleteUnusedProfile_` does check contact requests

tests/ChatTests/Utils.hs
```diff
@@ -293,16 +293,21 @@ lastChatFeature = snd $ last chatFeatures
...
 groupFeatureStrs :: [String]
 groupFeatureStrs = map (\(a, _, _) -> snd a) $ groupFeatures'' 0
 
 groupFeatures'' :: Int -> [((Int, String), Maybe (Int, String), Maybe String)]
-groupFeatures'' dir =
-  [ ((dir, e2eeInfoNoPQStr), Nothing, Nothing),
-    ((dir, "Disappearing messages: off"), Nothing, Nothing),
+groupFeatures'' dir = ((dir, e2eeInfoNoPQStr), Nothing, Nothing) : groupFeatures_ dir
+
+groupFeatures_ :: Int -> [((Int, String), Maybe (Int, String), Maybe String)]
```
spaced4ndy, line 308: add comment when to use which one?

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -2903,31 +2929,85 @@ processChatCommand' vr = \case
...
+          cReqHash2 = cReqHash $ CRContactUri crData {crScheme = simplexChat}
       case groupLinkId of
         -- contact address
         Nothing ->
-          withFastStore' (\db -> getConnReqContactXContactId db vr user cReqHash) >>= \case
+          withFastStore' (\db -> getConnReqContactXContactId db vr user cReqHash1 cReqHash2) >>= \case
+            (Just Contact {activeConn = Just conn@Connection {connStatus = ConnPrepared}}, Just (xContactId, _)) -> do
+              incognitoProfile <- joinPreparedConn' xContactId conn
+              pure $ CRSentInvitation user (toPCC conn) incognitoProfile
+              where
```
spaced4ndy:
```suggestion
```

src/Simplex/Chat/Store/Shared.hs
epoberezkin: the second should be group_members

epoberezkin: deleteUnusedIncognitoProfileById_ - this one does not check contact_requests.

both deleteUnusedIncognitoProfileById_ and deleteUnusedProfile_ checks connections, which is what we don't need here.

this function also updates connection.

epoberezkin: This query was not tested and we either need to add tests that change from incognito to profile and vice versa, or to prohibit such changes after the first attempt.

PR #6032, opened 2025-07-02
core: allow to accept contact requests after address is deleted

spaced4ndy:


apps/ios/SimpleX NSE/NotificationService.swift
```diff
@@ -996,7 +996,11 @@ func receivedMsgNtf(_ res: NSEChatEvent) async -> (String, NSENotificationData)?
 //        case let .contactConnecting(contact):
 //            TODO profile update
     case let .receivedContactRequest(user, contactRequest):
-        return (UserContact(contactRequest: contactRequest).id, .contactRequest(user, contactRequest))
+        if let userContactLinkId = contactRequest.userContactLinkId_ {
+            return (UserContact(userContactLinkId: userContactLinkId).id, .contactRequest(user, contactRequest))
+        } else {
+            return nil
+        }
```
spaced4ndy, line 1003: it's ok because there can be no notification without address

epoberezkin: why?

PR #6034, opened 2025-07-03
core: refactor connectViaContact

spaced4ndy:


src/Simplex/Chat/Library/Commands.hs
```diff
@@ -2928,53 +2946,47 @@ processChatCommand' vr = \case
...
+      case preparedEntity_ of
+        Just (PCEContact ct@Contact {activeConn}) -> case activeConn of
+          Nothing -> connect' Nothing cReqHash1 Nothing
+          Just conn@Connection {connStatus, xContactId} -> case connStatus of
+            ConnPrepared -> joinPreparedConn' xContactId conn False
+            _ -> pure $ CVRConnectedContact ct
+        Just (PCEGroup _gInfo GroupMember {activeConn}) -> case activeConn of
+          Nothing -> connect' groupLinkId cReqHash1 Nothing
+          Just conn@Connection {connStatus, xContactId} -> case connStatus of
+            ConnPrepared -> joinPreparedConn' xContactId conn $ isJust groupLinkId
```
epoberezkin, line 2968: This correctly retries group join and business profile connection after error, it was creating another connection before (I think)

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -2928,53 +2946,47 @@ processChatCommand' vr = \case
...
+            _ -> connect' groupLinkId cReqHash1 xContactId -- why not "already connected" for host member?
```
epoberezkin, line 2969: as the comment says, it feels like it should be a similar event to contact already exists.

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -2928,53 +2946,47 @@ processChatCommand' vr = \case
...
-            (Nothing, Just (xContactId, Just conn@Connection {connId, connStatus = ConnPrepared})) -> do
-              incognitoProfile <- joinPreparedConn' xContactId conn
-              -- TODO [short links] align Connection and PendingContactConnection so it doesn't need to be read again
-              pcc <- withStore $ \db -> getPendingContactConnection db userId connId
-              pure $ CRSentInvitation user pcc incognitoProfile
-            (Nothing, xContactId_) -> procCmd $ do
+            Right ct@Contact {activeConn} -> case groupLinkId of
+              Nothing -> case activeConn of
+                Just conn@Connection {connStatus = ConnPrepared, xContactId} -> joinPreparedConn' xContactId conn False
+                _ -> pure $ CVRConnectedContact ct
```
epoberezkin, line 2975: This is probably too broad, but it is guarded by another branch for prepared contacts. But if a standard CLI command is used to connect to the link that has prepared contact for it, it would not work correctly. So we may restore the removed logic that effectively does the same as with prepared contact but without it being passed. In this case connect' should not use passed preparedEntity, but use the one constructed locally here.

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -2928,53 +2946,47 @@ processChatCommand' vr = \case
...
+              Just gLinkId ->
```
epoberezkin, line 2976: this all as simplified because xContactId is now in Connection. It's not possble, for example, that xContact is present and Connection is not. Also flipping cases removed duplicate code, as the logic for connection in the absense of contact is the same with and without groupLinkId.

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -2928,53 +2946,47 @@ processChatCommand' vr = \case
...
+                -- allow repeat contact request
+                -- TODO [short links] is this branch needed? it probably remained from the time we created host contact
+                connect' (Just gLinkId) cReqHash1 Nothing
```
epoberezkin, line 2979: see the comment. This can be improved

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -2928,53 +2946,47 @@ processChatCommand' vr = \case
...
-              connect' (Just gLinkId) cReqHash1 newXContactId True
-            (Nothing, Just (xContactId, Just conn@Connection {connStatus = ConnPrepared})) -> do
-              incognitoProfile <- joinPreparedConn' xContactId conn
-              pure $ CRSentInvitation user (toPCC conn) incognitoProfile
-            (_, xContactId_) -> procCmd $ do
-              -- TODO [short links] this is executed on repeat request after success
-              -- it probably should send the second message without creating the second connection?
-              xContactId <- mkXContactId $ fst <$> xContactId_
-              connect' (Just gLinkId) cReqHash1 xContactId True
+              Just Connection {xContactId} -> connect' groupLinkId cReqHash1 xContactId
```
epoberezkin, line 2984: also see comment

src/Simplex/Chat/Library/Commands.hs
spaced4ndy: yes, I think it's an old oversight

src/Simplex/Chat/Library/Commands.hs
spaced4ndy: > But if a standard CLI command is used to connect to the link that has prepared contact for it

I think we don't even have CLI API to prepare contacts and groups, so it's unlikely for real users

spaced4ndy: I'm not sure I see a fundamental difference that would change something here though: checks for presense of contact with active connection in Prepared status. what am I missing?

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -2992,33 +3004,21 @@ processChatCommand' vr = \case
...
-              localAlias = "",
-              createdAt,
-              updatedAt = createdAt
-            }
         mkXContactId = maybe (XContactId <$> drgRandomBytes 16) pure
-        connect' groupLinkId cReqHash xContactId inGroup = do
-          let pqSup = if inGroup then PQSupportOff else PQSupportOn
+        connect' groupLinkId cReqHash xContactId_ = do
+          let inGroup = isJust groupLinkId
+              pqSup = if inGroup then PQSupportOff else PQSupportOn
```
spaced4ndy, line 3012: wouldn't this enable pqSup in business chat?

src/Simplex/Chat/Store/Direct.hs
```diff
@@ -174,21 +169,51 @@ createConnReqConnection db userId acId cReqHash sLnk attachConnTo_ xContactId in
...
+  pure
+    Connection
+      { connId,
+        agentConnId = AgentConnId acId,
+        connChatVersion = chatV,
+        -- TODO (proposed):
+        -- - add agent version 8 for short links
+        -- - update agentToChatVersion to convert 8 to 16
+        -- - return and correctly set peer's range from link (via connRequestPQSupport)
+        peerChatVRange = chatInitialVRange, -- this is 1-1
```
spaced4ndy, line 190: basically to set peer chat version range earlier based on short link data, right?

PR #6035, opened 2025-07-03
android: fix crash on opening group link

spaced4ndy:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/group/GroupLinkView.kt
```diff
@@ -33,7 +33,7 @@ fun GroupLinkView(
   creatingGroup: Boolean = false,
   close: (() -> Unit)? = null
 ) {
-  var groupLinkVar by rememberSaveable { mutableStateOf(groupLink) }
+  var groupLinkVar by rememberSaveable(stateSaver = GroupLink.nullableStateSaver) { mutableStateOf(groupLink) }
```
spaced4ndy, line 36: this was lost here when I updated to GroupLink type:
https://github.com/simplex-chat/simplex-chat/pull/5950/files#diff-46d19d39ef05125ad175ff60449f3047bbd1c772685f4595dc58370ff8580648L37

PR #6043, opened 2025-07-06
core: use longer timeouts for background network requests, support retries with increased timeouts

epoberezkin:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/SimpleXAPI.kt
```diff
@@ -666,7 +673,87 @@ object ChatController {
...
+      title = generalGetString(alert.first),
+      text = alert.second,
+      confirmText = generalGetString(MR.strings.retry_verb),
+      onConfirm = onRetry,
+      onDismiss = onCancel,
+      onDismissRequest = onCancel
+    )
+  }
+
+  private fun retryableNetworkErrorAlert(err: ChatError): Pair<StringResource, String>? {
```
spaced4ndy, line 716: it seems more narrow than `networkErrorAlert`, is that why they are both kept?

spaced4ndy (approved):

PR #6055, opened 2025-07-09
core, ui: short bio/description for contact and group profiles

epoberezkin:


src/Simplex/Chat/View.hs
```diff
@@ -2672,8 +2679,8 @@ groupScopeInfoStr = \case
 ttyFilePath :: FilePath -> StyledString
 ttyFilePath = plain
 
-optFullName :: ContactName -> Text -> StyledString
-optFullName localDisplayName fullName = plain $ optionalFullName localDisplayName fullName
+optFullName :: ContactName -> Text -> Maybe Text -> StyledString
+optFullName localDisplayName fullName shortDescr = plain $ optionalFullName localDisplayName fullName shortDescr
```
spaced4ndy, line 2683: I don't think it's good idea to show up to 160 characters as part of name in cli

PR #6058, opened 2025-07-10
fix: fix AndroidManifest.xml so that the dynamic icon is actually applied

### LINKED ISSUE #1301: Material You app icon
It would be nice if a Material You app icon version could be made and added as the app icon doesn't blend well for users with themed icons enabled, such as myself. The official documentation for how to do so can be found [here](https://developer.android.com/about/versions/13/features#themed-app-icons)

An-anonymous-coder:
it seems the code is already implemented [here](https://github.com/simplex-chat/simplex-chat/blob/stable/apps/multiplatform/android/src/main/res/mipmap-anydpi-v26/icon.xml), but it isn't used [here](https://github.com/simplex-chat/simplex-chat/blob/stable/apps/multiplatform/android/src/main/AndroidManifest.xml). Line 42 should be `android:icon="@mipmap/ic_launcher"` instead of `android:icon="@mipmap/icon"` now the MDY icon should be applied properly. This closes #1301

### DIFF
--- a/apps/multiplatform/android/src/main/AndroidManifest.xml
+++ b/apps/multiplatform/android/src/main/AndroidManifest.xml
@@ -39,7 +39,7 @@
       android:name="SimplexApp"
       android:allowBackup="false"
       android:fullBackupOnly="false"
-      android:icon="@mipmap/icon"
+      android:icon="@mipmap/ic_launcher"
       android:label="${app_name}"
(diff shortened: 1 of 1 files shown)


An-anonymous-coder:

An-anonymous-coder:
Credit to https://github.com/simplex-chat/simplex-chat/pull/5687

PR #6176, opened 2025-08-11
android: reduce battery usage by not using permanent wakelock

epoberezkin:


apps/multiplatform/common/src/androidMain/kotlin/chat/simplex/common/views/usersettings/NotificationsSettingsView.android.kt
```diff
@@ -0,0 +1,26 @@
...
+import chat.simplex.common.platform.AppPlatform
+import chat.simplex.common.platform.appPlatform
+import chat.simplex.common.platform.chatModel
+import chat.simplex.res.MR
+import dev.icerock.moko.resources.compose.painterResource
+import dev.icerock.moko.resources.compose.stringResource
+
+@Composable
+actual fun ServiceSaveBatterySection(notificationsMode: State<NotificationsMode>) {
+  if (appPlatform == AppPlatform.ANDROID && notificationsMode.value == NotificationsMode.SERVICE && chatModel.controller.appPrefs.developerTools.get()) {
```
spaced4ndy: why is it needed to check `appPlatform == AppPlatform.ANDROID` if it's android module? also everywhere in code it's `appPlatform.isAndroid`

epoberezkin: removed option anyway

PR #6178, opened 2025-08-12
core: rework synchronous group forwarding in receive loop into asynchronous delivery tasks

spaced4ndy:


docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,143 @@
...
+- In channels members will not connect directly to other members and channel owners.
+- Chat relays will not take into consideration status of introductions.
+- Chat relays will forward messages between owners and all members (from members to owners, for example, for reactions, comments).
+
+### All chat relays forward all messages
+
+- Instead of inviting admins all chat relays will forward all messages.
+- Members will deduplicate messages (already implemented).
+- Members should highlight differences in deduplicated messages. This would solve trust issue.
+  - Question: If messages are to be signed by owners, what differences can chat relays introduce? Does this point in initial doc imply owner sending different versions of message to chat relays? Losses in delivery?
```
epoberezkin: not all messages will be signed

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,143 @@
...
+
+- For batch of received messages a forwarding instruction has to be persisted.
+- Special worker would pick up forwarding instructions for sending.
+- Forwarding instruction can be split into smaller batches by members to make smaller group sends one at a time.
+
+How to present a forwarding instruction in storage?
+
+Currently we build ad-hoc forwarding instruction for each message based on its scope (GroupForwardScope). Then for each scope message batches are sent separately to different member lists.
+
+Question: should we persist all forwarding instructions of or only for main (GFSMain - for regular messages) scope? For main and all (GFSAll), and don't persist for support scope (GFSMemberSupport)?
```
epoberezkin: Listing pros/cons for either option would help.

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,143 @@
...
+
+We roughly need following data to be persisted for a forwarding instruction:
+- group id,
+- forwarding scope - TBC as above,
+- sending member - to include memberId in XGrpMsgForward,
+- broker timestamp of received messages batch to include in XGrpMsgForward,
+- list of messages (events).
```
epoberezkin, line 96: feels like "message from channel" should be part of this scope.

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,143 @@
...
+  ```
+  msg_body = [{"v":"1-16","msgId":"ZUd3bzVDTXFHWmc3Z29YSQ==","event":"x.msg.del","params":{"msgId":"QVVuN2RkaXg0aVJJdTZXSA=="}},{"v":"1-16","msgId":"amM5dzV2RkRjNmNYSDRLQQ==","event":"x.msg.del","params":{"msgId":"NzFlV3pNRStpQTRrQjRYUg=="}}]
+  ```
+  On the other hand for forwarding operation we require ChatMessage 'Json (for XGrpMsgForward).
+- For a forwarding instruction we can persist list of messages ids from messages table (message_id), and save ChatMessage 'Json on records to restore.
+- Alternatively, we can read full msg_body (it can even be duplicated on forwarding instruction), and restore chat messages (ChatMessage 'Json) via parseChatMessages. However, this conflicts with possibility that a single batch body will have messages of different scopes, which current processing logic allows, although there's no legitimate case as of now. So, persisting ChatMessage' Json on each record seems preferable. Also it avoids repeated parsing.
+- As a side note, it seems we can stop saving msg_body for received messages, as it's not used for any purpose. This field is only used for retrieving and sending pending group messages.
```
epoberezkin: I don't quite follow this argument, it's confusing.

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,143 @@
...
+
+Schema draft:
+
+```sql
+ALTER TABLE messages ADD COLUMN chat_message_json TEXT;
+
+CREATE TABLE forwarding_instructions (
```
epoberezkin: name is wrong

epoberezkin: and the design should also account for "feed" command, and also include business chats into feed, even if implementation is done separately.

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,143 @@
...
+Forwarding worker will read forwarding instructions sorted by created_at, retrieve chat messages corresponding to message_ids, retrieve group members based on group_id and forward_scope and send batched forward events to them.
+
+How to split forwarding to large number of group members into smaller batches?
+
+One option is to determine forwarding lists at the moment of creating forwarding instruction and persist them, possibly already split into predefined batches. However, this unnecessarily excludes newly joined members (those that joined between creating instruction and worker picking it up for forwarding operation).
+
+Another approach would be to split retrieved members into batches at time of processing instruction, and persist group member ids that instruction was processed for after sending one batch at a time. Group member retrieval should then filter out already processed members in case of operation failure leading to repeat processing.
+
+```sql
+CREATE TABLE forwarding_instructions_members (
```
epoberezkin: this seems insufficient to track forwarding status per member.

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,143 @@
...
+  forwarding_instruction_member_id INTEGER PRIMARY KEY,
+  forwarding_instruction_id INTEGER NOT NULL REFERENCES forwarding_instructions ON DELETE CASCADE,
+  group_member_id INTEGER NOT NULL REFERENCES group_members ON DELETE CASCADE
+);
+```
+
+Overall forwarding worker algorithm would then be:
+1. Retrieve next forwarding instruction.
```
epoberezkin: It is not "instruction". The conventional name is "job".

epoberezkin: also, sequential execution of forwarding jobs would result in very large delivery delays - I don't see why they can't be concurrent.

epoberezkin: I think even batches within one job can be concurrent.

epoberezkin: for which it probably is worth pre-allocating members to batch IDs (although probably no need to materialize batches). And it's actually unclear why we need to materialize jobs and not just have members if message is referenced anyway via ID? (so message ID determines job). Order of execution of jobs should match message ordering though, and it's the order of reception.

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,143 @@
...
+    - Retrieve chat messages corresponding to message_ids, build forward events (XGrpMsgForward).
+    - Retrieve member list based on group_id and forward_scope, filtering out already processed members via forwarding_instructions_members.
```
epoberezkin: that will be inefficient. Pre-allocating to batches may help it

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,143 @@
...
+2. Split member list into smaller batches. TBC how to choose optimal batch size.
```
epoberezkin: cannot be ad hoc, won't scale

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,143 @@
...
+3. For each batch of members:
+    1. Group send (sendGroupMessages_) batch of XGrpMsgForward events.
+    2. Persist processed member ids to forwarding_instructions_members.
+        - Persist after save is ok because receiving client can deduplicate in case chat relay fails in-between.
+    3. Possibly small delay to avoid overloading database.
```
epoberezkin: there is no need for it, we have two separate pools already, these jobs would use low priority pool

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,143 @@
...
+        - Question: Since chat relays will be working on postgres with multiple connections, do we really need this splitting into member batches? Group send is already batched in itself.
```
epoberezkin: yes, it won't work for a million members

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,143 @@
...
+4. Once forwarding instruction is fully processed, delete it.
+
+## Other considerations
+
+### Hiding sending owner id
+
+Initial doc mentions "messages sent from channel name" in minimal testable scope. This would be a feature allowing owners to send messages to channel without it being clear for members which owner sent it. For this, we'd also like to hide sending owner's member id from forwarding operation.
+
+This means MemberId in XGrpMsgForward should become optional.
```
epoberezkin, line 173: yes, consistent with having it in this scope.

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,143 @@
...
+
+Receiving members would then save this message as received from group. We already have necessary machinery for chat items persistence - see ShowGroupAsSender. However, message processing logic would have to be reworked to allow for absence of group member. Also, sending owner would have to indicate it to chat relays in XMsgNew (TBC - other events?).
+
+```haskell
+XMsgNew :: ShowGroupAsSender -> MsgContainer -> ChatMsgEvent 'Json
+```
+
+Question: Should this be resolved in the same scope, or separately?
+
+Question: Should chat relays also hide sending owner from other owners or not? Probably not, then forwarding instruction would have to be split into separate sends with different sets of XGrpMsgForward events (with and without sending owner's member id).
```
epoberezkin: interesting point. Yes, owners should probably know by whom it's sent.

epoberezkin: yes, one more argument to solve both problems together

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Connection deleting events
+
+Forwarding regular messages, reactions, updates and many other events is straightforward. However, naive processing of some events currently breaks forwarding logic, specifically member and group deletion.
+
+As it is now, forwarding operation follows event processing. So, processing of XGrpMemDel, XGrpDel first deletes member connection (or connections with all members in case of group deletion), then attempts to forward these events, which in reality never works as at this point connection is already deleted.
+
+Currently we ignore this problem (there are TODOs), as all members according to protocol connect with each other, and forwarding serves only as message delivery improvement/backup, and not a main route, and so members in at least supposedly most cases receive these events from the original sender.
+
+With chat relays, however, no messages are sent directly from owners to members, so if logic is kept as is these events would never be received. So, their processing should be special cased to delay until after forwarding operation.
```
epoberezkin, line 197: yes, also part of this scope.

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,143 @@
...
+
+Question: Should this be resolved in the same scope, or separately? This is already an issue in current group design, so it could be a separate improvement.
```
epoberezkin: yes

docs/rfcs/2025-08-11-channels-forwarding.md
spaced4ndy: It discusses mechanism on how to save and restore messages for forwarding. Currently we save full batch body on each received message record (batch can create many received message records with the same batch body), restoring from which would require repeated parsing and does not account for possibility of messages with different scopes in a single batch.

docs/rfcs/2025-08-11-channels-forwarding.md
spaced4ndy: I think it may be better done separately into master to reduce scope of this change

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,188 @@
...
+Currently we simply ignore duplicate messages - see createNewRcvMessage.
+
+Some options on how to highlight difference in messages from chat relays:
+1. Show only the fact that there is difference.
+    - Persist duplicate error (SEDuplicateGroupMessage) as flag on chat item to then display warning sign in UI.
+2. Save message hashes as some additional entity linked to message or chat item for each chat relay.
+    - If any new hash differs, set flag on chat item for warning sign in UI.
+    - Add ability to load additional info (with other chat item info via APIGetChatItemInfo) showing which relays sent different hashes (e.g. 1 differs from 2 others - show 1 as warning, all differ - show all as warning).
+3. As previous, but save content. Possibly save content additionally to hashes, for faster comparison.
+    - Saving content means converting duplicate messages to chat item content, or only saving text, but latter will fail to show difference in files.
```
epoberezkin, line 45: why can't we have them as usual chat items that would show next to duplicates? Or make a special "different version of another chat item" as a special reply (to utilize reply machinery)?

As they would have the same timestamp (would they?), they would show next to original. If timestamp is different (probably), we could make it the same. It won't work for replies though.

In this case we could simply have a flag "different message from X relay" so users can see which relay?

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,188 @@
...
+
+Service events will not have content. We still can/should compare them and indicate difference, e.g. by creating a special chat item (similar to integrity violation).
+
+How to compute hashes. Problem is with file descriptions, as it is legitimate case that they will be different. Hash could be computed as hash of message without file invitation + file digest from file invitation (FileInvitation.fileDigest).
```
epoberezkin: it should use FileInvitation, but without description. Digest and names should be the same, the problem of different names needs to be fixed.

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,188 @@
...
+
+Files can also cause problem with showing difference in content if file preview is the same, but file hash is different. So showing difference in content does not exclude necessity of showing difference in hashes as in option 2.
+
+As a note, this whole section discusses an edge case of chat relays maliciously changing messages, and countermeasures to prevent them doing so undetectably. May be not worth implementing overly complex solution as MVP (option 3), and option 2 or even 1 can suffice. TBC further.
```
epoberezkin: correct, for MVP we could just create event item, as long as it would show straight after the first version it's good enough. Also put a flag on the first.

epoberezkin: And I guess in this case we need events from all relays so they can decide which is likely to misbehave, and track which relays delivered main item.

epoberezkin: possibly even simpler - a marker on item, that would show in info which relays delivered it with a sign which relay delivered different hash.

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,188 @@
...
+- group id,
+- forwarding scope,
+- sending member - to include memberId in XGrpMsgForward,
+- broker timestamp of received messages batch to include in XGrpMsgForward,
+- "message from channel" flag from sending owner - see "Hiding sending owner id" section below,
+- list of messages (events).
+
+**How to build message batches for a forwarding job:**
+
+Currently for each message a full MsgBody is saved on `messages` table record, which is just a ByteString. Also if received messages were batched, full batch body is saved for each message record (it's a questionable design). For example, here is message body of batched deletion of 2 chat items, which is saved on 2 message records:
```
epoberezkin, line 100: it should be painless to change

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,188 @@
...
+
+```
+msg_body = [{"v":"1-16","msgId":"ZUd3bzVDTXFHWmc3Z29YSQ==","event":"x.msg.del","params":{"msgId":"QVVuN2RkaXg0aVJJdTZXSA=="}},{"v":"1-16","msgId":"amM5dzV2RkRjNmNYSDRLQQ==","event":"x.msg.del","params":{"msgId":"NzFlV3pNRStpQTRrQjRYUg=="}}]
+```
+
+On the other hand for forwarding operation we require ChatMessage 'Json (for XGrpMsgForward), which could be saved on message records instead.
+
+As a side note, it seems we can stop saving full batch body as msg_body for received messages, as it's not used for any purpose. This field is only used for retrieving and sending pending group messages (sent, not received).
```
epoberezkin, line 108: yes

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,188 @@
...
+
+It's unclear if there's a need for a separate lower level abstraction for delivery jobs, that could be reused for feeds and other purposes, as they would require different logic of building delivery lists, and scheduling a delivery for agent already serves a similar purpose. For now, it seems this additional abstraction is unnecessary.
+
+**How a forwarding job will work overall:**
+
+0. There is some process on start that launches necessary workers for each group/scope client serves as a forwarding agent. Also message receive loop "kicks" necessary workers. Below we start with a worker for some group/scope trying to retrieve next work item, that is being next message(s) to forward.
```
epoberezkin: only for those that have messages that need forwarding

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,188 @@
...
+1. Worker retrieves next message from `messages` table that was marked for forwarding, that matches the worker's group/scope.
+    - Question: Do we need a separate flag on message record to mark it as "for forwarding"? Perhaps having forward_scope (see schema below) is enough.
```
epoberezkin: we need "forward pending" state to know what to forward I think

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,188 @@
...
+2. Worker checks, if this message is already attached to some forwarding job, and the state of the job.
+    1. Worker gets job record.
+        - If message is not attached to any job yet, worker creates a new job record, sets job id on the message record (normal execution path).
+        - Otherwise worker gets existing job attached to the message (recovery path).
+    2. Worker determines forwarding batch encoding for the job.
+        - If encoding is not saved on the job, it means it wasn't determined before (normal execution):
+          1. Worker retrieves more messages that match its group/scope (in loop or in bulk? also message list has to preserve reception order).
```
epoberezkin: in bulk is better I think

epoberezkin: but we won't know how much we'd need. Some heuristic could be read 100, then mark those that are used with jobId, or load more if needed. It's important to put encoding into job and mark all relevant messages in one transaction.

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,188 @@
...
+          2. Worker prepares encoding (wraps messages in XGrpMsgForward events with metadata) and saves it on the job record.
+        - Otherwise worker uses saved encoding (recovery).
+3. With moving cursor on group member id for member retrieval, forward encoded batch. In loop:
+    1. Retrieve members up to some limit and filtering according to a cursor. In normal execution path on first iteration job would not have a previously saved cursor yet.
+        - For Main scope all current members will be retrieved;
+        - For Support scopes - moderators and above + scope member;
```
epoberezkin: we discussed we don't need a "job" for support scope, we just forward to all.

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,188 @@
...
+        - For All scope - all current and pending members (Question: will channels have member review? If not, possibly we don't have to account for All scope).
```
epoberezkin: no need for member review in channels.

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,188 @@
...
+        - Main difference from forwarding in regular groups here is that for inviting admins we retrieve only introduced and invited members for the message's sending member, that are not yet connected. For channels there will be no such filtering.
+    2. Group send encoded batch to retrieved members.
+    3. Update cursor on the job.
+4. Marks all messages attached to the job as forwarded (forward_complete), deletes the job record.
```
epoberezkin, line 146: let's not delete at once, better do it in cleanup manager after say a week.

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,188 @@
...
+This means MemberId in XGrpMsgForward should become optional.
+
+```haskell
+XGrpMsgForward :: Maybe MemberId -> ChatMessage 'Json -> UTCTime -> ChatMsgEvent 'Json
+```
+
+Receiving members would then save this message as received from group. We already have necessary machinery for chat items persistence - see ShowGroupAsSender. However, message processing logic would have to be reworked to allow for absence of group member. Also, sending owner would have to indicate it to chat relays in XMsgNew (TBC - other events?).
+
+```haskell
+XMsgNew :: ShowGroupAsSender -> MsgContainer -> ChatMsgEvent 'Json
```
epoberezkin, line 182: can be in MsgContainer I think? why it has to be separate?

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,188 @@
...
+```
+
+Question: Should this be resolved in the same scope, or separately?
```
epoberezkin: resolved what?

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -0,0 +1,188 @@
...
+
+Question: Should chat relays also hide sending owner from other owners or not? Probably not, then forwarding job would have to be split into separate sends with different sets of XGrpMsgForward events (with and without sending owner's member id).
```
epoberezkin: no, should be visible to owners, but shown as from channel.

docs/rfcs/2025-08-11-channels-forwarding.md
epoberezkin: ok

docs/rfcs/2025-08-11-channels-forwarding.md
```diff
@@ -43,14 +43,20 @@ Some options on how to highlight difference in messages from chat relays:
     - Add ability to load additional info (with other chat item info via APIGetChatItemInfo) showing which relays sent different hashes (e.g. 1 differs from 2 others - show 1 as warning, all differ - show all as warning).
 3. As previous, but save content. Possibly save content additionally to hashes, for faster comparison.
     - Saving content means converting duplicate messages to chat item content, or only saving text, but latter will fail to show difference in files.
+4. Create new versions as "different version of another message" chat items.
+    - Timestamp would have to be made the same as on original chat item, as it's not guaranteed to be the same - each chat relay will forward a broker ts based on its own metadata of receiving message from sender.
+    - It could be done via special replies, to re-use reply machinery. However, this won't work for messages being replies themselves then.
+    - Alternatively, it could say "different message from X relay".
 
 Service events will not have content. We still can/should compare them and indicate difference, e.g. by creating a special chat item (similar to integrity violation).
```
epoberezkin, line 51: I don't think they need to. We'd just compare the hash with whoever is the first, without trying to decide which is right.

src/Simplex/Chat/Messages.hs
```diff
@@ -1178,14 +1178,51 @@ data RcvMessage = RcvMessage
 
 type MessageId = Int64
 
--- TODO [channels fwd] review
-data ForwardingJob = ForwardingJob
-  { jobId :: Int64,
-    msgBatchEncoding :: Maybe (NonEmpty (ChatMessage 'Json)),
-    cursorGroupMemberId :: Maybe Int64
+-- TODO [channels fwd] review types
+data ForwardTask = ForwardTask
```
epoberezkin:
```suggestion
data DeliveryTask = DeliveryTask
```

src/Simplex/Chat/Messages.hs
```diff
@@ -1178,14 +1178,51 @@ data RcvMessage = RcvMessage
...
+  { taskId :: Int64,
+    taskContext :: ForwardTaskContext
   }
   deriving (Show)
 
+data ForwardTaskContext
```
epoberezkin: it's not exactly context, it's the task itself. if ID is the only field that is common, there is no point having a wrapping type at all - ID can be used as selector if it's present in all constructors I think.

epoberezkin:
```suggestion
data DeliveryTask
```

src/Simplex/Chat/Messages.hs
```diff
@@ -1178,14 +1178,51 @@ data RcvMessage = RcvMessage
...
+  = FTCMessage {messageForwardContext :: MessageForwardContext}
+  | FTCRelayRemoval {relayRemovalForwardContext :: RelayRemovalForwardContext}
+  | FTCReactionCount {reactionCountForwradContext :: ReactionCountForwradContext}
+  deriving (Show)
+
+-- prevSenderInteractionTs: members are split based on this timestamp;
+-- postInteractionCursorGMId: for these members sender profiles are packaged;
+-- question: how to retrieve sender profiles for multiple messages when we already created encoding/list of them?
+-- perhaps we should save references to messages on the task, instead of encoding;
+data MessageForwardContext = MessageForwardContext
```
epoberezkin:
```suggestion
data MessageForwardTask = MessageForwardTask
```

I don't understand why it is a "context"? Context usually means something around the thing, not the thing itself.

src/Simplex/Chat/Messages.hs
```diff
@@ -1178,14 +1178,51 @@ data RcvMessage = RcvMessage
...
+  { prevSenderInteractionTs :: Maybe UTCTime,
+    postInteractionCursorGMId :: Maybe Int64,
+    preInteractionCursorGMId :: Maybe Int64,
```
epoberezkin: I don't understand why it needs two? And what is "interaction"?

epoberezkin: Is it about which members need profile and which not? The name is wrong then.

src/Simplex/Chat/Messages.hs
```diff
@@ -1178,14 +1178,51 @@ data RcvMessage = RcvMessage
...
+    messages :: Maybe (NonEmpty (ChatMessage 'Json)),
+  }
+  deriving (Show)
+
+-- messageId: we don't know is it group deletion or relay removal;
+-- we could read ChatMessage 'Json by reference instead;
+data RelayRemovalForwardContext = RelayRemovalForwardContext
```
epoberezkin:
```suggestion
data RelayRemovedTask = RelayRemovedTask
```

src/Simplex/Chat/Messages.hs
```diff
@@ -1178,14 +1178,51 @@ data RcvMessage = RcvMessage
...
+  { cursorGMId :: Maybe Int64,
+    messageId :: MessageId
+  }
+  deriving (Show)
+
+-- reaction counts to be retrieved from chat items;
+-- question: how to keep track which chat items received reactions since last such task?
+-- requires additional protocol message
+data ReactionCountForwradContext = ReactionCountForwradContext
```
epoberezkin:
```suggestion
data ChatItemCountsTask = ChatItemCountsTask
```

it would cover comments count, so shouldn't be specialized to reactions.

Also I don't understand why you need separate types and not just fields in constructors?

There are three layers of types when only one seems necessary.

src/Simplex/Chat/Messages.hs
```diff
@@ -1178,14 +1178,51 @@ data RcvMessage = RcvMessage
...
+  { cursorGMId :: Maybe Int64
+  }
+  deriving (Show)
+
+-- to save on task record in db
+data ForwardContextTag = FCTMessage | FCTRelayRemoval | FCTReactionCount
```
epoberezkin: why is it needed?

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -3303,3 +3305,48 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
                   else pure Nothing
               else pure Nothing
           _ -> pure Nothing
+
+-- TODO [channels fwd] forwarding worker implementation, below is not working code
+-- TODO [channels fwd] refactor agent Worker for export and reuse
+
+startForwardWorkers :: AM ()
```
epoberezkin:
```suggestion
startDeliveryWorkers :: AM ()
```

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -3303,3 +3305,48 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+startForwardWorkers = do
+  -- TODO [channels fwd] getPendingFwdAssignments - select based on pending tasks
+  pendingFwdAssignments <- withStore' c $ \db -> getPendingFwdAssignments db
+  lift . forM_ pendingFwdAssignments resumeForwardWork
+
+data ForwardPriority = FPRegular | FPImportant
```
epoberezkin:
```suggestion
data DeliveryPriority = FPRegular | FPImportant
```

What's the use case for different delivery priorities?

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -3303,3 +3305,48 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+
+-- for each group and forwad scope 2 workers can be created for each priority
+type ForwardWorkerAssignment = (GroupId, GroupForwardScope, ForwardPriority)
```
epoberezkin:
```suggestion
type DeliveryWorkerGroupScope = (GroupId, GroupForwardScope, DeliveryPriority)
```

src/Simplex/Chat/Messages.hs
spaced4ndy: yes

bots/api/TYPES.md
```diff
@@ -2365,6 +2367,15 @@ Known:
 - lastMsgFromMemberTs: UTCTime?
 
 
+---
+
+## GroupType
+
+**Enum type**:
+- "smallGroup"
```
epoberezkin: why "small"?

src/Simplex/Chat/Delivery.hs
```diff
@@ -0,0 +1,245 @@
...
+import Simplex.Messaging.Agent.Store.DB (fromTextField_)
+import Simplex.Messaging.Encoding.String
+
+type DeliveryWorkerKey = (GroupId, GroupDeliveryScopeType)
+
+data GroupDeliveryScope
+  = GDSGroup {includePending :: Bool}
+  | GDSMemberSupport {supportGMId :: GroupMemberId}
+  -- | GDSMemberProfile -- planned for sending member profiles -- TODO comment
+  deriving (Eq, Ord, Show)
```
epoberezkin:
```suggestion
data GroupDeliveryTask
  = GDSGroup {includePending :: Bool, deliveryTask :: GroupDeliveryTask}
  | GDSMemberSupport {supportGMId :: GroupMemberId, msgForwardTask :: MessageForwardTask}
  -- | GDSMemberProfile -- planned for sending member profiles -- TODO comment
  deriving (Eq, Ord, Show)
```

src/Simplex/Chat/Delivery.hs
```diff
@@ -0,0 +1,245 @@
...
+
+data GroupDeliveryScopeType
```
epoberezkin:
```suggestion
data GroupDeliveryScope
```

src/Simplex/Chat/Delivery.hs
```diff
@@ -0,0 +1,191 @@
...
+
+data DeliveryJobScope
+  = DJSGroup {jobSpec :: DeliveryJobSpec}
+  | DJSMemberSupport {supportGMId :: GroupMemberId}
+  -- | DJSMemberProfileUpdate
+  deriving (Show)
+
+data DeliveryJobSpec
+  = DJSpecDeliveryJob {includePending :: Bool}
+  | DJSpecRelayRemoved
```
epoberezkin:
```suggestion
  = DJSDeliveryJob {includePending :: Bool}
  | DJSRelayRemoved
```

PR #6183, opened 2025-08-13
rfc: brainstorm on group links and adding relays to groups

epoberezkin:


docs/rfcs/2025-08-12-chat-relay-group-links.md
```diff
@@ -0,0 +1,46 @@
...
+- connection with relay is a "contact", and it has presence in the UI as a conversation. It allows to implement both commands and UI actions, so more flexible, but the downside is that this connection will be owned by one owner, and cannot be shared, so if owner disappears it would be hard to transfer management to other owners.
+- connection with relay is a special "service group" for public group, and each new owner is added to this group. This is probably the best option, as it also would give owners the space to discuss any matters about the group and also send commands to the relay. But the downside here is that we don't support any e2e encryption in this group, and it won't be obvious to the users, so we should probably disable messaging until we have e2e encryption in support and this new "owners" scopes for new channels. But we still can support commands and other UI actions with relay (e.g., suspend accepting new members or removing the relay), and all owners will see these changes in this channel.
```
spaced4ndy: owners will have as many service groups as there are relays for group, so it's questionable place for non relay related discussions even with e2ee protecting from relay. maybe it should be and stay purely technical thing, even if available in UI

spaced4ndy: nvm, that's next paragraph

docs/rfcs/2025-08-12-chat-relay-group-links.md
```diff
@@ -0,0 +1,46 @@
...
+
+The protocol for adding relay to a public group channel:
+- owner creates a "relay service group" for this relay and this public group/channel/.
+- owner sends a special request to relay's contact address to join this service group (it's "invitation" in SMP protocol terms, with group profile in metadata)
+- relay would join the group
+- relay can analyze the invitation and reject if it doesn't like the group profile (it should still join the group, as currently we don't have rejection mechanism that would work without connecting, but we can reconsider it),
+- if relay is happy to service this group, it would generate it's dedicated address for this group and send back in a special message.
+- owner would add this address in delegate section in profile data - it's outside of group profile.
+- owner may notify relay that group profile is updated.
+- relay would validate that it's address is present in profile and update the state.
```
spaced4ndy: does owner add other owners at this point? or in the beginning, before adding relay? what happens if multiple owners add same relay to the same group simultaneously? how can relay deduplicate?

epoberezkin: that's the problem to solve in consensus protocol with proposal/agreement phases. But it may be disrupted. But yes, chat relay would see it's the same link.

Issue #1081, opened 2022-09-21
[Feature Request


EricJ2190:

EricJ2190:
UnifiedPush allows the user to choose a "distributor" to handle the push delivery, and the security would depend on that choice. I would expect that SimpleX would put minimal information in the actual push message and just use it as a cue to fetch data from the message queue server, similar to how push is handled on iOS. A malicious distributor could prevent you from getting immediately notified about new messages but could not view, manipulate, or censor messages.

epoberezkin:
Our code for receiving messages follows exactly the same approach as ntfy, so you would only have benefits with ntfy if you have multiple apps using the same channel.

We need to systematically eliminate any sources of unnecessary traffic (e.g. too many retires, in cases where the failure would be more appropriate) - it is likely to have more positive impact on battery usage than switching to ntfy.

ThatOneCalculator:
I have 5 apps using NTFY on my phone, so UnifiedPush would be very much appreciated

mdebus:
[https://f-droid.org/en/2022/12/18/unifiedpush.html](Here) is a good explanation of UnifiedPush:

@epoberezkin Would It be even be possible to combine your approach with UnifiedPush?

mrg358:
Big plus for ntfy as it can be easily selfhosted.
FWIW popular matrix clients (fluffy chat and element) are automatically utilizing `ntfy` if it is detected running on a users device.

ltGuillaume:
@mrg358 Conversations can be used by other apps for UnifiedPush as well.

khimaros:
> Our code for receiving messages follows exactly the same approach as ntfy, so you would only have benefits with ntfy if you have multiple apps using the same channel.
> 
> We need to systematically eliminate any sources of unnecessary traffic (e.g. too many retires, in cases where the failure would be more appropriate) - it is likely to have more positive impact on battery usage than switching to ntfy.

I also have six apps on my ntfy right now. with simplex that would become seven. instead I'm using periodic notifications to reduce battery drain.

iroedius:
I use NextPush (which is a UnifiedPush distributor) and I would also appreciate it if it was supported.

KaiserCalm:
I've recently discovered ntfy and on a degoogled phone it's really awesome. Would be a great feature!

Darin755:
I would really like to see this be implemented. I use unifiedpush with several other apps using ntfy and its very efficient and allows for a single app to handle everything

[...]

PR #6205, opened 2025-08-19
feat: Add Push notifications to Android (UnifiedPush/Play services)

p1gp1g:
This pull request adds push notifications to Android (Mode=INSTANT). It works with the Play Services or UnifiedPush. (https://github.com/simplex-chat/simplex-chat/issues/1081 which seems to be the most upvoted issue)

The user can use the Play Services or UnifiedPush to get push notifications. When a push message arrives, the MessagesFetcherWorker is started.

CLI commands are added to allow the user to define their own ntf server. And at this moment, the implementation requires the user to self-host the ntf server; this can easily be change, and the db should be updated to add to SimpleX NTF servers their own VAPID key fingerprints (which is requires by the implementation).

I've improved MessagesFetcherWorker in the same time, cf https://github.com/simplex-chat/simplex-chat/commit/4683d47259153c795780e5b5cbfdb213660b1e12 and its message for the details.

Because we start the worker on push message, when and if a new PING NtfMode will be introduced, it will be possible to use it, with a minimal modification here

This PR doesn't include any proprietary library (which is often the case when we support FCM)

**UX**

To activate, the user go to the Settings > Notifications > Notification Service and Select "Instant"
If the user doesn't have any ntf server, a dialog will ask them to add one,
else, either their only push service available will be selected, or a dialog will ask them which one to use

Then, the user will be able to get new notifications for SimpleX even if the app is killed

**Security**

I expect a question about security, at least from some readers (*please use the issue to discuss it instead of the PR*), most of the points are answered here: https://unifiedpush.org/news/20250513_push_security_privacy/

For the current implementation, **if the user self-host their own ntf server** (which doesn't have to be exposed on the Internet) **and their own push service**, as soon as their servers are trusted, **it doesn't add any additional security/privacy risk**: from SimpleX network, there is just a constant connection from their home, like another device would do.

Else, a compromised push service may be used to link a simplex profile with the push service profile (by sending a message on SimpleX, and see who gets a push message). Of course, it requires the user server to be compromised, which may require being able to compromise many push servers. cf. the blog post.

Nevertheless, push notifications allows having an instant messaging experience, without heavily relying on the battery. Even a few % a day may be a no go when using an application for rare occasions (this is an important point to accept installing an application when you only have a few contacts)

### DIFF
--- a/apps/multiplatform/android/build.gradle.kts
+++ b/apps/multiplatform/android/build.gradle.kts
@@ -144,6 +144,11 @@ dependencies {
     //Camera Permission
     implementation("com.google.accompanist:accompanist-permissions:0.34.0")
 
+    // Push notifications
+    implementation("org.unifiedpush.android:connector:3.0.10")
+    // Allow using Play Services if available
+    implementation("org.unifiedpush.android:embedded-fcm-distributor:3.0.0")
+
     //implementation("androidx.compose.material:material-icons-extended:$compose_version")
     //implementation("androidx.compose.ui:ui-util:$compose_version")
 
--- a/apps/multiplatform/android/src/main/AndroidManifest.xml
+++ b/apps/multiplatform/android/src/main/AndroidManifest.xml
@@ -185,6 +185,13 @@
         android:foregroundServiceType="mediaPlayback|microphone|camera|remoteMessaging"
     />
 
+    <service android:name=".PushService"
+      android:exported="false">
+      <intent-filter>
+        <action android:name="org.unifiedpush.android.connector.PUSH_EVENT"/>
+      </intent-filter>
+    </service>
+
     <receiver
         android:name=".CallService$CallActionReceiver"
         android:enabled="true"
--- a/apps/multiplatform/android/src/main/java/chat/simplex/app/MessagesFetcherWorker.kt
+++ b/apps/multiplatform/android/src/main/java/chat/simplex/app/MessagesFetcherWorker.kt
@@ -15,15 +15,30 @@ import java.util.concurrent.TimeUnit
 object MessagesFetcherWorker {
   private const val UNIQUE_WORK_TAG = BuildConfig.APPLICATION_ID + ".UNIQUE_MESSAGES_FETCHER"
 
-  fun scheduleWork(intervalSec: Int = 600, durationSec: Int = 60) {
-    val initialDelaySec = intervalSec.toLong()
-    Log.d(TAG, "Worker: scheduling work to run at ${Date(System.currentTimeMillis() + initialDelaySec * 1000)} for $durationSec sec")
+  /**
+   * Schedule [MessagesFetcherWork]
+   *
+   * @param initialDelayMs Delay in milliseconds before starting the jobs, usually = [intervalMs]
+   * @param intervalMs Interval in milliseconds between the end of a jobs and the start of a new one.
+   *     If set to `0`, the job isn't restarted
+   * @param timeoutMs Timeout in milliseconds until we stop a job, if there are still new messages, a new job is started directly
+   * @param durationMs Duration in milliseconds to wait for new messages
+   */
+  fun scheduleWork(
+    initialDelayMs: Int = MessagesFetcherWork.DEFAULT_INTERVAL_MS,
+    intervalMs: Int = MessagesFetcherWork.DEFAULT_INTERVAL_MS,
+    timeoutMs: Int = MessagesFetcherWork.DEFAULT_TIMEOUT_MS,
+    durationMs: Int = MessagesFetcherWork.DEFAULT_DURATION_MS
+  ) {
+    Int.MAX_VALUE
+    Log.d(TAG, "Worker: scheduling work to run at ${Date(System.currentTimeMillis() + initialDelayMs)} for $timeoutMs ms max")
     val periodicWorkRequest = OneTimeWorkRequest.Builder(MessagesFetcherWork::class.java)
-      .setInitialDelay(initialDelaySec, TimeUnit.SECONDS)
+      .setInitialDelay(initialDelayMs.toLong(), TimeUnit.MILLISECONDS)
       .setInputData(
         Data.Builder()
-          .putInt(MessagesFetcherWork.INPUT_DATA_INTERVAL, intervalSec)
-          .putInt(MessagesFetcherWork.INPUT_DATA_DURATION, durationSec)
+          .putInt(MessagesFetcherWork.INPUT_DATA_INTERVAL, intervalMs)
+          .putInt(MessagesFetcherWork.INPUT_DATA_TIMEOUT, timeoutMs)
+          .putInt(MessagesFetcherWork.INPUT_DATA_DURATION, durationMs)
           .build()
       )
       .setConstraints(Constraints.Builder().setRequiredNetworkType(NetworkType.CONNECTED).build())
@@ -45,25 +60,49 @@ class MessagesFetcherWork(
   workerParams: WorkerParameters
 ): CoroutineWorker(context, workerParams) {
   companion object {
+    /**
+     * Interval, in milliseconds, between the end of a job and the start of a new one
+     */
     const val INPUT_DATA_INTERVAL = "interval"
+
+    /**
+     * Timeout, in milliseconds, until we force stop the jobs, if there are new messages, a new job is started directly
+     */
+    const val INPUT_DATA_TIMEOUT = "timeout"
+
+    /**
+     * Minimum duration, in milliseconds, to wait for new messages
+     */
     const val INPUT_DATA_DURATION = "duration"
-    private const val WAIT_AFTER_LAST_MESSAGE: Long = 10_000
+    const val DEFAULT_DURATION_MS = 10_000
+    const val DEFAULT_TIMEOUT_MS = 60_000
+    const val DEFAULT_INTERVAL_MS = 600_000
   }
 
   override suspend fun doWork(): Result {
     // Skip when Simplex service is currently working
+    val durationMs = inputData.getInt(INPUT_DATA_DURATION, DEFAULT_DURATION_MS)
+    val timeoutMs = inputData.getInt(INPUT_DATA_TIMEOUT, DEFAULT_TIMEOUT_MS)
+    val intervalMs = inputData.getInt(INPUT_DATA_INTERVAL, DEFAULT_INTERVAL_MS)
+    // initialDelayMs may be = 0
+    // in this case, reschedule won't start a worker.
+    var initialDelayMs = intervalMs
     if (SimplexService.getServiceState(SimplexApp.context) == SimplexService.ServiceState.STARTED) {
-      reschedule()
+      reschedule(
+        initialDelayMs = initialDelayMs,
+        intervalMs = intervalMs,
+        timeoutMs = timeoutMs,
+        durationMs = durationMs
+      )
       return Result.success()
     }
-    val durationSeconds = inputData.getInt(INPUT_DATA_DURATION, 60)
     var shouldReschedule = true
     try {
       // In case of self-destruct is enabled the initialization process will not start in SimplexApp, Let's start it here
       if (DatabaseUtils.ksSelfDestructPassword.get() != null && chatModel.chatDbStatus.value == null) {
         initChatControllerOnStart()
       }
-      withTimeout(durationSeconds * 1000L) {
+      withTimeout(timeoutMs.toLong()) {
         val chatController = ChatController
         SimplexService.waitDbMigrationEnds(chatController)
         val chatDbStatus = chatController.chatModel.chatDbStatus.value
@@ -75,26 +114,48 @@ class MessagesFetcherWork(
         }
         Log.w(TAG, "Worker: starting work")
         // Give some time to start receiving messages
-        delay(10_000)
+        delay(durationMs + 200L)
         while (!isStopped) {
-          if (chatController.lastMsgReceivedTimestamp + WAIT_AFTER_LAST_MESSAGE < System.currentTimeMillis()) {
+          val lastMsgMs = System.currentTimeMillis() - chatController.lastMsgReceivedTimestamp
+          if (lastMsgMs > durationMs) {
             Log.d(TAG, "Worker: work is done")
             break
           }
-          delay(5000)
+          delay(durationMs - lastMsgMs + 200L)
         }
       }
     } catch (_: TimeoutCancellationException) { // When timeout happens
-      Log.d(TAG, "Worker: work is done (took $durationSeconds sec)")
+      Log.d(TAG, "Worker: Still work to do, restarting a new work (took $durationMs sec)")
+      // We reschedule a new work now (in 200 ms)
+      shouldReschedule = true
+      initialDelayMs = 200
     } catch (_: CancellationException) { // When user opens the app while the worker is still working
       Log.d(TAG, "Worker: interrupted")
     } catch (e: Exception) {
       Log.d(TAG, "Worker: unexpected exception: ${e.stackTraceToString()}")
     }
 
-    if (shouldReschedule) reschedule()
+    if (shouldReschedule) {
+      reschedule(
+        initialDelayMs = initialDelayMs,
+        intervalMs = intervalMs,
+        timeoutMs = timeoutMs,
+        durationMs = durationMs
+      )
+    }
     return Result.success()
   }
 
-  private fun reschedule() = MessagesFetcherWorker.scheduleWork()
+  /**
+   * Reschedule a work if [initialDelayMs] > 0
+   */
+  private fun reschedule(
+    initialDelayMs: Int,
+    intervalMs: Int,
+    timeoutMs: Int,
+    durationMs: Int
+  ) = {
+    if (initialDelayMs > 0)
+      MessagesFetcherWorker.scheduleWork(initialDelayMs, intervalMs, timeoutMs, durationMs)
+  }
 }
--- a/apps/multiplatform/android/src/main/java/chat/simplex/app/PushService.kt
+++ b/apps/multiplatform/android/src/main/java/chat/simplex/app/PushService.kt
@@ -1,0 +1,82 @@
+package chat.simplex.app
+
+import chat.simplex.common.model.CC
+import chat.simplex.common.platform.Log
+import chat.simplex.common.platform.chatModel
+import kotlinx.coroutines.*
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+import org.unifiedpush.android.connector.FailedReason
+import org.unifiedpush.android.connector.PushService
+import org.unifiedpush.android.connector.data.PushEndpoint
+import org.unifiedpush.android.connector.data.PushMessage
+
+class PushService: PushService() {
+  private val json = Json {
+    ignoreUnknownKeys = true
+  }
+
+  @Serializable
+  data class PNMessage(
+    val verification: String? = null
+  )
+
+  private fun onVerification(code: String) {
+    CoroutineScope(Dispatchers.Default).launch {
+      chatModel.controller.sendCmd(
+        null,
+        CC.APIVerifySavedNtf(code),
+        log = true
+      )
+    }
+  }
+
+  override fun onMessage(message: PushMessage, instance: String) {
+    Log.d(TAG, "onMessage")
+    val pn: PNMessage = json.decodeFromString(String(message.content))
+    when {
+      pn.verification != null -> onVerification(pn.verification)
+    }
+    // TODO: Start same job than the periodic service ?
+    // Receiving the push notif is enough to wake the app and fetch msgs
+    // But it may not be enough when the phone is in doze, or with some
+    // vendors
+  }
+
+  override fun onNewEndpoint(endpoint: PushEndpoint, instance: String) {
+    Log.d(TAG, "onNewEndpoint")
+    endpoint.pubKeySet ?: run {
+      // Should not happen
+      Log.w(TAG, "Missing pubKeySet")
+      return
+    }
+    CoroutineScope(Dispatchers.Default).launch {
+      chatModel.controller.sendCmd(
+        null,
+        CC.APIRegisterWebPush(endpoint.url, endpoint.pubKeySet!!.auth, endpoint.pubKeySet!!.pubKey),
+        log = true
+      )
+    }
+  }
+
+  override fun onRegistrationFailed(reason: FailedReason, instance: String) {
+    Log.d(TAG, "onRegistrationFailed: $reason")
+    // TODO: notification to inform about failed registration
+  }
+
+  override fun onUnregistered(instance: String) {
+    Log.d(TAG, "onUnregistered")
+    // TODO: notification to inform about unregistration
+    CoroutineScope(Dispatchers.Default).launch {
+      chatModel.controller.sendCmd(
+        null,
+        CC.APIDeleteSavedNtf(),
+        log = true
+      )
+    }
+  }
+
+  companion object {
+    private const val TAG = "PushService"
+  }
+}
--- a/apps/multiplatform/android/src/main/java/chat/simplex/app/PushService.kt
+++ b/apps/multiplatform/android/src/main/java/chat/simplex/app/PushService.kt
@@ -37,10 +37,7 @@ class PushService: PushService() {
     when {
       pn.verification != null -> onVerification(pn.verification)
     }
-    // TODO: Start same job than the periodic service ?
-    // Receiving the push notif is enough to wake the app and fetch msgs
-    // But it may not be enough when the phone is in doze, or with some
-    // vendors
+    MessagesFetcherWorker.scheduleWork(200, 0)
   }
 
   override fun onNewEndpoint(endpoint: PushEndpoint, instance: String) {
--- a/apps/multiplatform/android/src/main/java/chat/simplex/app/PushService.kt
+++ b/apps/multiplatform/android/src/main/java/chat/simplex/app/PushService.kt
@@ -3,6 +3,9 @@ package chat.simplex.app
 import chat.simplex.common.model.CC
 import chat.simplex.common.platform.Log
 import chat.simplex.common.platform.chatModel
+import chat.simplex.common.platform.ntfManager
+import chat.simplex.common.views.helpers.generalGetString
+import chat.simplex.res.MR
 import kotlinx.coroutines.*
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.Json
@@ -58,12 +61,22 @@ class PushService: PushService() {
 
   override fun onRegistrationFailed(reason: FailedReason, instance: String) {
     Log.d(TAG, "onRegistrationFailed: $reason")
-    // TODO: notification to inform about failed registration
+    val title = generalGetString(MR.strings.icon_descr_instant_notifications)
+    val text = when (reason) {
+      FailedReason.NETWORK -> generalGetString(MR.strings.unifiedpush_registration_failed_network)
+      FailedReason.VAPID_REQUIRED, // Should not happen, VAPID will be required
+      FailedReason.INTERNAL_ERROR -> generalGetString(MR.strings.unifiedpush_registration_failed_unknown)
+      FailedReason.ACTION_REQUIRED -> generalGetString(MR.strings.unifiedpush_registration_failed_action)
+
+    }
+    ntfManager.showMessage(title, text)
   }
 
   override fun onUnregistered(instance: String) {
     Log.d(TAG, "onUnregistered")
-    // TODO: notification to inform about unregistration
+    val title = generalGetString(MR.strings.icon_descr_instant_notifications)
+    val text = generalGetString(MR.strings.unifiedpush_unregistered)
+    ntfManager.showMessage(title, text)
     CoroutineScope(Dispatchers.Default).launch {
       chatModel.controller.sendCmd(
         null,
--- a/apps/multiplatform/android/src/main/java/chat/simplex/app/SimplexApp.kt
+++ b/apps/multiplatform/android/src/main/java/chat/simplex/app/SimplexApp.kt
@@ -20,6 +20,7 @@ import androidx.work.*
 import chat.simplex.app.MainActivity.Companion.OLD_ANDROID_UI_FLAGS
 import chat.simplex.app.model.NtfManager
 import chat.simplex.app.model.NtfManager.AcceptCallAction
+import chat.simplex.app.platform.PushManager
 import chat.simplex.app.views.call.CallActivity
 import chat.simplex.common.helpers.*
 import chat.simplex.common.model.*
@@ -269,6 +270,7 @@ class SimplexApp: Application(), LifecycleEventObserver {
         // Prevents from showing "Enable notifications" alert when onboarding wasn't complete yet
         if (chatModel.controller.appPrefs.onboardingStage.get() == OnboardingStage.OnboardingComplete) {
           SimplexService.showBackgroundServiceNoticeIfNeeded()
+          PushManager.initStart(context)
           if (appPrefs.notificationsMode.get() == NotificationsMode.SERVICE)
             withBGApi {
               platform.androidServiceStart()
--- a/apps/multiplatform/android/src/main/java/chat/simplex/app/SimplexApp.kt
+++ b/apps/multiplatform/android/src/main/java/chat/simplex/app/SimplexApp.kt
@@ -270,11 +270,15 @@ class SimplexApp: Application(), LifecycleEventObserver {
         // Prevents from showing "Enable notifications" alert when onboarding wasn't complete yet
         if (chatModel.controller.appPrefs.onboardingStage.get() == OnboardingStage.OnboardingComplete) {
           SimplexService.showBackgroundServiceNoticeIfNeeded()
-          PushManager.initStart(context)
-          if (appPrefs.notificationsMode.get() == NotificationsMode.SERVICE)
-            withBGApi {
+          when (appPrefs.notificationsMode.get()) {
+            NotificationsMode.SERVICE -> withBGApi {
               platform.androidServiceStart()
             }
+            NotificationsMode.INSTANT -> withBGApi {
+              PushManager.initStart(context)
+            }
+            else -> {}
+          }
         }
       }
 
--- a/apps/multiplatform/android/src/main/java/chat/simplex/app/SimplexApp.kt
+++ b/apps/multiplatform/android/src/main/java/chat/simplex/app/SimplexApp.kt
@@ -236,6 +236,16 @@ class SimplexApp: Application(), LifecycleEventObserver {
         if (mode != NotificationsMode.PERIODIC) {
           MessagesFetcherWorker.cancelAll()
         }
+        if (mode == NotificationsMode.INSTANT) {
+          CoroutineScope(Dispatchers.Default).launch {
+            SimplexService.initUnifiedPush(this) {
+              // Change notifications mode only if everything is correctly setup
+              chatModel.controller.appPrefs.notificationsMode.set(mode)
+            }
+          }
+        } else {
+          chatModel.controller.appPrefs.notificationsMode.set(mode)
+        }
         SimplexService.showBackgroundServiceNoticeIfNeeded(showOffAlert = false)
       }
 
@@ -245,6 +255,7 @@ class SimplexApp: Application(), LifecycleEventObserver {
           NotificationsMode.SERVICE -> CoroutineScope(Dispatchers.Default).launch { platform.androidServiceStart() }
           NotificationsMode.PERIODIC -> SimplexApp.context.schedulePeriodicWakeUp()
           NotificationsMode.OFF -> {}
+          NotificationsMode.INSTANT -> {}
         }
       }
 
@@ -369,6 +380,7 @@ class SimplexApp: Application(), LifecycleEventObserver {
       override fun androidCreateActiveCallState(): Closeable = ActiveCallState()
 
       override val androidApiLevel: Int get() = Build.VERSION.SDK_INT
+      override val supportsPushNotifications = true
     }
   }
 
--- a/apps/multiplatform/android/src/main/java/chat/simplex/app/SimplexService.kt
+++ b/apps/multiplatform/android/src/main/java/chat/simplex/app/SimplexService.kt
@@ -22,6 +22,7 @@ import androidx.core.app.ServiceCompat
 import androidx.core.content.ContextCompat
 import androidx.work.*
 import chat.simplex.app.model.NtfManager
+import chat.simplex.app.platform.PushManager
 import chat.simplex.common.AppLock
 import chat.simplex.common.helpers.requiresIgnoringBattery
 import chat.simplex.common.model.ChatController
@@ -418,6 +419,10 @@ class SimplexService: Service() {
     }
 
     private fun getPreferences(context: Context): SharedPreferences = context.getSharedPreferences(SHARED_PREFS_ID, Context.MODE_PRIVATE)
+
+    suspend fun initUnifiedPush(scope: CoroutineScope, onSuccess: () -> Unit) {
+      PushManager.initUnifiedPush(androidAppContext, scope, onSuccess)
+    }
 
     fun showBackgroundServiceNoticeIfNeeded(showOffAlert: Boolean = true) {
       val appPrefs = ChatController.appPrefs
--- a/apps/multiplatform/android/src/main/java/chat/simplex/app/model/NtfManager.android.kt
+++ b/apps/multiplatform/android/src/main/java/chat/simplex/app/model/NtfManager.android.kt
@@ -231,6 +231,9 @@ object NtfManager {
     val builder = NotificationCompat.Builder(context, MessageChannel)
       .setContentTitle(title)
       .setContentText(text)
+      .setStyle(
+        NotificationCompat.BigTextStyle().bigText(text)
+      )
       .setPriority(NotificationCompat.PRIORITY_HIGH)
       .setGroup(MessageGroup)
       .setGroupAlertBehavior(NotificationCompat.GROUP_ALERT_CHILDREN)
--- a/apps/multiplatform/android/src/main/java/chat/simplex/app/platform/PushManager.kt
+++ b/apps/multiplatform/android/src/main/java/chat/simplex/app/platform/PushManager.kt
@@ -24,12 +24,14 @@ import dev.icerock.moko.resources.compose.painterResource
 import dev.icerock.moko.resources.compose.stringResource
 import kotlinx.coroutines.*
 import org.unifiedpush.android.connector.UnifiedPush
+import androidx.core.net.toUri
 
 /**
  * Object with functions to interact with push services
  */
 object PushManager {
   private const val TAG = "PushManager"
+  private val VAPID_REGEX = Regex("^[A-Za-z0-9_-]{87}(=+)?$")
 
   /**
    * Check if a NTF server is setup and ask user's push distributor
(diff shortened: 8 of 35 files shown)


p1gp1g:

balintbarna:
Thanks so much for your work. Do I understand it right that this will only work with ntfy push server and app or also with other unified push providers, eg. I currently use the Sunup app which uses Mozilla's web push server.

p1gp1g:
No, ntfy and ntf are 2 different things. Ntf is the SimpleX component responsible to send push notifications, only used with iOS APNS at this moment

It works with Sunup

epoberezkin:
That's a major effort, thank you.

I just released the beta version that should substantially reduce battery usage in instant mode, let's re-assess once it becomes stable.

epoberezkin:
Sent you "let's connect" in another repo :)

What I am missing here to connect the dots is an RFC doc we always write for this scale changes (and even smaller). PR comment partially replaces it, but it's not becoming part of git (e.g., if we migrate), and it has several gaps for me (I don't know enough about Unified Push) and several UX questions. So I think we need to make this RFC doc anyway.

Things I don't yet understand/not sure about:
1. Is supporting user's ntf server required for this change? Possibly, it is needed because it would have some credentials from push server and would have user token? As is, there are some hard to review parts/regression risks, especially in simplexmq, so maybe we can split to several PRs and reduce diff noise there.
2. Is there a scenario when user can have zero configuration push notifications, as this is what most people would prefer? E.g., if we decide to provide an option with similar security level as iOS, can we do it? Probably yes.
3. Am I right from reading this PR and [this doc](https://unifiedpush.org/news/20250131_push_for_decentralized/#webpush-to-google-services) that user can use google services without having any google services code in the app / without FCM libraries? This is indeed unknown to most people I think. And from what I read, technically it is possible to do without any closed-source code. This raises several questions:
- can a fully open-source app determine if it has google services to know that it can/should use FCM? Or any device, even without google services can use FCM by requesting a token? This I wasn't able to connect yet.
- to what extent this approach is consistent with play store policy? If the possibility of using web pushes is at least mentioned anywhere in google docs, then it may be consistent with the policy, but if it's completely undisclosed (I could not find any reference to it), then using it in play store app may be counter to play store policy. While this would not fall under "unauthorized use", because VAPID is used, it may fall under using undocumented APIs, and possibly under circumventing the APIs ("get rid of push gateways"), and also there may be additional data processed by the app (tokens etc.), that would need to be disclosed. So it needs to be assessed.
- will we need to have different builds - one for play store, another for github/fdroid etc. distribution? Possibly, it's a parameterized build. I didn't yet figure out how to embed the information into the code if it was built as APK (not play store) or bundle (for play store), but there is a parameter in gradle.properties that toggles a not-too-important thing. If any run-time difference is needed, relying on this parameter would be a bad idea, it's easy to forget to change.
4. Adding a 4th option for notifications is one too many, but if the latest change for service notifications results in similar (or lower) battery usage as periodic fetch, we can just drop periodic completely.
5. UX: How will it be offered during the onboarding? If push is a good thing, then enabling it for the new users is more valuable that an option hidden somewhere that only advanced users would find and benefit from. And it's related to the question whether a self-hosted ntf server should be part of this change. My intuition is that it shouldn't be, and stage one would be releasing a zero-configuration support for push notifications (as battery/security trade off) that new users can enable from onboarding and existing users can enable from what's new (and it should happen *after* the release that would recommend switching from periodic to service mode).

There are some smaller questions (e.g., I think there are simpler options to wrap custom servers into existing APIs) but, let's talk about big ones first.

What I like about this change is that it's rather surgical and seems to wire push support into the existing flows, unless I missed something. My initial reaction was that we should postpone this until after rewiring SMP/NTF notifications protocol from the current pull to push (see the braindump in [this rfc](https://github.com/simplex-chat/simplexmq/blob/stable/rfcs/2025-03-30-ios-notifications-3.md) - the current situation is better than at the time of writing RFC, as we moved to service subscriptions for ntf server, but other problems remain, and switching to push is still justified), but from looking at the code it seems like it may be completely unrelated to the proposed change.

Let's talk.

Thank you!

p1gp1g:
Thanks for the answer, I'll answer point per point

> Sent you "let's connect" in another repo :)
> 
> What I am missing here to connect the dots is an RFC doc we always write for this scale changes (and even smaller). PR comment partially replaces it, but it's not becoming part of git (e.g., if we migrate), and it has several gaps for me (I don't know enough about Unified Push) and several UX questions. So I think we need to make this RFC doc anyway.

Do you think we can write it together? The best would probably be if I'm able to give you all the details necessary to write it, then I review it.

> 1. Is supporting user's ntf server required for this change? Possibly, it is needed because it would have some credentials from push server and would have user token? As is, there are some hard to review parts/regression risks, especially in simplexmq, so maybe we can split to several PRs and reduce diff noise there.

1. Supporting user's ntf server is a good thing for self-hosters. Self-hosting the ntf server and the push service is a good thing regarding privacy. Because, for people who host their ntf server and their push server, for external observers, it is simply like a device were connected to SimpleX, and it isn't possible to do any network analysis to link SimpleX identities to push identities.
2. For the development of the feature, it was very useful to be able to change the ntf server.
3. ATM, the implementation requires to use a user ntf server, because I wasn't able to add webpush support to existing ntf servers (of course)
4. I think that, unlike xftp and smp servers, ntf servers weren't threaten as user servers until then for one reason: it was used for APNS only, and APNS requires a secret configured on their console. So it wasn't possible for anyone to self-host the ntf server _with SimpleX keys_, making self-hosting impossible.
5. It doesn't change how the ntf server works to fetch user messages, and works like for APNS. I think the whole point is, even with a compromise server (may it be a self-hosted one, or any ntfX.simplex.im), the impact is limited for user data, isn't it ?

> 2. Is there a scenario when user can have zero configuration push notifications, as this is what most people would prefer? E.g., if we decide to provide an option with similar security level as iOS, can we do it? Probably yes.

Yes, of course: if pre-recorded ntf servers (ntfX.simplex.im) support webpush, it is then possible to use the default push service on the device (which will be the play service if the user doesn't have any UnifiedPush service). But using some push services may have implications for the privacy (this is mainly about FCM), for specially targeted people. This is why it may be better to not enable by default - and it is probably better to add notifications settings in the onboarding process.

> 3. Am I right from reading this PR and [this doc](https://unifiedpush.org/news/20250131_push_for_decentralized/#webpush-to-google-services) that user can use google services without having any google services code in the app / without FCM libraries? This is indeed unknown to most people I think. And from what I read, technically it is possible to do without any closed-source code. This raises several questions:

You're right, this is why is PR adds FCM support without any proprietary lib

> *
>     * can a fully open-source app determine if it has google services to know that it can/should use FCM? Or any device, even without google services can use FCM by requesting a token? This I wasn't able to connect yet.

It is possible to determine if the device have the Play Services. With this PR, the Play Services is suggested only if it is installed on the device. In other words, the Google Services are a requirement to use FCM 

>  *
>     * to what extent this approach is consistent with play store policy? If the possibility of using web pushes is at least mentioned anywhere in google docs, then it may be consistent with the policy, but if it's completely undisclosed (I could not find any reference to it), then using it in play store app may be counter to play store policy. While this would not fall under "unauthorized use", because VAPID is used, it may fall under using undocumented APIs, and possibly under circumventing the APIs ("get rid of push gateways"), and also there may be additional data processed by the app (tokens etc.), that would need to be disclosed. So it needs to be assessed.

They provide this API for web application: https://firebase.google.com/docs/cloud-messaging/js/client , which is probably also used for PWA bundled in APKs. Note that the push gateways is often used to translate from on application protocol to firebase protocol, but in centralized software world, they just push to firebase without any gateway

> *
>     * will we need to have different builds - one for play store, another for github/fdroid etc. distribution? Possibly, it's a parameterized build. I didn't yet figure out how to embed the information into the code if it was built as APK (not play store) or bundle (for play store), but there is a parameter in gradle.properties that toggles a not-too-important thing. If any run-time difference is needed, relying on this parameter would be a bad idea, it's easy to forget to change.

F-Droid allow using our lib, so you don't need multiple flavors, and it gives FCM support to F-Droid users. This may be tagged as an anti-feature (promote non-free net), but I think it is easy to negotiate in this case, because it is opt-in and FCM is not promoted. The Mastodon application supports FCM without firebase messaging, and it isn't tagged for this.

> 
> 4. Adding a 4th option for notifications is one too many, but if the latest change for service notifications results in similar (or lower) battery usage as periodic fetch, we can just drop periodic completely.

I see the point of reducing options but I don't think it is a good idea to remove the option, because there will always be some users who prefer one option over the other, and I bet you'll have an issue to ask to restore the periodic fetch

> 5. UX: How will it be offered during the onboarding? If push is a good thing, then enabling it for the new users is more valuable that an option hidden somewhere that only advanced users would find and benefit from. And it's related to the question whether a self-hosted ntf server should be part of this change. My intuition is that it shouldn't be, and stage one would be releasing a zero-configuration support for push notifications (as battery/security trade off) that new users can enable from onboarding and existing users can enable from what's new (and it should happen _after_ the release that would recommend switching from periodic to service mode).

I agree, I think it would be valuable to add a screen during the onboarding to ask what they prefer. And I think it would be better if the default ntf servers support webpush, so they can be used even without self-hosting it.

If you web push support is added to the default ntf servers, I can add their VAPID fingerprint to the db migration, and remove the requirement for self-hosted ntf server (this is something I've been expected when opening the PR)

> 
> There are some smaller questions (e.g., I think there are simpler options to wrap custom servers into existing APIs) but, let's talk about big ones first.
> 
> What I like about this change is that it's rather surgical and seems to wire push support into the existing flows, unless I missed something. My initial reaction was that we should postpone this until after rewiring SMP/NTF notifications protocol from the current pull to push (see the braindump in [this rfc](https://github.com/simplex-chat/simplexmq/blob/stable/rfcs/2025-03-30-ios-notifications-3.md) - the current situation is better than at the time of writing RFC, as we moved to service subscriptions for ntf server, but other problems remain, and switching to push is still justified), but from looking at the code it seems like it may be completely unrelated to the proposed change.

It should be able to follow any modification to the NTF protocol. I think you should consider the PING ntf mode (max one per second), even for iOS, it should considerably reduce the number of your push notifications.

> 
> Let's talk.
> 
> Thank you!

Thanks !

epoberezkin:
> Do you think we can write it together?

Yes, good idea. I will make a draft of what I think it should be (based on all the comments), and what's the gap with the PR.

> Supporting user's ntf server is a good thing for self-hosters.

100%, I am not debating this, and it should be added. I am only arguing for splitting it to a separate PR and beta release, not even to a separate stable release. It would reduce the PR size, and would allow us to test and stabilize webpush support on the server(s) we can debug and monitor prior to shipping all the potential problems to the end users.

To use custom server for development it just needs to be replaced in the source code.

> and it is probably better to add notifications settings in the onboarding process.

Yes, I think it's necessary for the release.

> They provide this API for web application: https://firebase.google.com/docs/cloud-messaging/js/client

Yes, but it is not documenting web push endpoints, we only know them via what they might qualify as reverse engineering. On another hand, the stance that open-source code is equivalent to public documentation is also viable, but no easily findable precedents where it was tested. Are there any widely used apps (over 1m downloads) that use this approach/library to push to FCM? We could take a stance that using parts of opensource code falls under the license, and is a common practice, and doesn't not circumvent the requirement for authorization, so it is an acceptable use within the policy...

The problem here is that when google finds you in non-compliance with their policies they give a very tight deadline to comply, and it would be non-trivial to swap this FCM implementation with FCM libraries. I think we might be better off by completely removing FCM support from the initial scope, and instead do this:

- add support for web-push in all preconfigured ntf servers (and add ability to use others of course).
- deploy our unified push servers and have them as preconfigured.
- add the code into the app that receives unified pushes too.

This is also more valuable for the current stage than supporting FCM.

Then default zero configuration option would be to use no other servers (and no self-hosted servers) and to use the app itself as the push recipient, but give users an option to receive pushes via another app, use other push servers, and use other ntf servers. I will sketch how it might look in the onboarding screen, but in short - three options ("When app running", "Instant with on-device service", "Push notifications", with default being "Instant", as it's best for security and good for battery, given the latest change), with clickable link for advanced configuration screen that would allow adding all three configuration options:
- ntf server
- push server - possibly, we could include some widely used community servers here?
- app to receive pushes, with SimpleX being the default.

At some later point we could 1) seek explicit clarification from google about their stance on using these endpoints (we can do it now, if we agree in general) 2) use them if they confirm it's within the policy, or use their FCM library if not, and only in playstore build.

> This may be tagged as an anti-feature (promote non-free net), but I think it is easy to negotiate in this case, because it is opt-in and FCM is not promoted.

I am not worried about it, having options is good, I'd be surprised if F-Droid objects, but we may ask in advance. Also, FCM would know absolutely nothing other than it's used and how many notifications - notifications are "pings" with encrypted metadata telling the app where to get the message (you probably know that).

> The Mastodon application supports FCM without firebase messaging, and it isn't tagged for this.

Do they do it in playstore app?

> I bet you'll have an issue to ask to restore the periodic fetch

Not if it uses more battery than instant. And I don't have problem saying "no" to features :) So we could deprecate it now-ish (that is allow using it if it was enabled, but not allow switching to it), and then add push in 6.5.

> If you web push support is added to the default ntf servers, I can add their VAPID fingerprint to the db migration, and remove the requirement for self-hosted ntf server (this is something I've been expected when opening the PR)

Cool, we seem in agreement here.

> I think you should consider the PING ntf mode (max one per second), even for iOS, it should considerably reduce the number of your push notifications.

It already does some coalescing and bundling, primarily to compensate for Apple dropping notifications, and it works well - on iOS you may see notifications that say "you have messages from a, b, c" (iOS only allows to show only one notification for each received, when app is suspended). The thing that would have a major impact on the number of pushes is including message priority into the messages (instead of current T/F flag), so that groups that have mention-only notifications only send them on mentions, and not on all messages as they do now (it's described in RFC).

I think that for users in large groups (even once we migrate to chat relays) it may have more impact on battery usage than supporting push :)

Another question: I don't see where in the PR the app calls apiGetNtfConns/apiGetConnNtfMessages when notification arrives (and there are some events too)? Or how does it work? It looks like it simply treats notifications as wake-up pings, and subscribes to all queues, as usual. This is probably ok for most users, especially after migrating large groups to chat relays.

p1gp1g:
OK, so let's split this PR. I will open a first PR with support with the pre-recorded ntf server. Can you host a ntf server with webpush support for the beta/dev version and share its url ?

I can remove support for FCM first, this is a single line in the gradle file (one import)

---

> > They provide this API for web application: https://firebase.google.com/docs/cloud-messaging/js/client
> 
> Yes, but it is not documenting web push endpoints, we only know them via what they might qualify as reverse engineering. On another hand, the stance that open-source code is equivalent to public documentation is also viable, but no easily findable precedents where it was tested. Are there any widely used apps (over 1m downloads) that use this approach/library to push to FCM? We could take a stance that using parts of opensource code falls under the license, and is a common practice, and doesn't not circumvent the requirement for authorization, so it is an acceptable use within the policy...
> 

<details><summary>Full reply</summary>

> The problem here is that when google finds you in non-compliance with their policies they give a very tight deadline to comply, and it would be non-trivial to swap this FCM implementation with FCM libraries. I think we might be better off by completely removing FCM support from the initial scope, and instead do this:
> 
>     * add support for web-push in all preconfigured ntf servers (and add ability to use others of course).
> 
>     * deploy our unified push servers and have them as preconfigured.
> 
>     * add the code into the app that receives unified pushes too.
> 
> 
> This is also more valuable for the current stage than supporting FCM.
> 
> Then default zero configuration option would be to use no other servers (and no self-hosted servers) and to use the app itself as the push recipient, but give users an option to receive pushes via another app, use other push servers, and use other ntf servers. I will sketch how it might look in the onboarding screen, but in short - three options ("When app running", "Instant with on-device service", "Push notifications", with default being "Instant", as it's best for security and good for battery, given the latest change), with clickable link for advanced configuration screen that would allow adding all three configuration options:
> 
>     * ntf server
> 
>     * push server - possibly, we could include some widely used community servers here?
> 
>     * app to receive pushes, with SimpleX being the default.
> 
> 
> At some later point we could 1) seek explicit clarification from google about their stance on using these endpoints (we can do it now, if we agree in general) 2) use them if they confirm it's within the policy, or use their FCM library if not, and only in playstore build.
> 

</details>

Most of firebase messaging is open source, and the intents are here: https://github.com/firebase/firebase-android-sdk/blob/c85227e344230aa676066352184bc6ccd510294f/firebase-messaging/src/main/java/com/google/firebase/messaging/GmsRpc.java#L264

And their flutter lib shows how to use the vapid key: https://github.com/firebase/quickstart-flutter/blob/446125c030038b429816737033e43aa36336fa05/messaging/lib/main.dart#L159

And IIRC, it is possible to add asym keys to do authorizations on the firebase console, but I don't remember how to do it.

I understand if you want to clarify with them first. But it needs to be done the right way, because that's a really needed feature for decentralized apps, and I need to use it for another project too :)

> > The Mastodon application supports FCM without firebase messaging, and it isn't tagged for this.
> 
> Do they do it in playstore app?

Yes, but they use the FCM API (see [their implementation](https://github.com/mastodon/mastodon-android/blob/2c9e92a254907772757cdc932fd37667fc9218bb/mastodon/src/main/java/org/joinmastodon/android/PushNotificationReceiver.java#L42)), and not the VAPID based one

> > I bet you'll have an issue to ask to restore the periodic fetch
> 
> Not if it uses more battery than instant. And I don't have problem saying "no" to features :) So we could deprecate it now-ish (that is allow using it if it was enabled, but not allow switching to it), and then add push in 6.5.

As you wish, I think you have a better view on how you'd like to manage your app. At least, maybe it should be deprecated only once push is available

> > I think you should consider the PING ntf mode (max one per second), even for iOS, it should considerably reduce the number of your push notifications.
> 
> It already does some coalescing and bundling, primarily to compensate for Apple dropping notifications, and it works well - on iOS you may see notifications that say "you have messages from a, b, c" (iOS only allows to show only one notification for each received, when app is suspended). The thing that would have a major impact on the number of pushes is including message priority into the messages (instead of current T/F flag), so that groups that have mention-only notifications only send them on mentions, and not on all messages as they do now (it's described in RFC).
> 
> I think that for users in large groups (even once we migrate to chat relays) it may have more impact on battery usage than supporting push :)
> 
> Another question: I don't see where in the PR the app calls apiGetNtfConns/apiGetConnNtfMessages when notification arrives (and there are some events too)? Or how does it work? It looks like it simply treats notifications as wake-up pings, and subscribes to all queues, as usual. This is probably ok for most users, especially after migrating large groups to chat relays.

I see a little confusion here, between the UI notification ("you have x messages") and the push notification (which usually ends with a UI message).

What I meant by PING ntf mode is a dummy message sent from the ntf server to the device to simply wake it up. Then the device connect to its relays for a few seconds. We usually call that "sync-on-push". This is what Signal does for instance. With that solution, you don't need to send a push notification per message (for instant messaging, this is common to see many messages being sent within the same second), you don't need to store messages on the server and you don't need to send actual data with push messages.

This is actually the strategy that is already implemented, and why the app never call apiGetNtfConns or apiGetNtfMessages

> This is probably ok for most users, especially after migrating large groups to chat relays.

This is why I see push support and connection efficiency as 2 complementary features

epoberezkin:
> OK, so let's split this PR.

I think it would be most productive if we first agree the scope and all release steps - that's what rfc is for. Once we have it all agreed, it would be easier and less iterative. Maybe we need feature branch so I can review in stages - it'll be all clear once we align on the plan.

> Can you host a ntf server with webpush support for the beta/dev version and share its url ?

Yes, we have a test server - once we have the code we'll release there.

> I understand if you want to clarify with them first.

No, I actually don't :) it was just an idea. What you write increases confidence, and maybe it's all good and we should just do it. Let me sleep on it (literally:).

> Yes, but they use the FCM API (see [their implementation](https://github.com/mastodon/mastodon-android/blob/2c9e92a254907772757cdc932fd37667fc9218bb/mastodon/src/main/java/org/joinmastodon/android/PushNotificationReceiver.java#L42)), and not the VAPID based one

And that reduces confidence :)

> At least, maybe it should be deprecated only once push is available

Yes, possibly together.

> What I meant by PING ntf mode is a dummy message sent from the ntf server to the device to simply wake it up.

The problem why it won't work on iOS is that the device has per-connection (queue) credentials, so it would have to subscribe to all queues to get any messages. Signal can send pings because there is a "user", and a single authentication for all messages. With SimpleX it has to authenticate for each connection separately, and even though it bundles ~135 authentications in the block, it is likely to be not good enough with the current design for large groups to fit in iOS 30 sec processing limit. Also, on iOS the notification service is a separate process, with its own set of opened sockets, so it cannot simply subscribe to these connections - it would disrupt subscriptions that exist in the sockets opened by the app (they remain connected even while the app is suspended, so it doesn't have to resubscribe when the app wakes up). So the current design takes into account both lack of user identity/authentication in the protocol and iOS constraints - it doesn't subscribe at all, it only fetches messages (GET instead of SUB), and only from connection(s) mentioned in the notification. For Android it should be ok - I don't think there is any time limit for processing pushes, and it is not another process too - it is all one app.

We could in theory have a different mode of notification delivery for web push tokens that does more aggressive coalescing by time, and maybe doesn't even send connection information, but I don't think it's worth it - I'll think about it. Introducing message priority is important though, as it would reduce the number of pushes much more than simply coalescing by time.

epoberezkin:
... I'll send rfc draft this week.

balintbarna:
@epoberezkin  I'd recommend that you try using an app with UnifiedPush to see how it works.

This infographic helps a lot: https://unifiedpush.org/developers/intro/

Normally the user has 1 app that is the unified push (UP) distributor on their device. For me this is Sunup but another popular alternative is ntfy. This app connects to the push server, so this is where it's decided which push server is used - correct me if I'm wrong though. Sunup connects to Mozilla push servers while Ntfy allows configuring which Ntfy servers are used, and that can also be self-hosted.

> push server - possibly, we could include some widely used community servers here?

If I understood your comments about the user choosing the push server in the SimpleX app, I don't think that makes sense. The SimpleX servers should send a web push to the user's (in the UP app) chosen push server, that talks to the user's UP app which then wakes up the SimpleX app. For users that choose Sunup it requires no configuration and for users using Ntfy they should have it configured already for their other apps.

So when I use Molly (FOSS Signal app) I just select that I want to use Unified Push - instead of direct socket connection - and nothing else needs to be configured, it automatically registers with Sunup on my phone and Sunup also doesn't require any configuration. Same with Mercurygram, ElementX, and FindMyDevice. The experience should be the same with SimpleX.

Apologies if I misunderstood your earlier comment.

I think it's good to have FCM as an alternative for the less-technical people who don't want to take the time to understand how to use Unified Push and don't want the battery drain resulting from 1-1 connections because this can increase adoption of SimpleX, but if it's implemented it should not use any proprietary libraries, should not be the default option (because of privacy) and it should be clearly understandable that it's an option using Google servers.

Thank you both for your work on this :)

epoberezkin:
@balintbarna I understand how UnifiedPush works, and I understand your perspective on how UnifiedPush should be supported in the app - because you already use UnifiedPush. But this is not what a majority of SimpleX Android app users would think - most of them do not use UnifiedPush, and never heard about it.

So for them, push notifications that require to have some other app installed and have any additional configuration is a very bad UX. I heard this argument a lot about Tor, but my counter-argument was that if installing Orbot feels like a complex task for somebody, then they should avoid Tor completely - it is an advanced technology, and to get security benefits from it requires *some* technical expertise. At least that's what I believe. Push notifications are the opposite - they improve usability (lower batter usage) and convenience (you don't have to charge the device as frequently) at the cost of reducing security. So it should not require any technical expertise, and it is not a good approach to require another app installation for UnifiedPush to be used.

UnifiedPush supporters say to that - just let them use FCM. But again, it's bad trade-off for the majority of the users, and also for us. Users would have to settle for trusting their data to yet another company - all of them would prefer the network operators they already use over FCM servers - and we would either have to accept the risk of being in non-compliance with PlayStore policy if we use an open-source FCM distributor or we would have to maintain two different builds. I disagree that the users who want it simple should use FCM, I do think that we should see embedded "UnifiedPush distributor" as an MVP for UnifiedPush support.

While I am 100% in support of allowing users to use another push distributor (unlike it is in Conversations, that also acts as a distributor) - ntfy.sh or sunup or any other - the default option must be to use an embedded distributor, for zero-configuration - so that if the users choose "push notifications" during onboarding and do not modify any settings, they would be using servers provided by any of the preconfigured network operators.

At the same time, we don't have to make embedded distributor usable by other apps, if it's technically possible - we can reduce MVP scope by only making it usable by SimpleX app itself.

epoberezkin:
@p1gp1g https://github.com/simplex-chat/simplex-chat/pull/6213 - this is the doc :)

p1gp1g:
Regarding push notifications, an application either:

- Receive push notifications through another application, the distributor or push service
- Provide push notifications to other applications, becoming a distributor

It is also possible for an application to be able to work in both way - if the user have another distributor, it uses it to receive push, and if the app receive subscriptions, it works as a distributor.

As you mentioned, Conversation works only as a distributor, and I think it would be nice if they are also able to use other distributors to receive them.

So, this PR is about receiving push notifications, and not providing them.

Providing push notifications is also a great feature: the more applications provide push notifications, the more likely users have a push service installed on their system without even knowing it. And if you wish to make SimpleX a distributor as well, it would be a good thing in that direction. But this is another feature.

When an application maintain a connection to a server to get their own notifications, it isn't push notifications anymore. Having an "embedded distributor" that maintain its connection may makes sens when the application doesn't have any (optimized) way to maintain its connection yet, to have a single interface to manage notifications. But if your connection is already optimized, you won't gain a lot or anything. You only change how your foreground service works. It is in the core of push notifications to use an external service.

## About fallbacks

So, with the number of application working as a distributor growing up, alternative Android OS coming with a default distrib, and applications supporting UnifiedPush that advertise distributors, the number of users having a push distributor without really knowing it is even possible.

But of course, some users interested in getting push notifications won't have one and here comes the question of fallbacks

#### FCM

The FCM fallback may be a good option as many user have already a service for push notifications: the Play Services.

Regarding the license to use their web push API (you mention in the spec file), it is worth mentioning that Google was one of the author of webpush and vapid specs, this is also why they support this API. And we can see in their [open source code comments](https://github.com/firebase/flutterfire/blob/8c0802d098c970740a34e83952f56dbe9eb279fd/packages/firebase_messaging/firebase_messaging_platform_interface/lib/src/method_channel/method_channel_messaging.dart#L242) that they plan to have clear functions to use VAPID with their library, even on Android. The intents to registers are open source in the Apache 2.0 lib, and no code is not copy/pasted from there.

But I deeply understand concerns to provide push notifications with this service, regarding privacy and centralization. It may be OK for many users, even with a clear understanding of the meaning. But this is of course as you wish.

#### Other fallback

As a fallback you suggest to embed a distributor that keep a connection. As I said above, this is pretty misleading since it isn't push notifications anymore. It just changes how your foreground service works. And I don't think it is a good solution for SimpleX following recent changes.

I think push notifications can either be disabled if no push service (distrib) is installed. So the user will have to use the foreground service, which should have been improved recently, and that's OK to use it.

Or another solution would be to have a one-click way to install another distributor. For example:

- if F-Droid is installed, open the Sunup page
- else download and install Sunup -> if that's needed, I'll provide a flavor with a self-updatable app

#### My opinion

On my side, I think the best way would be to:

- disable the feature until a distrib is installed, with a pop up explaining it
- if F-Droid is installed, there could be a deep link to open Sunup there (it may works with Accrescent too, but I've to publish Sunup first)

So, the default would be to use the internal foreground service if no distrib is installed, and the onboarding should not offer push in this case

epoberezkin:
> When an application maintain a connection to a server to get their own notifications, it isn't push notifications anymore. Having an "embedded distributor" that maintain its connection may makes sens when the application doesn't have any (optimized) way to maintain its connection yet, to have a single interface to manage notifications. But if your connection is already optimized, you won't gain a lot or anything. You only change how your foreground service works. It is in the core of push notifications to use an external service.

Maybe we talk about slightly different things. The app will be dormant for all purposes other than receiving push notifications, so it won't consume any more battery than a distributor app would (that most users don't have installed, and won't want to install). There is no way to optimize the app to the same level as only receiving pushes can be optimized, as while messaging part is running it is receiving messages to all conversations, including muted, and has to process all events (edits, deletions, reactions, etc. - even when they don't result in notifications the user wants to see). And while only push part is running, it won't start receiving messages until an important message arrives that should result in notification. So the difference in battery usage can be substantial, however we optimize messaging part. RFC tries to explain it, but maybe it's not clear enough.

> Providing push notifications is also a great feature: the more applications provide push notifications, the more likely users have a push service installed on their system without even knowing it. And if you wish to make SimpleX a distributor as well, it would be a good thing in that direction. But this is another feature.

I understand that from the perspective of "solution" it's two different features. But from the point of view of most users who don't already use UnifiedPush it is all one feature - "an ability to save battery via push notifications, independent from Google FCM". And if using this feature requires installing some other app, most users will see it as half-baked and not usable. From our tech bubble it may seem that most people already use or know about UnifiedPush, but in a bigger bubble of all SimpleX Chat users it is not the case yet. By offering an ability to use UnifiedPush without any third party app we help them all.

> As a fallback you suggest to embed a distributor that keep a connection. As I said above, this is pretty misleading since it isn't push notifications anymore.

Why? That's how push notifications work on iOS - the app itself (or, to be more precise, a separate part of the app) acts as push recipient/distributor and when push notification arrives, it will start receiving messages. So in this case we would also have a constantly running part that only receives push notifications, while message reception would be fully suspended (and we may even disconnect from the servers, to prevent the app waking app on all message events). And messaging part would only be starting when push notification is received. Not sure why it's not "push notifications" - they are pushed to the app by webpush server, and running the app in "push only" mode would substantially save the battery compared with running it in "receiving messages" mode - at least 3x, or maybe even more.

> disable the feature until a distrib is installed, with a pop up explaining it

Let's talk how we can make it happen as a full feature, that would benefit all users, and not as a part feature, that about 10% of users want. F-droid as a whole is a minority of users, so releasing the feature that would only benefit F-droid users seems wrong.

balintbarna:
@epoberezkin 
> Let's talk how we can make it happen as a full feature, that would benefit all users, and not as a part feature, that about 10% of users want. F-droid as a whole is a minority of users, so releasing the feature that would only benefit F-droid users seems wrong.

This is twisting what @p1gp1g was saying, the feature (as a whole) is not for F-droid users, it's for all the people who want to save battery by using a **common unified push distributor** on their phone. To me it seems very clear that issue #1081 asks for exactly that. @p1gp1g  was suggesting that the onboarding could include a link to install a very popular UP distributor from F-droid for people who use F-droid, but that's a small convenience feature and doesn't equate the whole feature to being for F-droid users only, as people can install UP distributor apps from any other source. But many of the people who would use SimpleX would install apps from F-droid repos, I assume that's why SimpleX also has an F-droid repo. SimpleX maintains F-droid repos so it's reasonable to provide a helpful link for F-droid users in the onboarding, but it's not necessary.

> Maybe we talk about slightly different things. The app will be dormant for all purposes other than receiving push notifications, so it won't consume any more battery than a distributor app would (that most users don't have installed, and won't want to install). There is no way to optimize the app to the same level as only receiving pushes can be optimized, as while messaging part is running it is receiving messages to all conversations, including muted, and has to process all events (edits, deletions, reactions, etc. - even when they don't result in notifications the user wants to see). And while only push part is running, it won't start receiving messages until an important message arrives that should result in notification. So the difference in battery usage can be substantial, however we optimize messaging part. RFC tries to explain it, but maybe it's not clear enough.

The issue describes UP as an alternative to FCM. Using FCM the notification is sent to Google's servers and then to Play Services app, which then wakes up the app. With using UP as an alternative the notification is pushed to the UP server which has a connection with the UP distributor app which then wakes up the app. The battery saving comes from having one persistent connection (UP server and app) instead of each messaging app (9 on my phone) having its own persistent connection. FCM is not an option for people who want to degoogle and run on FOSS as much as possible. What your reply here suggests to me is that you misunderstood what people were asking for and why. SimpleX can already receive notifications from SimpleX servers but  it requires its own persistent connection draining battery, which is what issue #1081 is about. It will still need to process notifications afterwards to display them to the user in a meaningful way, unless you want to embed a ton of information in the push notification, which can be worse for privacy.

> I understand that from the perspective of "solution" it's two different features. But from the point of view of most users who don't already use UnifiedPush it is all one feature - "an ability to save battery via push notifications, independent from Google FCM". And if using this feature requires installing some other app, most users will see it as half-baked and not usable. From our tech bubble it may seem that most people already use or know about UnifiedPush, but in a bigger bubble of all SimpleX Chat users it is not the case yet. By offering an ability to use UnifiedPush without any third party app we help them all.

Having SimpleX servers as UP servers and the app as a distributor is redundant and unnecessary. It can be done, but I expect it to not bring substantial improvements over the current notification system, because both require a persistent connection and both works. From my understanding this pull request is made to resolve issue #1081 and that is done by SimpleX working with UP servers and distributor apps. In my view making it a distributor is bloat. Talking about "the majority" of SimpleX users not wanting this feature when it is the most commented and upvoted issue is confusing.

pm4rcin:
> Maybe we talk about slightly different things. The app will be dormant for all purposes other than receiving push notifications, so it won't consume any more battery than a distributor app would (that most users don't have installed, and won't want to install). There is no way to optimize the app to the same level as only receiving pushes can be optimized, as while messaging part is running it is receiving messages to all conversations, including muted, and has to process all events (edits, deletions, reactions, etc. - even when they don't result in notifications the user wants to see). And while only push part is running, it won't start receiving messages until an important message arrives that should result in notification. So the difference in battery usage can be substantial, however we optimize messaging part. RFC tries to explain it, but maybe it's not clear enough.

From my understanding the problematic part is not actually the notifications itself but the need for continuous processing of various events like you mentioned that has to happen on user device **regardless** of notifications being displayed. I'm not sure how Signal does it but it may be different because SimpleX's general architecture of being more decentralized and focused on user device. Perhaps something like a mailbox would be needed (Briar has something like that) but it's probably not going to be implemented for SimpleX. So either some compromise is needed or completely novel idea. That's what you meant?
PS. I've read rfc.

EDIT: Isn't the fundamental problem that SimpleX is so much different from other messaging apps that you can't just add support for UnifiedPush and call it a day?

epoberezkin:
@balintbarna 

> This is twisting what @p1gp1g was saying, the feature (as a whole) is not for F-droid users, it's for all the people who want to save battery by using a common unified push distributor on their phone.

I am not twisting anything :) And I understand that it's not only for F-Droid users. But the overlap between F-Droid users and UnifiedPush users is quite high, and the share they represent amongst SimpleX app users is also similar.

So to me, the starting point for UnifiedPush intergration would be a self-contained app design, when an app acts as push receiver, and allows external distributors as an advanced configuration option.

For most users, having external distributor as the only option is a bad UX they won't use at all.

> To me it seems very clear that issue https://github.com/simplex-chat/simplex-chat/issues/1081 asks for exactly that.

It's not really important. We do not treat issues as TODO list, unless they are important bugs. We treat them as a collection of ideas for improving the app for all users.

> the onboarding could include a link to install a very popular UP distributor

Asking to install another app during onboarding to use an important feature is a bad UX, we won't do that. That it's from F-Droid or from Play Store is of secondary importance. It must be a self-contained feature, that everybody who installs the app can use, without any additional pre-requisites. As extension of this feature, we would also allow using external distributors to the users who want it.

> The battery saving comes from having one persistent connection (UP server and app) instead of each messaging app (9 on my phone) having its own persistent connection.

I can only repeat, that for about 90% of SimpleX Chat users the saving from having a single distributor outside of SimpleX app would be almost the same (in fact, worse, because it would be two different processes instead of one) as from having a single distributor embedded in SimpleX app.

I don't understand why you are defending what is a bad UX for absolute majority of users (the need to install another app). I am not arguing against providing everybody who wants to use external distributor such an option - I 100% agree that you should not be forced to use embedded distributor. But similarly, I do not want to ask SimpleX app users to install any other app as a condition to use push notifications. Why is it so controversial?

> Having SimpleX servers as UP servers and the app as a distributor is redundant and unnecessary.

For people who have UnifiedPush distributor installed and want to use it with SimpleX app - that may be redundant. But you don't have to use an embedded distributor. And it's necessary for the remaining 90% of the users who don't know what UnifiedPush is, and won't be able to use it unless the app has embedded push receiver (and I don't see the need to also be a distributor for other apps - that's indeed unnecessary, and is better done via specialized apps). This is the same approach that resulted in Tor adoption - embedding it in apps like Cwtch. That's the only approach that may result in wider UnifiedPush adoption - make it usable to non-technical people via a single app they want to use, without requiring to install any additional apps.

So what I want to do is almost the opposite to what Conversations did:
- you can use unified push in SimpleX app without any 3rd party apps (same as Conversations).
- you can use external distributors, e.g. ntfy.sh or sunup with SimpleX app (Conversation does not allow it).
- you CANNOT use SimpleX app as distributors for other apps (something that Conversation allows) - that may be added later, but I don't see much value for that - everybody who wants to have a single distributor already uses some app. But my priority is to make this feature usable for everybody else, which is 90% of app users.

I think it's reasonable, and would increase usage of UnifiedPush spec via SimpleX app like 10x compared with requiring to use external distributor - so it seems like win win to me.

@pm4rcin wrote:

> From my understanding the problematic part is not actually the notifications itself but the need for continuous processing of various events like you mentioned that has to happen on user device regardless of notifications being displayed.

You are right, it's a large part of the problem. But it's not even the number of events, but also the number of reconnections. If the app only needs to connect when it knows there is something that will show as a notification, even if it has to process all events to get to that notification, it would still consume a lot less battery.

For future large groups we are currently working on, we already have two messaging queues with chat relays - high and low priority, where high priority will only be used for important messages that must trigger notifications (e.g., mentions) and also for RPC commands (such as loading history etc), and low priority for everything else. In this case the app would only need to process high priority queue if the group has "only mentions" notifications mode, which is how most users use public groups.

> I'm not sure how Signal does it but it may be different because SimpleX's general architecture of being more decentralized and focused on user device.

also correct, and Signal is centralized.

> Isn't the fundamental problem that SimpleX is so much different from other messaging apps that you can't just add support for UnifiedPush and call it a day?

No, that's not really the problem. I think fundamental difference here is philosophical rather than technical - we are really committed to maximizing the value to the end users, and the value is defined as the value per user and the share of users who would benefit from the feature. So we are trying really hard to avoid any features that would only help a minority of users, and focus on features that would benefit the majority of the users.

So if we can increase the cost of integrating UnifiedPush by say 50% (all we have to do is to embed a working ntfy.sh code into the app, and implement full suspension of messaging when the app is in background (probably not instantly, but say after 5 min) and it increases the value of UnifiedPush adoption by say 10x, then we absolutely should do it.

It's not ready yet anyway (see [RFC](https://github.com/simplex-chat/simplex-chat/pull/6213)), so if we do do it, it should create value to all SimpleX app users, and not only to UnifiedPush users.

epoberezkin:
From a bit of research, the app architecture for an embedded distributor can be even simpler than I thought by making an app and embedded distributor two memory-isolated processes.

This way, if the user chooses to use an external distributor, the internal one would never start. But if the user chooses to use an embedded distributor, it would start as a separate process, and would communicate with the app in the same way as ntfy.sh does, or maybe with minor modifications.

The code for UnifiedPush distributor would be taken from ntfy.sh as is, with minor adaptations, and when the app goes into background it would kill the app process and it will be re-launched by distributor process when push notification arrives.

The advantages of this approach over trying to run both parts in the same process:
- better memory management - haskell part would fully clear the memory on exit, without any heap fragmentation.
- we don't need to do anything in haskell to support "receiving push" mode - we simply kill that process when it is not needed, so it won't have anything running, or any network connections, etc.
- clean start when push notification arrives - stopping and disconnecting from servers without actually killing the core might have some complexities or residual state, etc. If it starts clean every time, it would work better.
- periodic starts sometimes resumes message delivery, e.g. when there are stale sockets that the app didn't detect yet, so this mode of operation would combine both instant message delivery but also higher reliability of delivery - an additional value of using UnifiedPush compared with service mode.

This architecture is somewhat similar to iOS app, but without shared database access, and without actual memory reception in embedded distributor process, like iOS app has to do in Notification Service Extension - it both receives pushes and receives messages, and also share database access with the main app, which is very complex. Android architecture will be quite simple in comparison.

Even though they are two memory-isolated processes, they still have the same UID, so share access to resources - files and preferences - so all necessary configuration to the distributor process can be done from the main app UI, no need for a separate activity, and then we can communicate via IPC that distributor has to read the updated configuration or to stop/restart, e.g. if the user switches to another notifications mode or to external distributor.

So it's not even 1.5 cost, but more like 1.25 cost - we take ntfy.sh code, make it run as a process, being part of SimpleX app, and that's about it.

pm4rcin:
So basically you want to replace notifications code in SimpleX by adopting UnifiedPush **standard**. And the additional value is that user can use any distributor or simply stick to the "previous" behavior but redone and probably improved. Correct?

epoberezkin:
> So basically you want to replace notifications code in SimpleX by adopting UnifiedPush standard. And the additional value is that user can use any distributor or simply stick to the "previous" behavior but redone and probably improved. Correct?

No, I don't want to replace it. The current "service" mode would remain, as it's a viable tradeoff that is better for some users. I want to offer a choice between two approaches to notifications, as seen by the users:

- background service: best for [meta-data] security, and good for battery, without any additional servers.
- push notifications: best for battery, good for security, without google FCM servers, and without any configuration required - "batteries included". But it would provide advanced users the option to receive the notifications via an external app and use self-hosted servers (both notifications subscriber and webpush)

I think it's really cool, and would result in a 10x higher adoption of UnifiedPush.

epoberezkin:
And as a positive side effect that may even put some pressure on some bigger apps to do the same, and use their own servers for push.

Signal would never adopt unified push if it requires using an external app. Signal might switch to unified push, or at least offer it as an option if it is all self-contained.

balintbarna:
@epoberezkin 

Thanks for the clarification.

I also thought after your reply to me that you want to replace the current notification system with the integrated unified push receiver that uses SimpleX servers. I could see the benefit in that by switching to a standardized protocol and reusing existing ntfy.sh code. But if that's not the goal and you still want to have the current solution, and an integrated unified push receiver, and also work with external distributors, then it seems like a lot of overlap in options. After all if the integrated UP solution becomes default and uses SimpleX servers, what would be the benefit of keeping the current system as well? And if there are benefits to the current system, what's the point of adding an integrated UP receiver, if that would be the worse option?

That point aside, as long as the final solution will work with an external distributor like Sunup I'm perfectly good with it.

Thanks for your work.

epoberezkin:
> what would be the benefit of keeping the current system as well?

Substantially better metadata security - there is no identifiers per app or even per profile if only messaging servers are used.

> And if there are benefits to the current system, what's the point of adding an integrated UP receiver, if that would be the worse option?

Lower battery usage, but without google, and without any advanced configuration - it's unified push that anybody can use, and not only people who are willing to install another app, which I see as maybe 10% of the users (at best).

It addresses the tradeoff between security and battery usage, making both options simple to use.

balintbarna:
@epoberezkin is it impossible to use the same identifiers with UP as with the custom implementation and have the same security and anonymity?

ismay-tue:
## User perspective on unified push

I've been following this PR and the related issue closely, and thought I'd offer my perspective as a user. I use android without the play store and fcm, and I think that the group doing that probably overlaps quite a bit with people who find simplex interesting.

One thing that unified push adds that I find very useful is the ability to use a single push notification server for multiple apps, and to swap it out at will. I use unified push for molly and element, and can easily switch from nextpush to ntfy to sunup or something else. Element and molly are not tied to the specific solution I use, as long as it implements unified push. That saves battery, and gives me a choice of notification provider.

### Distributor

Now one thing that I see happening in this PR that I personally would recommend against as a user, is integrating a unified push distributor in the app. As I see it that does not really add that much value. In my perspective the core benefit of universal push is exactly the independence from the apps that use it for notifications.

Note that in the linked issue: https://github.com/simplex-chat/simplex-chat/issues/1081, no one seems to be asking for simplex to become a distributor. See also a similar discussion [here](https://codeberg.org/iNPUTmice/Conversations/issues/216#issuecomment-2155140), where users of conversations are frustrated because conversations insists on being a distributor, whilst not allowing switching to a different distributor. Integrating a distributor is potentially nice to have. But the crucial feature request here is allowing the user to subscribe to a distributor of _their_ choice.

### Maintenance burden

Now from what I understand, the intention is to still allow for another distributor to be chosen, so users at least have a choice. But that still leaves you with a maintenance burden for code that is potentially just dead code. So I'd suggest first starting with the minimal addition: just allowing users to pick an external unified push provider. You can always add the distributor later if it is the more privacy conscious option, potentially even as a separate app (e.g. SimplexPush).

FlyingWombat:
@p1gp1g  Please split out the Google FCM support into a separate PR. There seems to be some disagreement or unresolved points about FCM here, so splitting it out will let UnifiedPush support move forward independently.

@epoberezkin Please merge this PR for UnfiedPush support as an option in the notifications settings menu for advanced users; off by default. No user on-boarding. No UX changes. No embedded ntfy.sh distributor. Just an option for regular Unified Push as an alternative to the built-in background message service.

It is clear that you care deeply about this project and making the best user experience possible. But incremental improvements are better than stalling on trying to work out every angle perfectly. Merging UnifiedPush now as an advanced option can also serve as a sort of beta test for the feature. Then, once any issues are worked out, all the other stuff like on-boarding for casual users, whether or not to support Google FCM, UX changes, etc. can be worked out later.

PR #6211, opened 2025-08-22
core: forward member name; ui: don't show "Member" label if unknown member was created with name

spaced4ndy:


src/Simplex/Chat/Library/Internal.hs
```diff
@@ -1196,9 +1196,15 @@ sendHistory user gInfo@GroupInfo {groupId, membership} m@GroupMember {activeConn
...
                   GroupMember {memberId} = sender
-                  msgForwardEvents = map (\cm -> XGrpMsgForward memberId cm itemTs) (xMsgNewChatMsg : fileDescrChatMsgs)
+                  memberName = Just $ memberShortenedName sender
+                  msgForwardEvents = map (\cm -> XGrpMsgForward memberId memberName cm itemTs) (xMsgNewChatMsg : fileDescrChatMsgs)
               pure msgForwardEvents
 
+memberShortenedName :: GroupMember -> ContactName
+memberShortenedName GroupMember {memberProfile = LocalProfile {displayName}}
+  | T.length displayName <= 16 = displayName
+  | otherwise = T.take 16 displayName <> "…"
```
```suggestion
  | otherwise = T.take 15 displayName `T.snoc` '…'
```

apps/ios/SimpleXChat/ChatTypes.swift
```diff
@@ -2555,7 +2555,7 @@ public struct GroupMember: Identifiable, Decodable, Hashable {
 
     private func pastMember(_ name: String) -> String {
         memberStatus == .memUnknown
-        ? String.localizedStringWithFormat(NSLocalizedString("Past member %@", comment: "past/unknown group member"), name)
+        ? String.localizedStringWithFormat(NSLocalizedString("Member %@", comment: "past/unknown group member"), name)
```
epoberezkin: Isn't name optional here? What would it show it there is no name?

PR #6243, opened 2025-09-05
cli: picolisp lib support

shumvgolove:
Picolisp setup:

1. Install deps:

    ```sh
    sudo apt install binutils make clang llvm libreadline-dev libffi-dev libssl-dev pkg-config
    ```

2. Get source code, unpack it:

    ```sh
    wget https://software-lab.de/pil21.tgz
    tar xfz pil21.tgz
    cd pil21
    ```

4. Enable "quiche" mode:

    ```sh
    vim src/Makefile
    ```

    ```diff
    base.ll: $(INC) $(SRC) lib/ex.l
    -   $(PIL) lib/llvm.l lib/ex.l main.l -bye > base.ll
    +   $(PIL) lib/llvm.l -quiche lib/ex.l main.l -bye > base.ll
        mv base.map $(LIB)/map
    ```

3. Build main app:

    ```sh
    (cd src; make)
    ```
    
4. Build lib:

    ```sh
    (cd src; make so)
    ```

5. Move library to ldconfig path:

    ```sh
    sudo mv lib/picolisp.so /usr/lib/x86_64-linux-gnu/libpicolisp.so
    ```

6. Regenerate cache:

    ```sh
    sudo ldconfig
    ```

Then compile with:

```sh
cabal build -fpicolisp exe:simplex-chat 
```


Akahige:
It is better to use the minimal library instead of full PicoLisp, because it does not depend on POSIX.

So instead of pil21.tgz do

   wget https://software-lab.de/minlib.tgz
   tar xfz minlib.tgz

Quiche mode is enabled already in Makefile, so just

   (cd src; make)

is needed

The libraries libreadline-dev, libffi-dev and libssl-dev can also be omitted

PR #6244, opened 2025-09-05
ui: delay closing support view on acceptance for items to be marked read

spaced4ndy:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/SimpleXAPI.kt
```diff
@@ -2913,10 +2913,13 @@ object ChatController {
             && ModalManager.end.hasModalOpen(ModalViewId.SECONDARY_CHAT)
             && chatModel.secondaryChatsContext.value?.secondaryContextFilter is SecondaryContextFilter.GroupChatScopeContext
           ) {
-            withContext(Dispatchers.Main) {
-              chatModel.secondaryChatsContext.value = null
+            CoroutineScope(Dispatchers.Default).launch {
+              delay(1000L)
```
spaced4ndy, line 2917: increased to one second in kotlin because 0.65 was not enough for android when keyboard is open

PR #6299, opened 2025-09-24
core: improve link highlighting: DNS-compliant domains, hyphens, paths, and numeric characters

m0numental:
Changelog:
- Allow hyphens and numeric characters in domain names (example1-site.org)
- Support URLs with paths (example.com/path)
- Support multiple subdomain levels (api.v1.example.com)
- Expand domain name and subdomain length limits to DNS standard (1-63 characters)
- Added domain tests upon request


m0numental:

epoberezkin:
@m0numental thank you! Could you please add tests for the additionally supported cases to this file: https://github.com/simplex-chat/simplex-chat/blob/stable/tests/MarkdownTests.hs

epoberezkin:
Not sure I agree with completely dropping length limits - the alternative would be to embed a large list of existing TLDs...

m0numental:
> @m0numental thank you! Could you please add tests for the additionally supported cases to this file: https://github.com/simplex-chat/simplex-chat/blob/stable/tests/MarkdownTests.hs

Done 👍

> Not sure I agree with completely dropping length limits - the alternative would be to embed a large list of existing TLDs...

We could use this [list](https://data.iana.org/TLD/tlds-alpha-by-domain.txt) from IANA Root Zone Database. Should we hardcode it or fetch for each new release?

epoberezkin:
> We could use this [list](https://data.iana.org/TLD/tlds-alpha-by-domain.txt) from IANA Root Zone Database. Should we hardcode it or fetch for each new release?

I don't like this idea tbh. I'd rather we just keep the current heuristics.

m0numental:
> > We could use this [list](https://data.iana.org/TLD/tlds-alpha-by-domain.txt) from IANA Root Zone Database. Should we hardcode it or fetch for each new release?
> 
> I don't like this idea tbh. I'd rather we just keep the current heuristics.

Sorry, I miscommunicated in my initial changelog. What I meant was that I _replaced_ the original subdomain limits (3-4 chars) with DNS-compliant ones (1-63 chars), not completely removed them. I've updated the clangelog to reflect all my modifications. Please let me know if anything else needs adjustments.

epoberezkin:
> What I meant was that I replaced the original subdomain limits (3-4 chars) with DNS-compliant ones (1-63 chars), not completely removed them.

The problem is that it creates lots of false positives. So the current approach is keep this limit low, to restrict to the most commonly used TLDs, rather than to theoretic limits in the spec. Whoever needs links in long TLDs can send it as markdown syntax for hyperlinks or as https links.

PR #6310, opened 2025-09-26
To clarify how to connect to a public group, I misunderstood and trie…

nrbray:
To clarify how to connect to a public group, I misunderstood and tried to use /join which is for a contact to accept a group invitation sent directly to them.


nrbray:

PR #6333, opened 2025-10-03
android: fix showing connection alert when app is opened via link

spaced4ndy:


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/App.kt
```diff
@@ -243,9 +243,9 @@ fun MainScreen() {
       ModalManager.fullscreen.showOneTimePasscodeInView()
       AlertManager.privacySensitive.showInView()
       if (onboarding == OnboardingStage.OnboardingComplete) {
-        LaunchedEffect(chatModel.currentUser.value, chatModel.appOpenUrl.value) {
+        LaunchedEffect(chatModel.chatRunning.value, chatModel.currentUser.value, chatModel.appOpenUrl.value) {
           val (rhId, url) = chatModel.appOpenUrl.value ?: (null to null)
-          if (url != null) {
+          if (url != null && chatModel.chatRunning.value == true) {
```
spaced4ndy, line 248: fixes "chat not started" error when opening app via link with existing profile

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chatlist/ChatListView.kt
```diff
@@ -670,7 +671,9 @@ private fun ChatListSearchBar(listState: LazyListState, searchText: MutableState
                 // if some other text is pasted, enter search mode
                 focusRequester.requestFocus()
               } else {
-                connectProgressManager.cancelConnectProgress()
+                if (!chatModel.appOpenUrlConnecting.value) {
```
spaced4ndy, line 674: fixes showing connection alert when opening app via link before profile exists (onboarding). Here problem was that search bar renders after appOpenUrl started processing and its [search bar's] effect cancelled progress

PR #6396, opened 2025-10-22
ios: attempt to fix app freezing when keyboard opens on iOS 26.1

epoberezkin:


epoberezkin:
fixed with Info.plist option

PR #6444, opened 2025-11-17
docs: acknowledgements for async command responses rfc

spaced4ndy:


docs/rfcs/2025-11-17-async-commands-acks.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+Type is `AEvent`, requires encoding for To and FromField instances.
+
+Application of chat continuations is very limited, so not all events need to be saved. In fact currently we only need 2 types of events to be recorded - see below. This breaks separation between chat and agent (agent knows which events to record), however that abstraction has long been violated. This can be a contract between agent and chat - which events to keep and acknowledge.
+
+TBC separate type for storing only necessary events:
+
+```haskell
+data AEventDB where
```
epoberezkin, line 26: even if we want to persist some commands, no need to define type - it can be just a Bool predicate.

docs/rfcs/2025-11-17-async-commands-acks.md
```diff
@@ -0,0 +1,74 @@
...
+## Solution
+
+An idea is to persist events in agent until chat acknowledges their processing, and replay them on next start of command processing.
+
+### Agent persistence
+
+Save response on command before notifying chat (event received by chat via subQ).
+
+```sql
+ALTER TABLE commands ADD COLUMN response BLOB;
```
epoberezkin:
```suggestion
ALTER TABLE commands ADD COLUMN event BLOB;
```

epoberezkin: and most likely we need one more field for status? or how it's tracked

spaced4ndy: there doesn't seem much to track, once event is acknowledged command is deleted. so it's no event -> command has event saved -> command deleted

docs/rfcs/2025-11-17-async-commands-acks.md
spaced4ndy: this is just to consider if we want to do less encodings

PR #6445, opened 2025-11-17
core: member relations vector migration; set index in group for members; rework logic for avoiding duplicate introductions

spaced4ndy:


src/Simplex/Chat/Store/SQLite/Migrations/M20251117_group_forward_bitvector.hs
```diff
@@ -0,0 +1,92 @@
...
+--   - possibly by tracking con_ts (add new field) - save con_ts on member's CON in same transaction as
+--      retrieving list of members to introduce to; if member has con_ts set, skip introduction to them
+m20251117_group_forward_bitvector :: Query
+m20251117_group_forward_bitvector =
+  [sql|
+ALTER TABLE group_members ADD COLUMN sequential_id INTEGER NOT NULL DEFAULT 0;
+
+ALTER TABLE groups ADD COLUMN last_member_sequential_id INTEGER NOT NULL DEFAULT 0;
+
+ALTER TABLE group_members ADD COLUMN forward_bitvector BLOB;
```
epoberezkin:
```suggestion
ALTER TABLE group_members ADD COLUMN all_member_statuses BLOB;
```

epoberezkin: or it can be member_link_statuses + comment "2 bits per member"

PR #6464, opened 2025-11-25
core: forward based on relations vector

spaced4ndy:


bots/api/TYPES.md
```diff
@@ -2203,6 +2203,7 @@ Known:
 - memberRole: [GroupMemberRole](#groupmemberrole)
 - memberCategory: [GroupMemberCategory](#groupmembercategory)
 - memberStatus: [GroupMemberStatus](#groupmemberstatus)
+- relationsVector: string?
```
epoberezkin: why do you need it in UI? it'll only waste memory I think

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -167,6 +168,9 @@ startChatController mainApp enableSndFiles = do
   runExceptT (syncConnections' users) >>= \case
     Left e -> liftIO $ putStrLn $ "Error synchronizing connections: " <> show e
     Right _ -> pure ()
+  runExceptT migrateMemberRelations >>= \case
```
epoberezkin, line 170: I don't understand - are they all migrated?

epoberezkin: if it starts background process, I'd check first if migration is necessary before starting it.

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -4163,6 +4171,39 @@ agentSubscriber = do
...
+  liftIO $ threadDelay' 5000000 -- 5 seconds (initial delay)
+  migrateUserNotAdminGroups
+  migrateMembers
+  where
+    stepDelay = 100000 -- 0.1 second
+    migrateUserNotAdminGroups = flip catchAllErrors eToView $ do
+      lift waitChatStartedAndActivated
+      gIds <- withStore' getUserNotAdminGroupIds
+      forM_ gIds $ \gId -> do
+        withStore' (\db -> setEmptyVectorForGroupId db gId) `catchAllErrors` eToView
```
epoberezkin: you don't need to read them, can be just single update query.

epoberezkin: also, this can be done as part of schema migration, so it doesn't need to be code.

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -4163,6 +4171,39 @@ agentSubscriber = do
...
+        liftIO $ threadDelay' stepDelay
+    migrateMembers = flip catchAllErrors eToView $ do
+      lift waitChatStartedAndActivated
```
epoberezkin, line 4180: it already waited before, why twice?

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -4163,6 +4171,39 @@ agentSubscriber = do
...
+      unless (null gmIds) migrateMembers
+      where
+        processMember gmId =
+          withGroupMemberLock "runRelationsVectorMigration" gmId $
+            withStore' (\db -> getMemberRelationsVector_' db gmId) >>= \case
+              Just _ -> pure () -- already has vector
+              Nothing -> do
+                introducedRelations <- withStore' $ \db -> getIntroducedRelations db gmId
+                introducedToRelations <- withStore' $ \db -> getIntroducedToRelations db gmId
+                connectedRelations <- withStore' $ \db -> getIntroConnectedRelations db gmId
```
epoberezkin: why these are three queries and not one?

epoberezkin: I would try to pack it all in a single query avoiding the need to read members completely, as the change is not transactional.

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -1024,65 +1029,134 @@ introduceToModerators :: VersionRangeChat -> User -> GroupInfo -> GroupMember ->
...
   modMs <- withStore' $ \db -> getGroupModerators db vr user gInfo
-  let rcpModMs = filter (\mem -> memberCurrent mem && maxVersion (memberChatVRange mem) >= groupKnockingVersion) modMs
+  let rcpModMs = filter introduceMemP modMs
   introduceMember vr user gInfo m rcpModMs (Just $ MSMember $ memberId' m)
+  where
+    introduceMemP :: GroupMember -> Bool
+    introduceMemP mem =
+      memberCurrent mem
+        && groupMemberId' mem /= groupMemberId' m
+        && maxVersion (memberChatVRange mem) >= groupKnockingVersion
```
epoberezkin: could just add condition, why take it out?

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -1024,65 +1029,134 @@ introduceToModerators :: VersionRangeChat -> User -> GroupInfo -> GroupMember ->
...
   forM_ (memberConn m) $ \mConn -> do
     let msg =
-          if (maxVersion (memberChatVRange m) >= groupKnockingVersion)
+          if maxVersion (memberChatVRange m) >= groupKnockingVersion
             then XGrpLinkAcpt GAPendingReview memberRole memberId
             else XMsgNew $ MCSimple $ extMsgContent (MCText pendingReviewMessage) Nothing
     void $ sendDirectMemberMessage mConn msg groupId
   modMs <- withStore' $ \db -> getGroupModerators db vr user gInfo
-  let rcpModMs = filter (\mem -> memberCurrent mem && maxVersion (memberChatVRange mem) >= groupKnockingVersion) modMs
+  let rcpModMs = filter introduceMemP modMs
```
```suggestion
  let rcpModMs = filter (\mem -> memberCurrent mem && groupMemberId' mem /= groupMemberId' m && maxVersion (memberChatVRange mem) >= groupKnockingVersion) modMs
```

src/Simplex/Chat/Library/Internal.hs
epoberezkin, line 1040:
```suggestion
```

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -1024,65 +1029,134 @@ introduceToModerators :: VersionRangeChat -> User -> GroupInfo -> GroupMember ->
...
 introduceToAll :: VersionRangeChat -> User -> GroupInfo -> GroupMember -> CM ()
-introduceToAll vr user gInfo m = do
-  members <- withStore' $ \db -> getGroupMembersForIntroduction db vr user gInfo m
-  let recipients = filter memberCurrent members
+introduceToAll vr user gInfo m@GroupMember {relationsVector} = do
+  members <- withStore' $ \db -> getGroupMembers db vr user gInfo
+  let recipients = filter introduceMemP members
   introduceMember vr user gInfo m recipients Nothing
+  where
+    introduceMemP :: GroupMember -> Bool
```
epoberezkin: we normally use such names for parsers, why P is here?

epoberezkin: do not change naming conventions, local predicates should be adjectives, without subject it is applied to.

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -1024,65 +1029,134 @@ introduceToModerators :: VersionRangeChat -> User -> GroupInfo -> GroupMember ->
...
+    introduceMemP mem = case relationsVector of
+      Nothing ->
+        memberCurrent mem && groupMemberId' mem /= groupMemberId' m
+      Just vec ->
+        let introducedMemsIdxs = getMemberRelationsIndexes (\r -> r == MRIntroduced || r == MRIntroducedTo || r == MRConnected) vec
```
epoberezkin: this should be made into a set, but it may be better to do direct lookup - I don't understand wha't the benefit of making it a list first, when direct lookup is more efficient. Why is array of indices needed here? and not lookup each member index directly while filtering? Seems unnecessarily complex and indirect.

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -2715,32 +2716,47 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+          vec <- withStore $ \db -> getMemberRelationsVector db sendingMem
+          let vec' = setMemberRelation (indexInGroup refMem) MRConnected vec
+          withStore' $ \db -> updateMemberRelationsVector db sendingMem vec'
+      when (isJust refMemVec) $
+        withGroupMemberLock "xGrpMemCon, referenced member" (groupMemberId' refMem) $ do
+          vec <- withStore $ \db -> getMemberRelationsVector db refMem
+          let vec' = setMemberRelation (indexInGroup sendingMem) MRConnected vec
+          withStore' $ \db -> updateMemberRelationsVector db refMem vec'
+      when (isNothing sendingMemVec || isNothing refMemVec) $
+        updateIntroductionRecord sendingMem refMem
```
epoberezkin: unclear - what does it do?

src/Simplex/Chat/Store/Groups.hs
```diff
@@ -1719,6 +1800,58 @@ getForwardScopeMember db vr user GroupMember {groupMemberId = sendingGMId} scope
...
+        SELECT m.index_in_group
+        FROM group_member_intros i
+        JOIN group_members m ON m.group_member_id = i.re_group_member_id
+        WHERE i.to_group_member_id = ? AND i.intro_status NOT IN (?,?,?)
+      |]
+      (gmId, GMIntroReConnected, GMIntroToConnected, GMIntroConnected)
+
+getIntroducedToRelations :: DB.Connection -> GroupMemberId -> IO [(Int64, MemberRelation)]
+getIntroducedToRelations db gmId =
+  map ((,MRIntroducedTo) . fromOnly) <$>
```
epoberezkin: if you read it separately anyway, then why map here?

src/Simplex/Chat/Store/Groups.hs
```diff
@@ -1719,6 +1800,58 @@ getForwardScopeMember db vr user GroupMember {groupMemberId = sendingGMId} scope
...
+    map ((,MRConnected) . fromOnly) <$>
+      DB.query
+        db
+        [sql|
+          SELECT m.index_in_group
+          FROM group_member_intros i
+          JOIN group_members m ON m.group_member_id = i.to_group_member_id
+          WHERE i.re_group_member_id = ? AND i.intro_status IN (?,?,?)
+        |]
+        (gmId, GMIntroReConnected, GMIntroToConnected, GMIntroConnected)
```
epoberezkin: why do they need to be two queries and not one query with OR in join?

JOIN group_members m ON m.group_member_id = i.re_group_member_id OR m.group_member_id = i.to_group_member_id

epoberezkin: and why map with status

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -2715,32 +2716,47 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
-            Right intro -> inviteeXGrpMemCon intro
-            Left _ ->
+    xGrpMemCon gInfo sendingMem@GroupMember {relationsVector = sendingMemVec} memId = do
+      refMem@GroupMember {relationsVector = refMemVec} <-
+        withStore $ \db -> getGroupMemberByMemberId db vr user gInfo memId
+      when (isJust sendingMemVec) $
+        withGroupMemberLock "xGrpMemCon, sending member" (groupMemberId' sendingMem) $ do
+          vec <- withStore $ \db -> getMemberRelationsVector db sendingMem
+          let vec' = setMemberRelation (indexInGroup refMem) MRConnected vec
+          withStore' $ \db -> updateMemberRelationsVector db sendingMem vec'
```
epoberezkin: this should be a function, and most likely it can be moved inside sql

src/Simplex/Chat/Library/Commands.hs
epoberezkin: and we will need to make this migration anyway, so maybe we should start from it, and here use the same function just for one member.

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -1024,65 +1029,134 @@ introduceToModerators :: VersionRangeChat -> User -> GroupInfo -> GroupMember ->
...
+            )
+            members
+      Just vec ->
+        let introducedMemsIdxs = getMemberRelationsIndexes (\r -> r == MRIntroduced || r == MRIntroducedTo || r == MRConnected) vec
+         in pure $
+              filter
+                ( \mem ->
+                    memberCurrent mem
+                      && groupMemberId' mem /= groupMemberId' m
+                      && indexInGroup mem `notElem` introducedMemsIdxs
```
epoberezkin: same predicate as above (although it won't be if you lookup while filtering)

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -1024,65 +1029,134 @@ introduceToModerators :: VersionRangeChat -> User -> GroupInfo -> GroupMember ->
...
+getMemberRelationsIndexes p (MemberRelationsVector v) = getRelationsIndexes p v
+{-# INLINE getMemberRelationsIndexes #-}
+
+setMemberRelation :: Int64 -> MemberRelation -> MemberRelationsVector -> MemberRelationsVector
+setMemberRelation i r (MemberRelationsVector v) = MemberRelationsVector $ setRelation i r v
+{-# INLINE setMemberRelation #-}
+
+setMemberRelations :: [(Int64, MemberRelation)] -> MemberRelationsVector -> MemberRelationsVector
+setMemberRelations relations (MemberRelationsVector v) = MemberRelationsVector $ setRelations relations v
+{-# INLINE setMemberRelations #-}
```
epoberezkin: Seems unnecessary (same as newtype). If this is to encode it into JSON, I would probably just drop it from JSON entirely (via aeson options), it is not needed outside of Haskell. Also worth revising if this relation is used in all cases that load members, and if most places don't use it (which seems likely), then this should not be part of member record at all.

src/Simplex/Chat/Library/Commands.hs
spaced4ndy: it's a loop

src/Simplex/Chat/Library/Commands.hs
spaced4ndy: it's inside member lock, logic mimicks forwarding logic

src/Simplex/Chat/Library/Internal.hs
spaced4ndy: too long, doesn't fit into screen

src/Simplex/Chat/Library/Internal.hs
spaced4ndy: this is to clearer see difference between introduce functions

src/Simplex/Chat/Library/Internal.hs
spaced4ndy: it was already named like this in some functions. also p as in predicate

src/Simplex/Chat/Library/Subscriber.hs
spaced4ndy: updates member vectors on x.grp.mem.con

src/Simplex/Chat/Store/Groups.hs
spaced4ndy: to avoid additional maps later

src/Simplex/Chat/Store/Groups.hs
spaced4ndy: 1. because filter is also different
2. to avoid additional maps later

src/Simplex/Chat/Store/SQLite/Migrations/M20251117_member_relations_vector.hs
```diff
@@ -46,6 +46,18 @@ SET member_index = COALESCE((
...
 ), 0);
+
+UPDATE group_members
+SET member_relations_vector = x''
+WHERE group_id IN (
+  SELECT mu.group_id
+  FROM group_members mu
+  WHERE mu.member_category = 'user'
+    AND (
+      mu.member_role NOT IN (CAST('admin' AS BLOB), CAST('owner' AS BLOB))
```
spaced4ndy, line 125: member_role is written as BLOB (even though it's text)

simplex-chat.cabal
```diff
@@ -123,6 +123,7 @@ library
           Simplex.Chat.Store.Postgres.Migrations.M20251007_connections_sync
           Simplex.Chat.Store.Postgres.Migrations.M20251017_chat_tags_cascade
           Simplex.Chat.Store.Postgres.Migrations.M20251117_member_relations_vector
+          Simplex.Chat.Store.Postgres.Migrations.M20251128_member_relations_vector_stage_2
```
spaced4ndy, line 126: can remove it from here too (it's commented in list of migrations)

epoberezkin: ok

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -1024,65 +1025,108 @@ introduceToModerators :: VersionRangeChat -> User -> GroupInfo -> GroupMember ->
...
-        isAdmin GroupMemberIntro {reMember = GroupMember {memberRole}} = memberRole >= GRAdmin
-        hasPicture GroupMemberIntro {reMember = GroupMember {memberProfile = LocalProfile {image}}} = isJust image
-    processIntro intro@GroupMemberIntro {introId} = do
-      void $ sendDirectMemberMessage conn (memberIntro $ reMember intro) groupId
-      withStore' $ \db -> updateIntroStatus db introId GMIntroSent
+        isAdmin GroupMember {memberRole} = memberRole >= GRAdmin
+        hasPicture GroupMember {memberProfile = LocalProfile {image}} = isJust image
+
+wasIntroduced :: MemberRelation -> Bool
+wasIntroduced = \case
```
epoberezkin: unnecessary indirection, it's just (/= MRNew), will remove

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -3255,42 +3224,27 @@ runDeliveryJobWorker a deliveryKey Worker {doWork} = do
...
-                                if memberCategory sender == GCInviteeMember
-                                  then withStore' $ \db -> getForwardIntroducedMembers db vr user sender highlyAvailable
-                                  else pure []
-                              -- invited members to which this member was introduced
-                              invitedMembers <- withStore' $ \db -> getForwardInvitedMembers db vr user sender highlyAvailable
-                              pure $ introducedMembers <> invitedMembers
+                        buildMemberList sender = do
+                          vec <- withStore $ \db -> migrateGetMemberRelationsVector db sender
+                          let introducedMemsIdxs = getRelationsIndexes (== MRIntroduced) vec
+                          ms <- withStore' $ \db -> getGroupMembersByIndexes db vr user gInfo introducedMemsIdxs
```
epoberezkin: it loads all introduced members, but for support scope it should only load introduced moderators.

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -3255,42 +3224,27 @@ runDeliveryJobWorker a deliveryKey Worker {doWork} = do
...
+                          pure $ case jobScope of
+                            DJSGroup {jobSpec}
+                              | jobSpecImpliedPending jobSpec -> filter memberCurrentOrPending ms
+                              | otherwise -> filter memberCurrent ms
+                            DJSMemberSupport scopeGMId -> filter shouldForwardTo ms
+                              where
+                                shouldForwardTo m =
+                                  let mId = groupMemberId' m
+                                   in mId /= senderId && (mId == scopeGMId || currentModerator m)
```
epoberezkin: I kept here the logic to never forward back to sender, previously it did it for moderators.

src/Simplex/Chat/Store/Groups.hs
```diff
@@ -994,6 +1005,14 @@ getGroupMemberById db vr user@User {userId} groupMemberId =
       (groupMemberQuery <> " WHERE m.group_member_id = ? AND m.user_id = ?")
       (groupMemberId, userId)
 
+getGroupMemberByIndex :: DB.Connection -> VersionRangeChat -> User -> GroupInfo -> Int64 -> ExceptT StoreError IO GroupMember
```
epoberezkin, line 1009: we also need getGroupModeratorsByIndex

src/Simplex/Chat/Types/MemberRelations.hs
```diff
@@ -20,55 +30,134 @@ import Foreign.Marshal.Utils (copyBytes, fillBytes)
...
+  | i >= 0 = setNewRelations [(i, dir, r)] v
+  | otherwise = v
+
+-- | Set multiple new relations with both direction and status at once.
+-- Expands the vector lazily if needed.
+setNewRelations :: [(Int64, IntroductionDirection, MemberRelation)] -> ByteString -> ByteString
+setNewRelations relations v = setRelations' [(i, Just dir, r) | (i, dir, r) <- relations] v
+
+-- | Internal function to set relations with optional direction update.
+-- If direction is Nothing, preserves existing direction; if Just, sets new direction.
```
epoberezkin: this is strange

src/Simplex/Chat/Types/MemberRelations.hs
```diff
@@ -20,55 +30,129 @@ import Foreign.Marshal.Utils (copyBytes, fillBytes)
...
+-- Returns (IDSubjectIntroduced, MRNew) if the vector is not long enough (lazy initialization).
+getRelation' :: Int64 -> ByteString -> (IntroductionDirection, MemberRelation)
+getRelation' i v
+  | i < 0 || fromIntegral i >= B.length v = (IDSubjectIntroduced, MRNew)
+  | otherwise =
+      let b = v `B.index` fromIntegral i
+       in (fromIntroductionInt $ (b .&. directionMask) `shiftR` 3, fromRelationInt $ b .&. statusMask)
 
+-- | Get the indexes of members that satisfy the given relation predicate.
+getRelationsIndexes :: (MemberRelation -> Bool) -> ByteString -> [Int64]
```
epoberezkin: unnecessary generalization, we only use it to get MRIntroduced relations, so can be done simpler in-place

src/Simplex/Chat/Store/Postgres/Migrations/M20251117_member_relations_vector.hs
```diff
@@ -6,10 +6,85 @@ import Data.Text (Text)
...
+RETURNS BYTEA AS $$
+DECLARE
+  new_len INT;
+  result BYTEA;
+  byte_val INT;
+BEGIN
+  IF idx < 0 THEN
+    RETURN v;
+  END IF;
+  byte_val := (direction * 8) + status;
```
epoberezkin: mask?

src/Simplex/Chat/Library/Subscriber.hs
spaced4ndy: it didn't because sender is never loaded by vector

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -3255,42 +3224,33 @@ runDeliveryJobWorker a deliveryKey Worker {doWork} = do
...
-                              pure $ introducedMembers <> invitedMembers
+                            vec <- withStore $ \db -> migrateGetMemberRelationsVector db sender
+                            let introducedMemsIdxs = getRelationsIndexes MRIntroduced vec
+                            ms <- withStore' $ \db -> getSupportScopeMembersByIndexes db vr user gInfo scopeGMId introducedMemsIdxs
+                            pure $ filter shouldForwardTo ms
+                            where
+                              shouldForwardTo m =
+                                let mId = groupMemberId' m
+                                  in mId /= senderId && (mId == scopeGMId || currentModerator m)
+                              senderId = groupMemberId' sender
```
epoberezkin:
```suggestion
                              shouldForwardTo m = groupMemberId' m == scopeGMId || currentModerator m
```

PR #6472, opened 2025-11-28
core: relations vector live migrations; stage 2 migration sql

spaced4ndy:


src/Simplex/Chat.hs
```diff
@@ -118,8 +118,8 @@ logCfg = LogConfig {lc_file = Nothing, lc_stderr = True}
 
 createChatDatabase :: ChatDbOpts -> MigrationConfig -> IO (Either MigrationError ChatDatabase)
 createChatDatabase chatDbOpts migrationConfig = runExceptT $ do
-  chatStore <- ExceptT $ createChatStore (toDBOpts chatDbOpts chatSuffix False) migrationConfig
-  agentStore <- ExceptT $ createAgentStore (toDBOpts chatDbOpts agentSuffix False) migrationConfig
+  chatStore <- ExceptT $ createChatStore (toDBOpts chatDbOpts chatSuffix False chatDBFunctions) migrationConfig
```
epoberezkin, line 121: no need to pass it from here, just change toDBOpts

PR #6484, opened 2025-12-02
core: preserve detailed information in relation vectors

spaced4ndy:


src/Simplex/Chat/Store/Groups.hs
```diff
@@ -1814,7 +1814,7 @@ getIntroducedRelations db gmId =
 
 getIntroducedToRelations :: DB.Connection -> GroupMemberId -> IO [(Int64, MemberRelation)]
 getIntroducedToRelations db gmId =
-  map ((,MRIntroducedTo) . fromOnly) <$>
+  map ((,MRIntroduced) . fromOnly) <$>
```
spaced4ndy, line 1817: for migration - to be reworked in sql

PR #6490, opened 2025-12-08
core: finalize introductions -> member relations vector stage 2 migration (drop introductions)

spaced4ndy:


src/Simplex/Chat/Store/SQLite/Migrations/M20251128_member_relations_vector_stage_2.hs
```diff
@@ -32,11 +31,31 @@ SET member_relations_vector = (
   )
 )
 WHERE member_relations_vector IS NULL;
+
+DROP INDEX idx_group_member_intros_to_group_member_id;
+DROP INDEX idx_group_member_intros_re_group_member_id;
+DROP INDEX idx_pending_group_messages_group_member_intro_id;
+ALTER TABLE pending_group_messages DROP COLUMN group_member_intro_id;
+DROP TABLE group_member_intros;
```
epoberezkin:
```suggestion
```

src/Simplex/Chat/Store/SQLite/Migrations/M20251128_member_relations_vector_stage_2.hs
```diff
@@ -32,11 +31,31 @@ SET member_relations_vector = (
...
 |]
 
--- TODO [relations vector] re-create group_member_intros
 down_m20251128_member_relations_vector_stage_2 :: Query
```
epoberezkin: down migration can be removed completely in this case

PR #6516, opened 2025-12-19
List available bot SDKs in bot API documentation

a1akris:


a1akris:

UnkwUsr:
Worth adding Python library: https://github.com/FailSpy/simpx-py

And maybe one more for Rust: https://github.com/rdbo/simplex-chat-rs (though it seems not maintained and no mechanism for auto generating bindings like in simploxide)

a1akris:
> And maybe one more for Rust: https://github.com/rdbo/simplex-chat-rs

I'm aware of this library and I was planning to use it and contribute to it before developing my own bindings but after a quick code review I found that their client implementation is [very](https://github.com/rdbo/simplex-chat-rs/blob/5d74ec2305296f1f4e1546e69b1957adb13a8793/simplex-chat/src/lib.rs#L35C26-L35C31)(I mean [VERY](https://github.com/rdbo/simplex-chat-rs/blob/5d74ec2305296f1f4e1546e69b1957adb13a8793/simplex-chat/src/lib.rs#L170)) inefficient and it would be easier to write everything from scratch. That's how `simploxide` was born))

> Worth adding Python library: https://github.com/FailSpy/simpx-py

Unfortunately, it seems like this library is abandoned too. There are some community forks, though, so if somebody confirms that they maintain a fork I will include the fork.

PR #6562, opened 2026-01-09
scripts/desktop: tweaks to be more portable/forgiving

alexfmpe:
`/bin/bash` doesn't exist on some systems. I ran into this issue on my NixOS Linux. Would also happen on BSDs, but don't think that's a concern. AFAICT the windows/macos scripts don't need this change.

### DIFF
--- a/scripts/desktop/build-lib-mac.sh
+++ b/scripts/desktop/build-lib-mac.sh
@@ -3,7 +3,7 @@
 set -e
 
 OS=mac
-ARCH="$(uname -m)"
+ARCH="${1:-$(uname -m)}"
 COMPOSE_ARCH=$ARCH
 GHC_VERSION=9.6.3
 DATABASE_BACKEND="${2:-sqlite}"
--- a/scripts/desktop/prepare-vlc-mac.sh
+++ b/scripts/desktop/prepare-vlc-mac.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-ARCH="$(uname -m)"
+ARCH="${1:-$(uname -m)}" 
 if [ "$ARCH" == "arm64" ]; then
     ARCH=aarch64
     vlc_arch=arm64


scripts/desktop/build-lib-mac.sh
```diff
@@ -3,7 +3,7 @@
 set -e
 
 OS=mac
-ARCH="${1:-`uname -a | rev | cut -d' ' -f1 | rev`}"
+ARCH="$(uname -m)"
```
alexfmpe: I have no idea why, but on my macos in particular, running uname inside a nix-shell gives extra output at the end which breaks the rev/cut logic.


```
$ uname -a
Darwin MacBookPro.Home 24.6.0 Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:29 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6000 arm64
$ nix-shell -p hello --run 'uname -a'
Darwin MacBookPro.Home 24.6.0 Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:29 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6000 arm64 arm Darwin
```

scripts/desktop/build-lib-linux.sh
```diff
@@ -8,7 +8,7 @@ function readlink() {
 
 OS=linux
 ARCH="$(uname -m)"
-GHC_VERSION=9.6.3
+GHC_VERSION="$(ghc --version | rev | cut -d ' ' -f 1 | rev)"
```
alexfmpe: At one point I was using 9.6.4, which caused a broken path later on.

epoberezkin: we shouldn't use different GHC version for sure

epoberezkin:
```suggestion
GHC_VERSION=9.6.3
```

scripts/desktop/build-lib-mac.sh
epoberezkin: this is also suspicious and may have unintended effects

scripts/desktop/build-lib-mac.sh
```diff
@@ -3,9 +3,9 @@
...
 COMPOSE_ARCH=$ARCH
-GHC_VERSION=9.6.3
+GHC_VERSION="$(ghc --version | rev | cut -d ' ' -f 1 | rev)"
```
epoberezkin:
```suggestion
GHC_VERSION=9.6.3
```

scripts/desktop/build-lib-mac.sh
epoberezkin: at a minimum, it has to be tested on all platforms, by creating a fork and triggering release builds to see that it succeeds building all libraries in production

epoberezkin: or just needs to be removed as it's not what we agreed

epoberezkin:
```suggestion
ARCH="${1:-$(uname -m)}"
```

scripts/desktop/prepare-vlc-mac.sh
```diff
@@ -2,7 +2,7 @@
 
 set -e
 
-ARCH="${1:-`uname -a | rev | cut -d' ' -f1 | rev`}"
+ARCH="$(uname -m)"
```
epoberezkin:
```suggestion
ARCH="${1:-$(uname -m)}" 
```

PR #6596, opened 2026-01-23
core: optional member in group rcv direction for messages from channel

spaced4ndy:


src/Simplex/Chat/Messages/Batch.hs
```diff
@@ -71,12 +71,13 @@ batchDeliveryTasks1 vr maxLen = toResult . foldl' addToBatch ([], [], [], 0, 0)
...
       | otherwise = (msgBodies, taskIds, largeTaskIds, len, n)
       where
-        MessageDeliveryTask {taskId, senderMemberId, senderMemberName, brokerTs, chatMessage, messageFromChannel = _messageFromChannel} = task
-        -- TODO [channels fwd] handle messageFromChannel (null memberId in XGrpMsgForward)
+        MessageDeliveryTask {taskId, taskContext = DeliveryTaskContext {sentAsGroup}, senderMemberId, senderMemberName, brokerTs, chatMessage} = task
         msgBody =
-          let fwdEvt = XGrpMsgForward senderMemberId (Just senderMemberName) chatMessage brokerTs
+          let memberId_ = if sentAsGroup then Nothing else Just senderMemberId
+              memberName_ = if sentAsGroup then Nothing else Just senderMemberName
+              fwdEvt = XGrpMsgForward memberId_ memberName_ chatMessage brokerTs
```
epoberezkin, line 78: why do you need parallel ifs here and not just one if?

epoberezkin: also it's strange that it happens in batching, logically this decision/knowledge belongs outside - batching should not be concerned with business decisions.

epoberezkin: this decision should be made and recorded at a point task is created, not at a point of batching. This is just bad design.

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -2913,40 +2913,36 @@ markReceivedGroupReportsDeleted db User {userId} GroupInfo {groupId, membership}
...
+getGroupCIBySharedMsgId' db user g@GroupInfo {groupId} sharedMsgId = do
   itemId <-
     ExceptT . firstRow fromOnly (SEChatItemSharedMsgIdNotFound sharedMsgId) $
       DB.query
         db
         [sql|
           SELECT chat_item_id
           FROM chat_items
-          WHERE user_id = ? AND group_id = ? AND group_member_id = ? AND shared_msg_id = ?
+          WHERE group_id = ? AND shared_msg_id = ?
```
epoberezkin, line 2925: this is incorrect, as it may get multiple items - shared message ID is unique within member, not globally. Why is it needed?

epoberezkin: also it means that both sender of asgroup member must ensure global uniqueness and relay should check it?

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -2913,40 +2913,36 @@ markReceivedGroupReportsDeleted db User {userId} GroupInfo {groupId, membership}
...
+  case cci of
+    CChatItem SMDRcv ChatItem {chatDir = CIGroupRcv itemMember_}
+      | matchesMemberId itemMember_ -> pure cci
+      | otherwise -> throwError $ SEChatItemSharedMsgIdNotFound sharedMsgId
+    CChatItem SMDSnd _ -> pure cci
+  where
+    matchesMemberId :: Maybe GroupMember -> Bool
+    matchesMemberId itemMember_ = case (memberId_, itemMember_) of
+      (Just mId, Just GroupMember {memberId = mId'}) -> mId == mId'
+      (Nothing, Nothing) -> True -- channel message
```
epoberezkin, line 2944: even if it's the right approach - not sure about it - why not do it in query? This per-member shared message ID raises the question whehter it's even correct approach to be honest.

src/Simplex/Chat/Store/Messages.hs
epoberezkin: why user_id removed and how it affects the indices?

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -2204,18 +2208,19 @@ saveGroupRcvMsg user groupId authorMember conn@Connection {connId} agentMsgMeta
...
+      refAuthorId = groupMemberId' <$> refAuthorMember_
+  -- TODO [relays] TBC highlighting difference between deduplicated messages (useRelays branch)
   withStore' (\db -> runExceptT $ createNewRcvMessage db (GroupId groupId) newMsg sharedMsgId_ refAuthorId fwdMemberId) >>= \case
     Right msg -> pure $ Just msg
     Left e@SEDuplicateGroupMessage {authorGroupMemberId, forwardedByGroupMemberId}
       | useRelays' gInfo -> pure Nothing -- with chat relays, duplicates are expected
-      | otherwise -> case (authorGroupMemberId, forwardedByGroupMemberId) of
-          (Just authorGMId, Nothing) -> do
+      | otherwise -> case (authorGroupMemberId, forwardedByGroupMemberId, refAuthorMember_) of
+          (Just authorGMId, Nothing, Just GroupMember {memberId = refMemberId}) -> do
```
epoberezkin, line 2223: why channel messages where refAuthorMember_ is Nothing are excluded here?

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -1412,22 +1416,30 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
       Nothing
         | memberRole > GRObserver || memberPending m -> a
         | otherwise -> messageError "member is not allowed to send messages" $> Nothing
 
+    memberCanSend' ::
+      Maybe GroupMember ->
+      Maybe MsgScope ->
+      CM (Maybe DeliveryTaskContext) ->
+      CM (Maybe DeliveryTaskContext)
+    memberCanSend' Nothing _ a = a -- forwarded message from channel, already allowed
```
epoberezkin, line 1440: this is strange.

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -632,68 +633,67 @@ insertChatItemMessage_ db ciId msgId ts = DB.execute db "INSERT INTO chat_item_m
...
-    getDirectChatItemQuote_ contactId userSent = do
-      fmap ciQuoteDirect . maybeFirstRow fromOnly $
+    getDirectChatItemQuote_ :: Int64 -> IO (CIQuote 'CTDirect)
+    getDirectChatItemQuote_ contactId = do
+      fmap ciQuoteDirect . maybeFirstRow id $
         DB.query
           db
-          "SELECT chat_item_id FROM chat_items WHERE user_id = ? AND contact_id = ? AND shared_msg_id = ? AND item_sent = ?"
-          (userId, contactId, msgId, BI userSent)
+          "SELECT chat_item_id, item_sent FROM chat_items WHERE user_id = ? AND contact_id = ? AND shared_msg_id = ?"
```
epoberezkin, line 646: why this filter moved from sql to code?

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -632,68 +633,67 @@ insertChatItemMessage_ db ciId msgId ts = DB.execute db "INSERT INTO chat_item_m
...
-            JOIN contact_profiles p ON p.contact_profile_id = COALESCE(m.member_profile_id, m.contact_profile_id)
-            LEFT JOIN contacts c ON m.contact_id = c.contact_id
-            LEFT JOIN chat_items i ON i.user_id = m.user_id
-                                      AND i.group_id = m.group_id
-                                      AND m.group_member_id = i.group_member_id
-                                      AND i.shared_msg_id = ?
-            WHERE m.user_id = ? AND m.group_id = ? AND m.member_id = ?
-          |]
-          (msgId, userId, groupId, mId)
+        <$> maybeFirstRow id
```
epoberezkin, line 660: why this change was made? it looks like support for quotes was not sent depends on it?

src/Simplex/Chat/Types.hs
```diff
@@ -491,9 +491,17 @@ data GroupInfo = GroupInfo
   }
   deriving (Eq, Show)
 
+groupId' :: GroupInfo -> GroupId
+groupId' GroupInfo {groupId} = groupId
+
 useRelays' :: GroupInfo -> Bool
 useRelays' GroupInfo {useRelays} = isTrue useRelays
 
+type SendAsGroup = Bool
```
epoberezkin, line 500: we already have ShowGroupAsSender, why another synonym?

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -2184,17 +2188,20 @@ processChatCommand vr nm = \case
     contactId <- withFastStore $ \db -> getContactIdByName db user fromContactName
     forwardedItemId <- withFastStore $ \db -> getDirectChatItemIdByText' db user contactId forwardedMsg
     toChatRef <- getChatRef user toChatName
-    processChatCommand vr nm $ APIForwardChatItems toChatRef (ChatRef CTDirect contactId Nothing) (forwardedItemId :| []) Nothing
+    asGroup <- shouldSendAsGroup user toChatRef
```
epoberezkin, line 2191: This reads the whole group only to determine if "send as group"

PR #6604, opened 2026-01-30
core: channel messages

epoberezkin:


src/Simplex/Chat/Library/Commands.hs
```diff
@@ -927,7 +927,7 @@ processChatCommand vr nm = \case
                 MCChat {} -> True
                 MCUnknown {} -> True
   -- TODO [knocking] forward from / to scope
-  APIForwardChatItems toChat@(ChatRef toCType toChatId toScope) fromChat@(ChatRef fromCType fromChatId _fromScope) itemIds itemTTL -> withUser $ \user -> case toCType of
+  APIForwardChatItems toChat@(ChatRef toCType toChatId toScope) fromChat@(ChatRef fromCType fromChatId _fromScope) itemIds itemTTL _sendAsGroup -> withUser $ \user -> case toCType of
```
spaced4ndy: why in forward API _sendAsGroup is ignored, and instead decision is made automatically via sendAsGroup'? sendAsGroup' should only be used in CLI APIs to mimic UI without additional parameters

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -2184,17 +2188,17 @@ processChatCommand vr nm = \case
     contactId <- withFastStore $ \db -> getContactIdByName db user fromContactName
     forwardedItemId <- withFastStore $ \db -> getDirectChatItemIdByText' db user contactId forwardedMsg
     toChatRef <- getChatRef user toChatName
-    processChatCommand vr nm $ APIForwardChatItems toChatRef (ChatRef CTDirect contactId Nothing) (forwardedItemId :| []) Nothing
+    processChatCommand vr nm $ APIForwardChatItems toChatRef (ChatRef CTDirect contactId Nothing) (forwardedItemId :| []) Nothing False
```
spaced4ndy: this is CLI API. shouldn't default to False, should pass flag based on sendAsGroup' - same as CLI Send APIs -> APIForwardChatItems should then correctly use flag, as in comment above

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -2184,17 +2188,17 @@ processChatCommand vr nm = \case
...
   ForwardGroupMessage toChatName fromGroupName fromMemberName_ forwardedMsg -> withUser $ \user -> do
     groupId <- withFastStore $ \db -> getGroupIdByName db user fromGroupName
     forwardedItemId <- withFastStore $ \db -> getGroupChatItemIdByText db user groupId fromMemberName_ forwardedMsg
     toChatRef <- getChatRef user toChatName
-    processChatCommand vr nm $ APIForwardChatItems toChatRef (ChatRef CTGroup groupId Nothing) (forwardedItemId :| []) Nothing
+    processChatCommand vr nm $ APIForwardChatItems toChatRef (ChatRef CTGroup groupId Nothing) (forwardedItemId :| []) Nothing False
```
spaced4ndy: same

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -2184,17 +2188,17 @@ processChatCommand vr nm = \case
...
   ForwardLocalMessage toChatName forwardedMsg -> withUser $ \user -> do
     folderId <- withFastStore (`getUserNoteFolderId` user)
     forwardedItemId <- withFastStore $ \db -> getLocalChatItemIdByText' db user folderId forwardedMsg
     toChatRef <- getChatRef user toChatName
-    processChatCommand vr nm $ APIForwardChatItems toChatRef (ChatRef CTLocal folderId Nothing) (forwardedItemId :| []) Nothing
+    processChatCommand vr nm $ APIForwardChatItems toChatRef (ChatRef CTLocal folderId Nothing) (forwardedItemId :| []) Nothing False
```
spaced4ndy: same

src/Simplex/Chat/Controller.hs
```diff
@@ -338,7 +338,7 @@ data ChatCommand
   | APIChatItemReaction {chatRef :: ChatRef, chatItemId :: ChatItemId, add :: Bool, reaction :: MsgReaction}
   | APIGetReactionMembers {userId :: UserId, groupId :: GroupId, chatItemId :: ChatItemId, reaction :: MsgReaction}
   | APIPlanForwardChatItems {fromChatRef :: ChatRef, chatItemIds :: NonEmpty ChatItemId}
-  | APIForwardChatItems {toChatRef :: ChatRef, fromChatRef :: ChatRef, chatItemIds :: NonEmpty ChatItemId, ttl :: Maybe Int}
+  | APIForwardChatItems {toChatRef :: ChatRef, fromChatRef :: ChatRef, chatItemIds :: NonEmpty ChatItemId, ttl :: Maybe Int, sendAsGroup :: ShowGroupAsSender}
```
spaced4ndy: I think here it makes more sense to have sendAsGroup as second parameter, as it directly relates to toChatRef: send to this chat and how - with "how" related to chat direction, not chat item ids and properties like ttl.

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -525,9 +525,9 @@ setSupportChatMemberAttention db vr user g m memberAttention = do
       m_ <- runExceptT $ getGroupMemberById db vr user (groupMemberId' m)
       pure $ either (const m) id m_ -- Left shouldn't happen, but types require it
 
-createNewSndChatItem :: DB.Connection -> User -> ChatDirection c 'MDSnd -> SndMessage -> CIContent 'MDSnd -> Maybe (CIQuote c) -> Maybe CIForwardedFrom -> Maybe CITimed -> Bool -> Bool -> UTCTime -> IO ChatItemId
```
spaced4ndy, line 528: same as in Controller API also I think makes more sense to have flag parameter after chat direction. it's closely related to SendRef semantics, which [SendRef] we don't have everywhere as type in item creation but still it is implied (through forward api for example)

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -4620,7 +4630,7 @@ chatCommandP =
       "/_reaction " *> (APIChatItemReaction <$> chatRefP <* A.space <*> A.decimal <* A.space <*> onOffP <* A.space <*> (knownReaction <$?> jsonP)),
       "/_reaction members " *> (APIGetReactionMembers <$> A.decimal <* " #" <*> A.decimal <* A.space <*> A.decimal <* A.space <*> (knownReaction <$?> jsonP)),
       "/_forward plan " *> (APIPlanForwardChatItems <$> chatRefP <*> _strP),
-      "/_forward " *> (APIForwardChatItems <$> chatRefP <* A.space <*> chatRefP <*> _strP <*> sendMessageTTLP),
+      "/_forward " *> (APIForwardChatItems <$> chatRefP <* A.space <*> chatRefP <*> _strP <*> sendMessageTTLP <*> pure False),
```
spaced4ndy: use asGroupP.

currently parameter in api doesn't make sense at all with how api is implemented. but removing it from api would also be wrong as we have it in send api, and forward is basically send. so we should either remove from both apis, and always imply, or correctly pass/parse for both.

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -200,30 +200,33 @@ toggleNtf m ntfOn =
     forM_ (memberConnId m) $ \connId ->
       withAgent (\a -> toggleConnectionNtfs a connId ntfOn) `catchAllErrors` eToView
 
-prepareGroupMsg :: DB.Connection -> User -> GroupInfo -> Maybe MsgScope -> MsgContent -> Map MemberName MsgMention -> Maybe ChatItemId -> Maybe CIForwardedFrom -> Maybe FileInvitation -> Maybe CITimed -> Bool -> ExceptT StoreError IO (ChatMsgEvent 'Json, Maybe (CIQuote 'CTGroup))
-prepareGroupMsg db user g@GroupInfo {membership} msgScope mc mentions quotedItemId_ itemForwarded fInv_ timed_ live = case (quotedItemId_, itemForwarded) of
+prepareGroupMsg :: DB.Connection -> User -> GroupInfo -> Maybe MsgScope -> MsgContent -> Map MemberName MsgMention -> Maybe ChatItemId -> Maybe CIForwardedFrom -> Maybe FileInvitation -> Maybe CITimed -> Bool -> ShowGroupAsSender -> ExceptT StoreError IO (ChatMsgEvent 'Json, Maybe (CIQuote 'CTGroup))
```
spaced4ndy: same as in Controller and createNewSndChatItem, saveSndChatItems - new flag is closely related to direction and scope, better put it after Maybe MsgScope

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -1239,9 +1244,10 @@ sendHistory user gInfo@GroupInfo {groupId, membership} m@GroupMember {activeConn
                   fInv_ = fst <$> fInvDescr_
                   (mc', _, mentions') = updatedMentionNames mc formattedText mentions
                   mentions'' = M.map (\CIMention {memberId} -> MsgMention {memberId}) mentions'
+                  asGroup = isNothing sender_
               -- TODO [knocking] send history to other scopes too?
-              (chatMsgEvent, _) <- withStore $ \db -> prepareGroupMsg db user gInfo Nothing mc' mentions'' quotedItemId_ Nothing fInv_ itemTimed False
-              let senderVRange = memberChatVRange' sender
+              (chatMsgEvent, _) <- withStore $ \db -> prepareGroupMsg db user gInfo Nothing mc' mentions'' quotedItemId_ Nothing fInv_ itemTimed False asGroup
+              let senderVRange = maybe (memberChatVRange' membership) memberChatVRange' sender_
```
spaced4ndy, line 1251: not clear why we default to membership without comment
```suggestion
              -- for channel messages default chat version range to membership range
              let senderVRange = maybe (memberChatVRange' membership) memberChatVRange' sender_
```

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -1250,9 +1256,9 @@ sendHistory user gInfo@GroupInfo {groupId, membership} m@GroupMember {activeConn
                   pure . L.toList $ L.map (XMsgFileDescr msgId) parts
                 _ -> pure []
               let fileDescrChatMsgs = map (ChatMessage senderVRange Nothing) fileDescrEvents
-                  GroupMember {memberId} = sender
-                  memberName = Just $ memberShortenedName sender
-                  msgForwardEvents = map (\cm -> XGrpMsgForward memberId memberName cm itemTs) (xMsgNewChatMsg : fileDescrChatMsgs)
+                  memberId_ = memberId' <$> sender_
+                  memberName = memberShortenedName <$> sender_
```
spaced4ndy: why memberId_ is with underscore and memberName not?

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -2255,8 +2261,9 @@ saveSndChatItems ::
   [Either ChatError (NewSndChatItemData c)] ->
   Maybe CITimed ->
   Bool ->
+  ShowGroupAsSender ->
```
spaced4ndy: same - parameter after direction

spaced4ndy: and it corresponds to previously implemented "show as group" parameter for mkChatItem where it's passed after direction

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -2605,10 +2617,10 @@ createChatItems user itemTs_ dirsCIContents = do
           let ci = mkChatItem cd showGroupAsSender ciId content Nothing Nothing Nothing Nothing Nothing False False itemTs Nothing createdAt
           pure $ AChatItem (chatTypeI @c) (msgDirection @d) (toChatInfo cd) ci
 
-memberAttentionChange :: Int -> (Maybe UTCTime) -> GroupMember -> GroupChatScopeInfo -> MemberAttention
-memberAttentionChange unread brokerTs_ rcvMem = \case
+memberAttentionChange :: Int -> (Maybe UTCTime) -> Maybe GroupMember -> GroupChatScopeInfo -> MemberAttention
```
spaced4ndy, line 2623: restoring comment
```suggestion
-- rcvMem_ Nothing means message from channel - treated same as message from moderator,
-- e.g. it can reset unanswered counter if newer than last unanswered message.
memberAttentionChange :: Int -> (Maybe UTCTime) -> Maybe GroupMember -> GroupChatScopeInfo -> MemberAttention
```

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -2605,10 +2617,10 @@ createChatItems user itemTs_ dirsCIContents = do
...
+memberAttentionChange unread brokerTs_ rcvMem_ = \case
   GCSIMemberSupport (Just suppMem)
-    | groupMemberId' suppMem == groupMemberId' rcvMem -> MAInc unread brokerTs_
+    | maybe False ((groupMemberId' suppMem ==) . groupMemberId') rcvMem_ -> MAInc unread brokerTs_
```
spaced4ndy, line 2626: I think the previous syntax (from first pr) was clearer, though we usually don't use it, but I won't argue
```suggestion
    | Just rcvMem <- rcvMem_, groupMemberId' suppMem == groupMemberId' rcvMem
```

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -932,19 +932,40 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+              XMsgFileDescr sharedMsgId fileDescr -> groupMessageFileDescription gInfo' memberIdForChannel sharedMsgId fileDescr
+              XMsgUpdate sharedMsgId mContent mentions ttl live msgScope ->
+                maybe id (`memberCanSend` msgScope) memberForChannel $
+                  groupMessageUpdate gInfo' memberForChannel sharedMsgId mContent mentions msgScope msg brokerTs ttl live False
+              XMsgDel sharedMsgId memberId_ scope_ -> groupMessageDelete gInfo' memberForChannel sharedMsgId memberId_ scope_ msg brokerTs
+              XMsgReact sharedMsgId memberId scope_ reaction add -> groupMsgReaction gInfo' (Just m'') sharedMsgId memberId scope_ reaction add msg brokerTs
               -- TODO discontinue XFile
               XFile fInv -> Nothing <$ processGroupFileInvitation' gInfo' m'' fInv msg brokerTs
-              XFileCancel sharedMsgId -> xFileCancelGroup gInfo' m'' sharedMsgId
+              XFileCancel sharedMsgId -> xFileCancelGroup gInfo' memberIdForChannel sharedMsgId
```
spaced4ndy: from XMsgFileDescr to XFileCancel - murky naming/logic, will be checking functions implementations but here it's already unclear.. why passed member (sender) is now called "member for channel". for regular groups it's simply sending member (what was "m" before).

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -932,19 +932,40 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+                memberForChannel = if isChannelOwner then Nothing else Just m''
+                memberIdForChannel = memberId' <$> memberForChannel
+                showGroupAsSender' = case event of
+                  XMsgNew mc -> fromMaybe False (asGroup (mcExtMsgContent mc))
+                  XMsgUpdate {} -> isChannelOwner
+                  XMsgDel {} -> isChannelOwner
+                  XMsgReact {} -> isChannelOwner
+                  XMsgFileDescr {} -> isChannelOwner
+                  XFileCancel {} -> isChannelOwner
+                  _ -> False
```
spaced4ndy: in separation this is questionable - showGroupAsSender'. why does it decide purely based on isChannelOwner. essentially it seems we're limiting these channel events only to non-member direction. and looking at delivery task creation below it is so (see comment below).

why this is wrong: we're allowing in api for sender to choose whether to send as group or not. here it seems to always default for forwarding (delivery task) that it's "was sent as group", instead of trying to understand by previously received and referenced chat item.

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -972,8 +993,7 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
               BFileChunk sharedMsgId chunk -> Nothing <$ bFileChunkGroup gInfo' sharedMsgId chunk msgMeta
               _ -> Nothing <$ messageError ("unsupported message: " <> tshow event)
             forM deliveryJobScope_ $ \jobScope ->
-              -- TODO [channels fwd] XMsgNew to return messageFromChannel
-              pure $ NewMessageDeliveryTask {messageId = msgId, jobScope, messageFromChannel = False}
+              pure $ NewMessageDeliveryTask {messageId = msgId, jobScope, showGroupAsSender = showGroupAsSender'}
```
spaced4ndy: here showGroupAsSender' is used, and it's only based on sender role and group type, while in sending APIs we allow sender to choose whether to send as group or member. so this decision should be "returning" from event processing functions based on how item was initially received from sender, not just always sent with "no member" for channel owner. this was done correctly in original pr

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -1791,39 +1815,48 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
               else pure Nothing
           mapM_ toView cEvt_
 
-    groupMsgReaction :: GroupInfo -> GroupMember -> SharedMsgId -> MemberId -> Maybe MsgScope -> MsgReaction -> Bool -> RcvMessage -> UTCTime -> CM (Maybe DeliveryJobScope)
-    groupMsgReaction g m@GroupMember {memberRole} sharedMsgId itemMemberId scope_ reaction add RcvMessage {msgId} brokerTs
+    groupMsgReaction :: GroupInfo -> Maybe GroupMember -> SharedMsgId -> Maybe MemberId -> Maybe MsgScope -> MsgReaction -> Bool -> RcvMessage -> UTCTime -> CM (Maybe DeliveryJobScope)
```
spaced4ndy: it seems this now allows to "react as channel" - is it intended?

spaced4ndy: but then `reactor` seems wrong - see comment below

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -1791,39 +1815,48 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
                 Nothing -> do
-                  withStore' $ \db -> setGroupReaction db g m itemMemberId sharedMsgId False reaction add msgId brokerTs
-                  pure $ Just DJSGroup {jobSpec = DJDeliveryJob {includePending = False}}
+                  withStore' $ \db -> setGroupReaction db g reactor itemMemberId sharedMsgId False reaction add msgId brokerTs
+                  pure Nothing
             else pure Nothing
       | otherwise = pure Nothing
       where
+        GroupInfo {membership} = g
+        reactor = fromMaybe membership m_
```
spaced4ndy: this seems wrong - if reaction was sent "as channel", reactor defaults to membership, and it is set as membership reaction to message (reactor goes to setGroupReaction)

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -1791,39 +1815,48 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+                          Just . DJSMemberSupport <$> getScopeMemberIdViaMemberId db user g m scopeMemberId
+                    | otherwise -> pure Nothing
+                  Nothing -> do
+                    withStore' $ \db -> setGroupReaction db g reactor itemMemberId sharedMsgId False reaction add msgId brokerTs
+                    pure $ Just DJSGroup {jobSpec = DJDeliveryJob {includePending = False}}
                 Nothing -> do
-                  withStore' $ \db -> setGroupReaction db g m itemMemberId sharedMsgId False reaction add msgId brokerTs
-                  pure $ Just DJSGroup {jobSpec = DJDeliveryJob {includePending = False}}
+                  withStore' $ \db -> setGroupReaction db g reactor itemMemberId sharedMsgId False reaction add msgId brokerTs
+                  pure Nothing
```
spaced4ndy: this also seems wrong - why if "reaction as channel" is allowed, it's not then converted into delivery task? (meaning it won't be forwarded)

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -1835,84 +1868,103 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+                Nothing -> do
+                  createMemberContentItem gInfo' m' scopeInfo
+                  pure $ Just $ infoToDeliveryScope gInfo scopeInfo
+      Nothing -> do
+        file_ <- processFileInvitation fInv_ content $ \db -> createRcvGroupFileTransfer db userId gInfo Nothing
+        (ci, cInfo) <- saveRcvChatItem' user (CDChannelRcv gInfo Nothing) msg sharedMsgId_ brokerTs (CIRcvMsgContent content, ts) (snd <$> file_) timed_ live' mentions
+        reactions <- maybe (pure []) (\sharedMsgId -> withStore' $ \db -> getGroupCIReactions db gInfo Nothing sharedMsgId) sharedMsgId_
+        groupMsgToView cInfo ci {reactions}
+        autoAcceptFile file_
+        pure $ Just $ infoToDeliveryScope gInfo Nothing
```
spaced4ndy: this is almost identical to old `createContentItem` - new function is called createMemberContentItem, but instead old one could be reused

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -1835,84 +1868,103 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+            (ci, cInfo) <- saveRcvChatItem' user chatDir msg (Just sharedMsgId) brokerTs (content, ts) Nothing timed_ live mentions'
             ci' <- withStore' $ \db -> do
               createChatItemVersion db (chatItemId' ci) brokerTs mc
               updateGroupChatItem db user groupId ci content True live Nothing
-            ci'' <- blockedMemberCI gInfo' m' ci'
+            ci'' <- case chatDir of
+              CDGroupRcv gInfo' _ m' -> blockedMemberCI gInfo' m' ci'
+              CDChannelRcv {} -> pure ci'
             toView $ CEvtChatItemUpdated user (AChatItem SCTGroup SMDRcv cInfo ci'')
             pure $ Just $ infoToDeliveryScope gInfo scopeInfo
```
spaced4ndy: here we should decide "delivery context" of the item update in case item is missing and we created a new item instead (we're in catchCINotFound code branch). in original PR I decided to default to "was sent as group" in  channel+owner case. like this:

```
-- owner messages in channel without matching item are assumed to be channel messages
let sentAsGroup = useRelays' gInfo && memberRole' m == GROwner
 in Just $ infoToDeliveryContext gInfo scopeInfo sentAsGroup
```

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -1921,84 +1973,99 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+    groupMessageDelete gInfo m_ sharedMsgId sndMemberId_ scope_ rcvMsg brokerTs = case m_ of
+      Nothing -> channelDelete
+      Just m -> memberDelete m
       where
-        moderate :: GroupMember -> CChatItem 'CTGroup -> CM (Maybe DeliveryJobScope)
-        moderate mem cci = case sndMemberId_ of
+        channelDelete = do
+          withStore' (\db -> runExceptT $ getGroupChatItemBySharedMsgId db user gInfo Nothing sharedMsgId) >>= \case
+            Right cci@(CChatItem _ ChatItem {chatDir}) -> case chatDir of
+              CIChannelRcv -> Just <$> delete Nothing cci Nothing
```
spaced4ndy: again return values are not considering "sent as group" context for delivery task.

spaced4ndy: also, case above:
```
case m_ of
      Nothing -> channelDelete
      Just m -> memberDelete m
```
will not allow to correctly differentiate it, because when it's relay receiving from sender m_ will be Just.

(we have 3 cases: 1. regular groups, m_ is always Just; 2. relay groups, relay receiving from owner, m_ is Just; 3. relay groups, member receiving forwarded message from relay, m_ is Nothing. both 2 and 3 are "channel message deletions" for wherever it matters. for delivery task, only relay cares about delivery context return - that option 2 with m_ actually being Just)

tests/ChatTests/Groups.hs
```diff
@@ -8669,31 +8678,247 @@ testChannels2RelaysIncognito ps =
...
+            bob <# "#team alice> > hello"
+            bob <## "    + 👍"
+            concurrentlyN_
+              [ do cath <# "#team> > hello"
+                   cath <## "    + 👍",
+                do dan <# "#team> > hello"
+                   dan <## "    + 👍",
+                do eve <# "#team> > hello"
+                   eve <## "    + 👍"
+              ]
```
spaced4ndy, line 8900: if my understanding of reaction event processing is correct, this test may not be catching that "channel reaction" defaults to "membership reaction" - see comment in Subscriber

plans/channel_message_bugs_fix_plan.md
```diff
@@ -0,0 +1,334 @@
...
+  XMsgNew mc -> fromMaybe False (asGroup (mcExtMsgContent mc))
+  XMsgUpdate {} -> itemShowGroupAsSender ci  -- from item lookup
+  XMsgDel {} -> itemShowGroupAsSender ci
+  XMsgReact {} -> itemShowGroupAsSender ci
+  XMsgFileDescr {} -> itemShowGroupAsSender ci
+  XFileCancel {} -> itemShowGroupAsSender ci
+  _ -> False
+```
+
+**Note:** The item is already looked up for these events. Need to thread the `showGroupAsSender` field from CIMeta through to this point.
```
spaced4ndy: or rather from item direction.. showGroupAsSender is used more broadly

plans/channel_message_bugs_fix_plan.md
```diff
@@ -0,0 +1,334 @@
...
+
+#### Option C: Store `sendAsGroup` flag on reaction record
+- Add `sendAsGroup :: Bool` field to reaction storage
+- **Pros:** Explicit flag, mirrors message approach
+- **Cons:** Database migration required
+
+#### Option D: Use a special synthetic member record
+- Create a marker member record representing "channel"
+- **Pros:** Works with existing types
+- **Cons:** Pollutes member space with fake data
```
spaced4ndy: Option E (I think it is the first and preferrable option) - don't add this feature. Owner can only react as member. At least in current scope.

plans/channel_message_bugs_fix_plan.md
```diff
@@ -0,0 +1,334 @@
...
+#### Option A: Use `Nothing` for reactor
+- Change type to `Maybe GroupMember` where `Nothing` = channel
+- Requires signature changes to `getGroupReactions`, `updateGroupReaction`, etc.
+- **Pros:** Clean representation, no fake data
+- **Cons:** Many function signature changes, database schema may need update
+
+#### Option B: Use group owner's member record
+- For channel reactions, use the group owner's member record as reactor
+- **Pros:** Minimal code changes, no schema changes
+- **Cons:** Semantically incorrect (owner didn't react, channel did)
```
spaced4ndy: - similar to "remove feature".
- which owner if many?

plans/channel_message_bugs_fix_plan.md
```diff
@@ -0,0 +1,334 @@
...
+- Logic: `if isChannelOwner then sendAsGroup=True else sendAsGroup=False`
+- **Pros:** No protocol change, reasonable heuristic
+- **Cons:** Can be wrong if owner originally sent as member
+
+#### Option B: Require sender to include flag in the event
+- Add `asGroup` field to `XMsgUpdate` message format
+- **Pros:** Explicit, always correct
+- **Cons:** Protocol change, backward compatibility
+
+**Recommendation:** Option A for simplicity (update events from owner in channel are almost always for channel messages)
```
spaced4ndy: also: this edge case only happens when relay had this item deleted for some reason.

becomes slightly more probable with multi-owners if owner 1 sends message, owner 2 updates it when relay hasn't yet received it. But it can be considered together with multi-owners

plans/channel_message_bugs_fix_plan.md
```diff
@@ -0,0 +1,334 @@
...
+All CLI forward commands hardcode `False` for `sendAsGroup` instead of computing based on destination.
+
+### Fix
+Compute `sendAsGroup` before calling API based on destination group's channel status:
+
+```haskell
+-- Lines 2191, 2196, 2201: Need to determine sendAsGroup based on toChatRef
+-- If toChatRef is a channel and user is owner, sendAsGroup should default to True
+
+-- Line 4633: Parser should accept optional flag or default based on context
```
spaced4ndy: parser wouldn't know context, so yes - optional flag

plans/channel_message_bugs_fix_plan.md
```diff
@@ -0,0 +1,334 @@
...
+- `src/Simplex/Chat/Library/Commands.hs`: Lines 2191, 2196, 2201, 4633
+
+---
+
+## Test Plan
+
+### New Tests (6 total)
+
+#### Test 1: `testChannelEventDeliveryContext`
+**Objective:** Verify x.msg.update, x.msg.delete use item's sendAsGroup, not current role.
```
spaced4ndy: it's also a problem for x.msg.file.descr, x.file.cancel at least; for x.msg.react if we proceed with feature "react as channel"

plans/channel_message_bugs_fix_plan.md
```diff
@@ -0,0 +1,334 @@
...
+**Coverage:** Bug 1
+
+---
+
+#### Test 2: `testChannelReactionAttribution`
+**Objective:** Verify channel reactions are attributed correctly (not to membership).
+
+**Scenario:**
+1. Owner sends channel message
+2. Owner adds reaction
```
spaced4ndy: > Owner adds reaction

currently there's no API for owner to "react as channel".

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -1835,178 +1854,258 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+                  applyModeration gInfo' m' scopeInfo ciModeration
+                  withStore' $ \db -> deleteCIModeration db gInfo' memberId sharedMsgId_
+                  pure Nothing
+                Nothing -> do
+                  createContentItem gInfo' (Just m') scopeInfo
+                  pure $ Just $ infoToDeliveryContext gInfo scopeInfo sentAsGroup
       where
         rejected gInfo' m' scopeInfo f = newChatItem gInfo' m' scopeInfo (ciContentNoParse $ CIRcvGroupFeatureRejected f) Nothing Nothing False
-        timed' gInfo' = if forwarded then rcvCITimed_ (Just Nothing) itemTTL else rcvGroupCITimed gInfo' itemTTL
+        timed_ = if forwarded then rcvCITimed_ (Just Nothing) itemTTL else rcvGroupCITimed gInfo itemTTL
```
spaced4ndy: lost gInfo'

PR #6627, opened 2026-02-11
ios: replace LPMetadataProvider with direct HTTP fetch for link previews

marcograss:
LPMetadataProvider silently fails on many URLs, making link previews unreliable on iOS. Replace with direct HTTP fetch and HTML parsing (regex-based extraction of og:tags, title, and favicon), mirroring the approach used on Android with JSoup.


marcograss:

PR #6634, opened 2026-02-13
ios: channels and chat relays ui

spaced4ndy:


src/Simplex/Chat/Controller.hs
```diff
@@ -157,6 +157,7 @@ data ChatConfig = ChatConfig
     ciExpirationInterval :: Int64, -- microseconds
     deliveryWorkerDelay :: Int64, -- microseconds
     deliveryBucketSize :: Int,
+    channelSubscriberRole :: GroupMemberRole, -- TODO [relays] starting role should be communicated in protocol from owner to relays
```
spaced4ndy, line 160: this is a workaround to keep testing subscribers sending messages in tests, until starting role is communicated in protocol

PR #6651, opened 2026-02-22
scripts: pin ubuntu manifest digest hash for reproducibility

BarbossHack:
Hi 👋

Simplex Chat for Desktop was reproducible for v6.4.10; I tested it when it was released. However, my GitHub Action recently triggered a new check for v6.4.10, and it now appears that it is no longer reproducible.

I examined the differences between the binaries using diffoscope and observed the following:

```diff
│  String dump of section '.comment':
│ +  [     0]  GCC: (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0
│ -  [     0]  GCC: (Ubuntu 11.4.0-1ubuntu1~22.04.3) 11.4.0
...
│ │ │ │ +GCC: (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0
│ │ │ │ -GCC: (Ubuntu 11.4.0-1ubuntu1~22.04.3) 11.4.0
```

You are using "FROM ubuntu:TAG" with TAG=22.04, for example. However, Ubuntu released a new image under the same tag (22.04.3).

## Proposed in this MR

Pin the Ubuntu base image by specifying the manifest digest instead of relying on the mutable tag. This ensures that even if Ubuntu updates the 22.04 tag, your Dockerfile.build will still use the exact same image manifest.

You can check the corresponding `manifest digest` hashes here : 

- 22.04: https://hub.docker.com/layers/library/ubuntu/22.04/
- 24.04: https://hub.docker.com/layers/library/ubuntu/24.04/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -111,6 +111,7 @@ jobs:
             arch: x86_64
             runner: "ubuntu-22.04"
             ghc: "8.10.7"
+            hash: 'sha256:5c8b2c0a6c745bc177669abfaa716b4bc57d58e2ea3882fb5da67f4d59e3dda5'
             should_run: ${{ !(github.ref == 'refs/heads/stable' || startsWith(github.ref, 'refs/tags/v')) }}
           - os: 22.04
             os_underscore: 22_04
@@ -118,24 +119,28 @@ jobs:
             runner: "ubuntu-22.04"
             should_run: true
             ghc: ${{ needs.variables.outputs.GHC_VER }}
+            hash: 'sha256:5c8b2c0a6c745bc177669abfaa716b4bc57d58e2ea3882fb5da67f4d59e3dda5'
           - os: 24.04
             os_underscore: 24_04
             arch: x86_64
             runner: "ubuntu-24.04"
             should_run: true
             ghc: ${{ needs.variables.outputs.GHC_VER }}
+            hash: 'sha256:98ff7968124952e719a8a69bb3cccdd217f5fe758108ac4f21ad22e1df44d237'
           - os: 22.04
             os_underscore: 22_04
             arch: aarch64
             runner: "ubuntu-22.04-arm"
             should_run: true
             ghc: ${{ needs.variables.outputs.GHC_VER }}
+            hash: 'sha256:6a62a4157b8775eaf4959cb629e757d32d39d1f4c8ac1b0ddc2510b555cf72f3'
           - os: 24.04
             os_underscore: 24_04
             arch: aarch64
             runner: "ubuntu-24.04-arm"
             should_run: true
             ghc: ${{ needs.variables.outputs.GHC_VER }}
+            hash: 'sha256:68434214381cb38287104e629fe8ee720167dd98cbb36ab1cbbab342515fa6ab'
     steps:
       - name: Checkout Code
         if: matrix.should_run == true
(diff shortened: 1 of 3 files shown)


shumvgolove:
Thanks, will look into it.

shumvgolove:
Sorry for the wait @BarbossHack 

Please add similar changes to [`build.yml`](https://github.com/simplex-chat/simplex-chat/blob/stable/.github/workflows/build.yml#L169-L179) and let's switch to latest Ubuntu tags, e.g.:

**22.04**: `sha256:3ba65aa20f86a0fad9df2b2c259c613df006b2e6d0bfcc8a146afb8c525a9751`
**24.04**: `sha256:d1e2e92c075e5ca139d51a140fff46f84315c0fdce203eab2807c7e495eff4f9`

BarbossHack:
> let's switch to latest Ubuntu tags, e.g.:
> 
> **22.04**: `sha256:3ba65aa20f86a0fad9df2b2c259c613df006b2e6d0bfcc8a146afb8c525a9751` **24.04**: `sha256:d1e2e92c075e5ca139d51a140fff46f84315c0fdce203eab2807c7e495eff4f9`

These are the `index digest`, but we should use the `manifest digest` instead (I already set the latest versions). This is the digest of the platform-specific manifest. While the index digest is the digest of the manifest list (or index), a higher-level structure that points to multiple platform-specific manifests for multi-architecture support (e.g., amd64, arm64).



That's also the digest you get while inspecting the tag on an amd64 machine



> Please add similar changes to [build.yml](https://github.com/simplex-chat/simplex-chat/blob/stable/.github/workflows/build.yml#L169-L179)

I force-pushed a new commit 👍 (I also set the arm64 manifest digests for ARM builds for homogeneity)

shumvgolove:
Got it, thanks for clarification.

shumvgolove:
```
[2](https://github.com/simplex-chat/simplex-chat/actions/runs/22796057299/job/66131028524?pr=6651#step:8:198)
Dockerfile.build:4
--------------------
   2 |     ARG TAG=24.04
   3 |     ARG HASH=sha256:98ff7968124952e719a8a69bb3cccdd217f5fe758108ac4f21ad22e1df44d237
   4 | >>> FROM ubuntu:${TAG}@{HASH} AS build
   5 |     
   6 |     ### Build stage
--------------------
ERROR: failed to build: failed to solve: failed to parse stage name "ubuntu:22.04@{HASH}": invalid reference format
```

Perhaps we need to strip `sha256:` from the `HASH`?

BarbossHack:
Oh, the yaml wokrflow misunderstood the `hash: sha256:xxxx` and tough sha256 was a yaml key, I will escape them with single quotes `hash: 'sha256:xxxx'`, sry 

(but yeah, "sha256:" is required when specifying a manifest digest)

BarbossHack:
For reference, Signal-Android reproducible image: https://github.com/signalapp/Signal-Android/blob/main/reproducible-builds/Dockerfile#L1

shumvgolove:
Unfortunately, CI fail with the same error.

BarbossHack:
it was just a typo 🥲

```
ubuntu:${TAG}@{HASH}
```

instead of

```
ubuntu:${TAG}@${HASH}
```

BarbossHack:
These new CI errors do not seem to come from my changes…?

shumvgolove:
Yes, all good. I'll test the reproducible script and make sure nothing else breaks.

PR #6678, opened 2026-03-14
core: forward signed messages as unchanged binary strings

epoberezkin:


src/Simplex/Chat/Messages/Batch.hs
```diff
@@ -69,33 +72,33 @@ batchMessages mode maxLen = addBatch . foldr addToBatch ([], [], [], 0, 0)
...
   where
     addToBatch :: ([ByteString], [Int64], [Int64], Int, Int) -> MessageDeliveryTask -> ([ByteString], [Int64], [Int64], Int, Int)
     addToBatch (msgBodies, taskIds, largeTaskIds, len, n) task
-      -- too large: skip msgBody, record taskId in largeTaskIds
+      -- too large: skip, record taskId in largeTaskIds
       | msgLen > maxLen = (msgBodies, taskIds, taskId : largeTaskIds, len, n)
       -- fits: include in batch
       -- batch overhead: '=' + count (2) + 2-byte length prefix per element
-      | len' + (n + 1) * 2 + 2 <= maxLen = (msgBody : msgBodies, taskId : taskIds, largeTaskIds, len', n + 1)
+      | len' + (n + 1) * 2 + 2 <= maxLen = (fwdBody : msgBodies, taskId : taskIds, largeTaskIds, len', n + 1)
```
epoberezkin:
```suggestion
      | len' + (n + 1) * 2 + 2 <= maxLen = (msgBody : msgBodies, taskId : taskIds, largeTaskIds, len', n + 1)
```

src/Simplex/Chat/Messages/Batch.hs
```diff
@@ -69,33 +72,33 @@ batchMessages mode maxLen = addBatch . foldr addToBatch ([], [], [], 0, 0)
...
       -- doesn't fit: stop adding further messages
       | otherwise = (msgBodies, taskIds, largeTaskIds, len, n)
       where
-        MessageDeliveryTask {taskId, fwdSender, brokerTs = fwdBrokerTs, chatMessage} = task
-        msgBody = encodeFwdElement GrpMsgForward {fwdSender, fwdBrokerTs} chatMessage
-        msgLen = B.length msgBody
+        MessageDeliveryTask {taskId, fwdSender, brokerTs = fwdBrokerTs, verifiedMsg} = task
+        fwdBody = encodeFwdElement GrpMsgForward {fwdSender, fwdBrokerTs} verifiedMsg
```
epoberezkin:
```suggestion
        msgBody = encodeFwdElement GrpMsgForward {fwdSender, fwdBrokerTs} verifiedMsg
```

src/Simplex/Chat/Messages/Batch.hs
```diff
@@ -69,33 +72,33 @@ batchMessages mode maxLen = addBatch . foldr addToBatch ([], [], [], 0, 0)
...
+        msgLen = B.length fwdBody
```
epoberezkin:
```suggestion
        msgLen = B.length msgBody
```

tests/ChatTests/Groups.hs
```diff
@@ -8409,6 +8409,107 @@ testChannels1RelayDeliver ps =
             eve <# "#team cath> > hi"
             eve <## "    + 👍"
 
+            -- admin operations are signed in channels
+
+            -- update group profile (XGrpInfo) - signed
```
spaced4ndy, line 8414: separate into different tests to make it clear what's being tested in which test

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -3192,32 +3190,34 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+          Just sm -> VMSigned sm chatMsg
+        verified = case signedMsg_ of
+          Just SignedMsg {chatBinding, signatures, signedBody}
+            | GroupMember {memberPubKey = Just pubKey, memberId} <- member ->
+                case chatBinding of
+                  CBGroup | Just GroupKeys {groupRootKey} <- groupKeys gInfo ->
+                    let prefix = smpEncode chatBinding <> smpEncode (groupRootPubKey groupRootKey, memberId)
+                     in all (\(MsgSignature KRMember sig) -> C.verify (C.APublicVerifyKey C.SEd25519 pubKey) sig (prefix <> signedBody)) signatures
+                  _ -> True -- can't reconstruct binding → accept (enforcement in Step 5)
+            | otherwise -> True
```
spaced4ndy, line 3219: if member has no key, but message was sent with forged SignedMsg, message would pass verification. I think it should be rejected here
```suggestion
            | otherwise -> False
```

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -3192,32 +3190,34 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+          Nothing -> VMUnsigned chatMsg
+          Just sm -> VMSigned sm chatMsg
+        verified = case signedMsg_ of
+          Just SignedMsg {chatBinding, signatures, signedBody}
+            | GroupMember {memberPubKey = Just pubKey, memberId} <- member ->
+                case chatBinding of
+                  CBGroup | Just GroupKeys {groupRootKey} <- groupKeys gInfo ->
+                    let prefix = smpEncode chatBinding <> smpEncode (groupRootPubKey groupRootKey, memberId)
+                     in all (\(MsgSignature KRMember sig) -> C.verify (C.APublicVerifyKey C.SEd25519 pubKey) sig (prefix <> signedBody)) signatures
+                  _ -> True -- can't reconstruct binding → accept (enforcement in Step 5)
```
spaced4ndy, line 3218: what's "enforcement in Step 5"? Why not reject here as well? It's a new feature, all clients would have chat binding for a new channel, wouldn't they?

src/Simplex/Chat/Protocol.hs
```diff
@@ -1361,6 +1368,20 @@ chatMsgToBody chatMsg = case encoding @e of
   SBinary -> chatMsgBinaryToBody chatMsg
   SJson -> LB.toStrict $ J.encode chatMsg
 
+verifiedChatMsg :: VerifiedMsg e -> ChatMessage e
```
spaced4ndy, line 1371: it's in previous PR, but in requiresSignature - why shouldn't XGrpLeave be signed?

tests/ChatTests/Groups.hs
spaced4ndy: also, add tests for members receiving unsigned group altering events? or signed with wrong signatures? this would probably require some flag in code to break it, modifying what's being tested, but it could be applied in minimal number of places and would allow to test behavior of member in those cases

PR #6701, opened 2026-03-23
directory: expand service README

Narasimha-sc:
## Summary

- Add build prerequisites (system packages, GHC 9.6.3 / Cabal 3.10.2+) and `cabal build` / `cabal run` instructions
- Document running the bot: bootstrap flow to discover contact ID, minimal invocation, and full CLI flag reference
- Document self-hosting the directory website via `--web-folder` + static server
- Add full bot command reference: search/browse, group registration, user group management (list, role, filter, link, delete), admin commands, super-user commands, and group lifecycle diagram

## Test plan

- [ ] README renders correctly on GitHub
- [ ] All commands and flags match current `simplex-directory-service` implementation

### DIFF
--- a/apps/simplex-directory-service/README.md
+++ b/apps/simplex-directory-service/README.md
@@ -72,33 +72,24 @@ simplex-directory-service \
 
 ---
 
-## Running Your Own Website
+## Hosting the Directory Page
 
-The bot writes `listing.json`, `promoted.json`, and group images to `--web-folder` (updated every 5 minutes or on group approval).
+The `web/` folder has a ready-to-use directory page — serve it with any static web server. No build step needed. Dark mode follows system preference. `directory.js` is a symlink to `website/src/js/directory.js` so it stays in sync.
 
-**Step 1 — point `--web-folder` at the website output directory:**
+The file must be served under a URL path starting with `/directory` (e.g. `http://example.com/directory.html` or `http://example.com/directory/`) — `directory.js` skips initialisation otherwise.
+
+By default the page shows the official SimpleX directory at `https://directory.simplex.chat/data/`. To use your own bot's data instead, change this line near the bottom of `directory.js`:
+```js
+const simplexDirectoryDataURL = 'https://your-domain.example/data/';
+```
+
+Then run the bot with `--web-folder` pointing to that path. The bot writes `listing.json`, `promoted.json`, and group images there (refreshed every 5 minutes or on approval):
 ```sh
 simplex-directory-service \
   --super-users 2:alice \
   --database /path/to/db \
-  --web-folder /path/to/simplex-chat/website/_site/directory-data
+  --web-folder /var/www/your-domain.example/data
 ```
-
-**Step 2 — switch the data URL in `website/src/js/directory.js`.** The file contains two URL definitions (around line 484); comment out the production one and uncomment the localhost one:
-```js
-// const simplexDirectoryDataURL = 'https://directory.simplex.chat/data/';
-const simplexDirectoryDataURL = 'http://localhost:8080/directory-data/';
-```
-
-**Step 3 — build the website and host `_site/` with a static server:**
-```sh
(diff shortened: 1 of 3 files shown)


apps/simplex-directory-service/README.md
```diff
@@ -1,5 +1,269 @@
...
+
+---
+
+### 6. Group Lifecycle
+
+```
+Invited by owner
+      │
+      ▼
+PendingConfirmation  ──(/confirm)──▶  Proposed
```
epoberezkin: PendingConfirmation state only happens in case on duplicate name.

apps/simplex-directory-service/README.md
```diff
@@ -1,5 +1,269 @@
...
+                               (add bot link to welcome)
+                                          │
+                                          ▼
+                                   PendingApproval
+                                   (admin reviews)
+                                          │
+                              ┌───────────┴───────────┐
+                              ▼                       ▼
+                           Active             (profile change
+                          (listed)          → re-enters pending)
```
epoberezkin: On "re-enters pending" - it's not as simple, and depends on whether correct link remained in the profile.

apps/simplex-directory-service/README.md
```diff
@@ -1,5 +1,269 @@
...
+```
+/confirm <ID>:<name>
+```
+
+Confirms registration when another group with the same display name already exists. `ID` and `name` are provided in the bot's prompt.
+
+#### View or set default join role
+
+```
+/role <ID>[:<name>] [member|observer]
```
epoberezkin, line 182: it's unclear here and below when name must be used

apps/simplex-directory-service/README.md
```diff
@@ -1,5 +1,269 @@
...
+#### List groups
+
+```
+/list
+/ls
+```
+
+Shows all groups you have registered with their current status.
+
+#### Confirm duplicate name
```
epoberezkin, line 169: unclear why it exists

apps/simplex-directory-service/README.md
Narasimha-sc: Fixed in 4d36e7be6. The lifecycle diagram now branches at registration: a unique name goes straight to `Proposed`, while a duplicate name goes through `PendingConfirmation` → `/confirm` → `Proposed`. The footnote and state notes clarify this only happens when the existing group is registered but not yet listed/suspended — if the duplicate is already listed or suspended, registration is blocked entirely.

apps/simplex-directory-service/README.md
Narasimha-sc: Fixed in 4d36e7be6. The footnote and the `PendingApproval` state note now explain that a profile change only re-triggers approval when fields other than the directory bot link change; if the only change is swapping the old bot link for the new one (or a whitespace-only edit to the description), the group stays `Active` — matching `onlyLinkChanged` in `Service.hs`.

apps/simplex-directory-service/README.md
Narasimha-sc: Fixed in 4d36e7be6. Added a note at the top of section 3: `/role`, `/filter` and `/link` accept just `<ID>` (the `:<name>` part is optional), while `/confirm`, `/delete` and all admin/super-user commands require `<ID>:<name>`. The bot always includes the exact `ID:name` string in its messages so it can be copied directly.

apps/simplex-directory-service/README.md
Narasimha-sc: Fixed in 4d36e7be6. The `/confirm` section now explains the purpose: when you invite the bot to a group whose display name matches an already-registered group, the bot pauses registration and asks for explicit confirmation, to avoid accidentally registering a different group that happens to share a name. `/confirm` acknowledges the duplicate and proceeds. (It is not offered when the name is already listed or suspended — that case is blocked outright.)

shumvgolove (approved):

PR #6805, opened 2026-04-14
core, ui: tolerate unknown MsgContentTag in chatContentTypes

Narasimha-sc:
## Summary
- Fixes the error opening a group containing items with unknown or legacy `msg_content_tag` values (e.g. `"liveText"`). Strict enum decoding of `List<MsgContentTag>` in `apiGetChatContentTypes` was failing on the whole response; backend already round-trips unknowns via `MCUnknown_`, so the clients now do the same.
- Follows the open-union pattern used by `GroupType` / `ReportReason`: preserve the original type string in the `Unknown` variant so it round-trips instead of collapsing to a sentinel.
- Multiplatform (Kotlin): `MsgContentTag` becomes a `sealed class` with `object` cases for known tags and `data class Unknown(val type: String)`; custom `MsgContentTagSerializer` maps `"liveText"` → `Text` and wraps everything else in `Unknown(s)`. `apiGetChatContentTypes` filters out `Unknown`.
- iOS (Swift): `MsgContentTag` becomes a `Codable, Hashable` enum with `case unknown(type: String)`; `rawValue` retained as a computed property so existing call sites compile unchanged. `"liveText"` → `.text`; `.chat` case added for parity with backend/multiplatform. `apiGetChatContentTypes` filters out `.unknown`.

## Test plan
- [ ] Open a group that contains a message with an unknown/legacy `msg_content_tag` on Android — group opens without the "Error loading details: chatContentTypes" error.
- [ ] Repeat on iOS.
- [ ] Regression: open a normal group — content types load unchanged, no unknowns shown.

### DIFF
--- a/apps/ios/Shared/Model/SimpleXAPI.swift
+++ b/apps/ios/Shared/Model/SimpleXAPI.swift
@@ -458,7 +458,7 @@ func apiGetChat(chatId: ChatId, scope: GroupChatScope?, contentTag: MsgContentTa
 
 func apiGetChatContentTypes(chatId: ChatId, scope: GroupChatScope? = nil) async throws -> [MsgContentTag] {
     let r: ChatResponse0 = try await chatSendCmd(.apiGetChatContentTypes(chatId: chatId, scope: scope))
-    if case let .chatContentTypes(types) = r { return types }
+    if case let .chatContentTypes(types) = r { return types.filter { if case .unknown = $0 { return false }; return true } }
     throw r.unexpected
 }
 
--- a/apps/ios/SimpleXChat/ChatTypes.swift
+++ b/apps/ios/SimpleXChat/ChatTypes.swift
@@ -4773,7 +4773,7 @@ extension MsgContent: Encodable {
     }
 }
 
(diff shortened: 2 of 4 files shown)


apps/ios/Shared/Model/SimpleXAPI.swift
```diff
@@ -458,7 +458,7 @@ func apiGetChat(chatId: ChatId, scope: GroupChatScope?, contentTag: MsgContentTa
 
 func apiGetChatContentTypes(chatId: ChatId, scope: GroupChatScope? = nil) async throws -> [MsgContentTag] {
     let r: ChatResponse0 = try await chatSendCmd(.apiGetChatContentTypes(chatId: chatId, scope: scope))
-    if case let .chatContentTypes(types) = r { return types }
+    if case let .chatContentTypes(types) = r { return types.filter { if case .unknown = $0 { return false }; return true } }
```
epoberezkin, line 461:
```suggestion
    if case let .chatContentTypes(types) = r { return types.filter { if case .unknown = $0 { false } else true } }
```

PR #6811, opened 2026-04-14
website: snap background-position to integer pixels

another-simple-pixel:
CSS background-position: center can land on half-pixels when the container width/height is odd. This makes background images visibly blurry at native resolution (most noticeable on 1920px displays). Added a small JS fix that snaps the position to whole pixels on load and resize.


epoberezkin:
recheck

epoberezkin:
needs throttling on resize (or using animation frame?) and taking pixel density into account

PR #6831, opened 2026-04-18
core, ui: delete member items from DB when fullDelete is on

shumvgolove:


kegomimu:
Could you update the PR to include the target OS (e.g. android, ui)?

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -510,23 +510,21 @@ updateACIGroupInfo gInfo' = \case
     AChatItem SCTGroup dir (GroupChat gInfo' chatScopeInfo) ci
   aci -> aci
 
-deleteGroupMemberCIs :: MsgDirectionI d => User -> GroupInfo -> GroupMember -> GroupMember -> SMsgDirection d -> CM ()
-deleteGroupMemberCIs user gInfo member byGroupMember msgDir = do
-  deletedTs <- liftIO getCurrentTime
-  filesInfo <- withStore' $ \db -> deleteGroupMemberCIs_ db user gInfo member byGroupMember msgDir deletedTs
+fullDeleteGroupMemberCIs :: User -> GroupInfo -> GroupMember -> CM ()
```
```suggestion
deleteGroupMemberCIs :: User -> GroupInfo -> GroupMember -> CM ()
```

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -510,23 +510,21 @@ updateACIGroupInfo gInfo' = \case
...
+fullDeleteGroupMemberCIs user gInfo member = do
+  filesInfo <- withStore' $ \db -> getGroupMemberFileInfo db user gInfo member
   deleteCIFiles user filesInfo
+  withStore' $ \db -> do
+    fullDeleteMemberCIs db user gInfo member
```
```suggestion
```

I think it should not be needed at all

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -510,23 +510,21 @@ updateACIGroupInfo gInfo' = \case
...
+    deleteGroupMember db user member
 
-deleteGroupMembersCIs :: User -> GroupInfo -> [GroupMember] -> GroupMember -> CM ()
-deleteGroupMembersCIs user gInfo members byGroupMember = do
-  deletedTs <- liftIO getCurrentTime
-  filesInfo <- withStore' $ \db -> fmap concat $ forM members $ \m -> deleteGroupMemberCIs_ db user gInfo m byGroupMember SMDRcv deletedTs
+fullDeleteGroupMembersCIs :: User -> GroupInfo -> [GroupMember] -> CM ()
```
```suggestion
deleteGroupMembersCIs :: User -> GroupInfo -> [GroupMember] -> CM ()
```

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -510,23 +510,21 @@ updateACIGroupInfo gInfo' = \case
...
+  filesInfo <- withStore' $ \db -> fmap concat $ forM members $ \m -> getGroupMemberFileInfo db user gInfo m
   deleteCIFiles user filesInfo
-
-deleteGroupMemberCIs_ :: MsgDirectionI d => DB.Connection -> User -> GroupInfo -> GroupMember -> GroupMember -> SMsgDirection d -> UTCTime -> IO [CIFileInfo]
-deleteGroupMemberCIs_ db user gInfo member byGroupMember msgDir deletedTs = do
-  fs <- getGroupMemberFileInfo db user gInfo member
-  updateMemberCIsModerated db user gInfo member byGroupMember msgDir deletedTs
-  pure fs
+  withStore' $ \db -> forM_ members $ \m -> do
+    fullDeleteMemberCIs db user gInfo m
```
```suggestion
```

and this

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -2814,39 +2814,19 @@ updateGroupChatItemModerated db User {userId} GroupInfo {groupId} ci m@GroupMemb
...
-  itemIds <- updateCIs =<< getCurrentTime
-#if defined(dbPostgres)
-  let inItemIds = Only $ In (map fromOnly itemIds)
-  DB.execute db "DELETE FROM messages WHERE message_id IN (SELECT message_id FROM chat_item_messages WHERE chat_item_id IN ?)" inItemIds
-  DB.execute db "DELETE FROM chat_item_versions WHERE chat_item_id IN ?" inItemIds
-#else
-  DB.executeMany db deleteChatItemMessagesQuery itemIds
-  DB.executeMany db "DELETE FROM chat_item_versions WHERE chat_item_id = ?" itemIds
-#endif
+fullDeleteMemberCIs :: DB.Connection -> User -> GroupInfo -> GroupMember -> IO ()
```
epoberezkin: so this function won't be needed at all then?

src/Simplex/Chat/Library/Internal.hs
shumvgolove: fullDeleteMemberCIs is needed because it deletes orphaned messages rows via chat_item_messages join before deleting chat_items — deleteGroupMember only deletes chat_items and doesn't clean up messages, which have no cascade from chat_items.

src/Simplex/Chat/Library/Internal.hs
shumvgolove: Same reason - it's the batch version of the same function. Without it, messages rows would be orphaned when deleteGroupMember deletes chat_items.

src/Simplex/Chat/Store/Messages.hs
shumvgolove: deleteGroupMember deletes chat_items which cascades to chat_item_messages, but messages rows don't cascade from chat_item_messages - they become permanently orphaned. fullDeleteMemberCIs deletes messages via the chat_item_messages join while it still exists, before deleteGroupMember removes chat_items.

src/Simplex/Chat/Store/Messages.hs
```diff
@@ -2814,39 +2814,19 @@ updateGroupChatItemModerated db User {userId} GroupInfo {groupId} ci m@GroupMemb
...
-            (updateQuery <> " AND group_member_id = ? RETURNING chat_item_id")
-            (columns :. (userId, groupId, memId))
-      where
-        columns = (deletedTs, groupMemberId' byGroupMember, msgDirToModeratedContent_ md, ciModeratedText, currentTs)
+    deleteWithCond :: ToRow p => Query -> p -> IO ()
+    deleteWithCond cond params = do
+      let ciQuery = "SELECT chat_item_id FROM chat_items WHERE user_id = ? AND group_id = ? " <> cond
+      -- delete messages first (they don't cascade from chat_items), using chat_item_messages join
+      DB.execute db ("DELETE FROM messages WHERE message_id IN (SELECT message_id FROM chat_item_messages WHERE chat_item_id IN (" <> ciQuery <> "))") params
+      -- delete chat_items; chat_item_messages, chat_item_versions, chat_item_mentions, files, calls, group_snd_item_statuses all cascade
```
epoberezkin, line 2828: This is still wrong, there is DELETE query syntax where you don't read IDs to delete messages, but references chat_items table right in DELETE. Should work in both SQLite and Postgres, we use it elsewhere.

PR #6865, opened 2026-04-22
desktop: keep text selection on the originally selected message when a new message arrives

Narasimha-sc:
**Problem:** On desktop, selected text jumped onto a neighbouring message whenever a new message arrived or was sent.
**Cause:** Selection stored positional indices into the reversed merged-items list, so any prepend (or deletion) shifted those indices onto different items.
**Fix:** Anchor the selection to `ChatItem.id` alongside the positions, and resync the indices to where the anchored items now live on every recomposition (`SideEffect`); clear the selection if either anchor item is gone.


Narasimha-sc:
Superseded by #6955 (recreated from a branch in the upstream repository instead of a fork).

PR #6867, opened 2026-04-23
android, desktop: fix link previews bypassing SOCKS proxy

iversonianGremling:
NEEDS TESTING ON ANDROID, DOESN'T WORK ON iOS

It's worth noting that despite this fix leaks to the DNS resolver can occur, the users are properly warned about this through alerts.

## Changes

- Route both the HTML fetch and image download through the configured SOCKS proxy via "java.net.Proxy" instead of vanilla getLinkPreview with URL.openStream()

EXTRA GOODIES:
- Fail closed: if the proxy address is misconfigured (unparseable port), the preview is cancelled and the user is alerted instead of falling back to a direct connection <- Might be a bit paranoid but better safe than sorry
- Warn users when enabling SOCKS with link previews active, or enabling link previews while SOCKS is active, about residual DNS leak risk
- Update the SOCKS limitations notice to reflect that link previews are now proxied, and highlight it in warning colour
- Changed the warning on the SOCKS section to orange to make it more noticeable

## Leak verification

Tested on Arch Linux using strace to trace all connect() syscalls on the app process while triggering a link preview with SOCKS enabled (Tor on localhost:9050).

**Before fix** — direct connection to destination on port 443:
```
connect(141, {sa_family=AF_INET6, sin6_port=htons(443),
  inet_pton(AF_INET6, "::ffff:104.18.18.5", &sin6_addr)}, 28) = 0
```

**After fix** — connection goes through Tor on port 9050, no direct 443:
```
connect(196, {sa_family=AF_UNIX, sun_path="/run/systemd/resolve/io.systemd.Resolve"}, 42) = 0  ← local DNS query query problem previously mentioned
connect(196, {sa_family=AF_INET, sin_port=htons(0), sin_addr=inet_addr("104.18.19.5")}, 16) = 0  ← DNS result probe (port 0, no data sent)
connect(196, {sa_family=AF_INET, sin_port=htons(0), sin_addr=inet_addr("104.18.18.5")}, 16) = 0  ← DNS result probe (port 0, no data sent)
connect(196, {sa_family=AF_INET6, sin6_port=htons(9050),
  inet_pton(AF_INET6, "::ffff:127.0.0.1", &sin6_addr)}, 28) = 0  <- actual connection via Tor
```

The DNS queries (hostname -> IP) happen locally before the SOCKS connection is made. The "port=htons(0)" probes are the JVM checking reachability of resolved IPs, no actual data is transmitted to the destination. The only real TCP connection goes to 127.0.0.1:9050 (Tor). No direct port 443 connection is made.

## Test if it works

- [ ] Check for leaks and if image preview connections are actually proxied
- [ ] Enable SOCKS proxy -> warning dialog appears if link previews are on, with option to disable them and vice versa
- [ ] SOCKS enabled: paste a URL -> preview loads, \`strace\` shows port 9050 traffic only, no port 443
- [ ] Misconfigured proxy (non-numeric port) -> preview silently fails with an alert, no direct connection
- [ ] SOCKS disabled: link previews work as before

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/helpers/LinkPreviews.kt
+++ b/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/helpers/LinkPreviews.kt
@@ -16,6 +16,7 @@ import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import chat.simplex.common.model.ChatController.appPrefs
+import chat.simplex.common.model.ChatController.getNetCfg
 import chat.simplex.common.model.LinkPreview
 import chat.simplex.common.platform.*
 import chat.simplex.common.ui.theme.*
@@ -26,6 +27,8 @@ import chat.simplex.res.MR
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 import org.jsoup.Jsoup
+import java.net.InetSocketAddress
+import java.net.Proxy
 import java.net.URL
 
 private const val OG_SELECT_QUERY = "meta[property^=og:]"
@@ -37,6 +40,20 @@ suspend fun getLinkPreview(url: String): LinkPreview? {
     try {
       val title: String?
       val u = kotlin.runCatching { URL(url) }.getOrNull() ?: return@withContext null
+      val netCfg = getNetCfg()
+      val proxy: Proxy? = if (netCfg.useSocksProxy && netCfg.socksProxy != null) {
+        val hostname = netCfg.socksProxy.substringBefore(":").ifEmpty { "localhost" }
+        val port = netCfg.socksProxy.substringAfter(":").toIntOrNull()
+        if (port != null) {
+          Proxy(Proxy.Type.SOCKS, InetSocketAddress(hostname, port))
+        } else {
+          AlertManager.shared.showAlertMsg(
+            title = generalGetString(MR.strings.network_socks_proxy),
+            text = generalGetString(MR.strings.socks_proxy_invalid_address)
+          )
+          return@withContext null
+        }
+      } else null
       var imageUri = when {
         IMAGE_SUFFIXES.any { u.path.lowercase().endsWith(it) } -> {
           title = u.path.substringAfterLast("/")
@@ -47,6 +64,7 @@ suspend fun getLinkPreview(url: String): LinkPreview? {
             .ignoreContentType(true)
             .timeout(10000)
             .followRedirects(true)
+            .proxy(proxy)
 
           val response = if (url.lowercase().startsWith("https://x.com/")) {
             // Apple sends request with special user-agent which handled differently by X.com.
@@ -68,7 +86,8 @@ suspend fun getLinkPreview(url: String): LinkPreview? {
       if (imageUri != null) {
         imageUri = normalizeImageUri(u, imageUri)
         try {
-          val stream = URL(imageUri).openStream()
+          val conn = URL(imageUri).openConnection(proxy ?: Proxy.NO_PROXY)
+          val stream = conn.getInputStream()
           val image = resizeImageToStrSize(stream.use(::loadImageBitmap), maxDataSize = 14000)
           //          TODO add once supported in iOS
           //          val description = ogTags.firstOrNull {
--- a/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/helpers/LinkPreviews.kt
+++ b/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/helpers/LinkPreviews.kt
@@ -16,8 +16,8 @@ import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import chat.simplex.common.model.ChatController.appPrefs
-import chat.simplex.common.model.ChatController.getNetCfg
 import chat.simplex.common.model.LinkPreview
+import chat.simplex.common.model.NetworkProxyAuth
 import chat.simplex.common.platform.*
 import chat.simplex.common.ui.theme.*
 import chat.simplex.common.views.chat.chatViewScrollState
@@ -27,7 +27,9 @@ import chat.simplex.res.MR
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 import org.jsoup.Jsoup
+import java.net.Authenticator
 import java.net.InetSocketAddress
+import java.net.PasswordAuthentication
 import java.net.Proxy
 import java.net.URL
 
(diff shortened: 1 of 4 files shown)


iversonianGremling:

epoberezkin:
recheck

epoberezkin:
can you merge master please? We're reviewing - thinking to include in 6.5 if all good.

Narasimha-sc:
I have reviewed your code with Claude, it found some critical bugs, please fix them before merge:
Critical bugs                                                   
                                                                                
  1. SOCKS proxy parser breaks on auth and IPv6 — LinkPreviews.kt:45-46         
  
  val hostname = netCfg.socksProxy.substringBefore(":").ifEmpty { "localhost" } 
  val port = netCfg.socksProxy.substringAfter(":").toIntOrNull()  
                                                                                
  netCfg.socksProxy is produced by NetworkProxy.toProxyString()                 
  (SimpleXAPI.kt:4774). It is not a plain host:port string. 
                                                                  
  NetworkProxyAuth.USERNAME and IPv6 host wrapping […] are real, supported      
  configurations in this codebase. For those users, the new code either pops the
   "invalid SOCKS proxy address" alert every time a URL is typed into the       
  composer, or silently passes a bogus host like "@". Combined with bug #3
  below, this becomes outright unusable.

  The fix is structural, not stringy: there is already a NetworkProxy data class
   with typed host: String / port: Int fields, fetched via
  appPrefs.networkProxy.get(). Build the Proxy from that — every edge case      
  disappears. (good-code-v5 §3, §27.)                             

  2. SOCKS authentication is silently dropped — LinkPreviews.kt:48              
  
  Proxy(Proxy.Type.SOCKS, InetSocketAddress(host, port)) carries no credentials.
   Java's SOCKS5 implementation only authenticates if a java.net.Authenticator
  is registered for the proxy. The new code never registers one. So even on     
  configurations where parsing happens to succeed alongside       
  NetworkProxyAuth.USERNAME with non-empty credentials, the proxy will reject
  the connection (or accept noauth), and the preview will fail or leak — with no
   user-visible signal beyond e.printStackTrace().

  3. Alert is shown per URL keystroke — LinkPreviews.kt:50                      
  
  getLinkPreview is called from ComposeView.loadLinkPreview on every detected   
  URL change (1.5 s debounce only on first attempt). On a         
  misconfigured/IPv6/auth proxy, the alert pops up repeatedly while the user    
  types or pastes URLs. There is no de-dup or per-session suppression. The alert
   also runs from Dispatchers.IO; that happens to work because showAlert is just
   state mutation, but the call site doesn't make that contract obvious.

iversonianGremling:
> I have reviewed your code with Claude, it found some critical bugs, please fix them before merge: Critical bugs
> 
>     1. SOCKS proxy parser breaks on auth and IPv6 — LinkPreviews.kt:45-46
> 
> 
> val hostname = netCfg.socksProxy.substringBefore(":").ifEmpty { "localhost" } val port = netCfg.socksProxy.substringAfter(":").toIntOrNull()
> 
> netCfg.socksProxy is produced by NetworkProxy.toProxyString() (SimpleXAPI.kt:4774). It is not a plain host:port string.
> 
> NetworkProxyAuth.USERNAME and IPv6 host wrapping […] are real, supported configurations in this codebase. For those users, the new code either pops the "invalid SOCKS proxy address" alert every time a URL is typed into the composer, or silently passes a bogus host like "@". Combined with bug #3 below, this becomes outright unusable.
> 
> The fix is structural, not stringy: there is already a NetworkProxy data class with typed host: String / port: Int fields, fetched via appPrefs.networkProxy.get(). Build the Proxy from that — every edge case disappears. (good-code-v5 §3, §27.)
> 
>     2. SOCKS authentication is silently dropped — LinkPreviews.kt:48
> 
> 
> Proxy(Proxy.Type.SOCKS, InetSocketAddress(host, port)) carries no credentials. Java's SOCKS5 implementation only authenticates if a java.net.Authenticator is registered for the proxy. The new code never registers one. So even on configurations where parsing happens to succeed alongside NetworkProxyAuth.USERNAME with non-empty credentials, the proxy will reject the connection (or accept noauth), and the preview will fail or leak — with no user-visible signal beyond e.printStackTrace().
> 
>     3. Alert is shown per URL keystroke — LinkPreviews.kt:50
> 
> 
> getLinkPreview is called from ComposeView.loadLinkPreview on every detected URL change (1.5 s debounce only on first attempt). On a misconfigured/IPv6/auth proxy, the alert pops up repeatedly while the user types or pastes URLs. There is no de-dup or per-session suppression. The alert also runs from Dispatchers.IO; that happens to work because showAlert is just state mutation, but the call site doesn't make that contract obvious.

Pushed bugfix. Tested with microsocks for both the IPv6 bug and the auth proxy bug. No leaks besides the documented DNS one, same testing method as before. Tested for the alerts, didn't find any issues. Added a comment bc of weird java sdk21 behavior that might be confusing. All testing done on desktop, waiting for android testing for possible leaks

Narasimha-sc:
New concerns from claude review:
 Status of prior concerns                                                         
  
  #: 1                                                                             
  Concern: Substring proxy parser breaks on auth/IPv6                           
  Status: Fixed — uses typed appPrefs.networkProxy.get() with host/port fields  
    directly.                                                                   
  ────────────────────────────────────────                                      
  #: 2                                                                             
  Concern: SOCKS auth silently dropped                                          
  Status: Partially fixed — Authenticator.setDefault registered; see new concern A 
    below for residual security bug.                                               
  ────────────────────────────────────────                                      
  #: 3                                                                             
  Concern: Per-keystroke alert spam on parse failure                            
  Status: Fixed — alert and string-parsing path removed entirely.                  
  ────────────────────────────────────────                                      
  #: Duplicated showAlertDialogStacked block in PrivacySettings.kt and             
    NetworkAndServers.kt                                          
  Concern: Not fixed — out of this commit's scope.                                 
  Status:
  ────────────────────────────────────────                                         
  #: Lost "messages and file relays are connected via SOCKS" assurance in footer   
    string
  Concern: Not fixed.                                                              
  Status:                                                         
  ────────────────────────────────────────
  #: Localized MR/<lang>/strings.xml files not updated                             
  Concern: Not fixed.
  Status:                                                                          
                                                                  
  New concerns introduced by d392b3a37

  A. Credential leak to target HTTP servers via RequestorType.SERVER — security bug
  
  LinkPreviews.kt:54:                                                              
  if (requestorType == RequestorType.PROXY || requestorType ==    
  RequestorType.SERVER)                                       
    PasswordAuthentication(networkProxy.username,                                  
  networkProxy.password.toCharArray())           
                                                                                   
  The comment correctly notes that JDK's SocksSocketImpl calls    
  Authenticator.requestPasswordAuthentication with RequestorType.SERVER for SOCKS5 
  negotiation. But RequestorType.SERVER is also what HttpURLConnection uses when 
  the target HTTP server returns 401 (WWW-Authenticate challenge). Inside          
  getLinkPreview, the target is an arbitrary user-pasted URL — its server can
  elicit the SOCKS5 credentials via a 401 response.

  So: paste a URL pointing to a 401-gated CDN/site →
  Jsoup.connect(...).proxy(proxy).execute() triggers the link preview → target
  server returns 401 → JDK calls Authenticator.requestPasswordAuthentication with
  RequestorType.SERVER → our authenticator returns the SOCKS5 username/password →
  those credentials are base64-encoded into a Authorization: Basic ... header and
  sent to the target site.

  The image-fetch path (URL(imageUri).openConnection(proxy).getInputStream()) has
  the same issue.

  The Authenticator must filter on the host/port so it only answers for the proxy  
  itself. Minimum fix:
                                                                                   
  if ((requestorType == RequestorType.PROXY || requestorType ==   
  RequestorType.SERVER)
      && requestingHost.equals(networkProxy.host, ignoreCase = true)
      && requestingPort == networkProxy.port)

  requestingProtocol could be checked instead ("SOCKS5"), but host+port is more
  robust across JDK versions.
  
  Dead string left behind

  socks_proxy_invalid_address (strings.xml:1020) is no longer referenced anywhere —
   the only call site was removed. Leaving it in strings.xml adds translation churn
   for nothing. Delete it.

epoberezkin:
@iversonianGremling we will be completing and merging it - thanks a lot - aiming to land it into v6.5

epoberezkin:
Included in #6907

PR #6870, opened 2026-04-23
core: support chats in channels, send as owner in support chats

epoberezkin:


apps/ios/Shared/Views/Chat/Group/MemberSupportView.swift
```diff
@@ -45,7 +45,7 @@ struct MemberSupportView: View {
         : membersWithChats.filter { $0.wrapped.localAliasAndFullName.localizedLowercase.contains(s) }
 
         if membersWithChats.isEmpty {
-            Text("No chats with members")
+            Text(groupInfo.fullGroupPreferences.support.on ? "No chats with members" : "Chats with members are disabled")
```
spaced4ndy, line 48: already created chats stay - does it allow to send in them?

tests/ChatTests/Groups.hs
```diff
@@ -8399,6 +8402,117 @@ testScopedSupportMemberLeaves =
...
+                bob <## "Chat with admins: off",
+              do
+                cath <## "alice updated group #team: (signed) [>>]"
+                cath <## "updated group preferences:"
+                cath <## "Chat with admins: off"
+            ]
+
+          threadDelay 500000
+
+          -- cath can't send support (no existing chat)
```
spaced4ndy, line 8501: can alice send to cath with disabled preference? (she's owner)

spaced4ndy: I mean - can add to test

PR #6906, opened 2026-04-28
desktop: fix RTL text rendering under the send button

### LINKED ISSUE #4137: [Bug]: Right to left text renders under the send button on desktop

### Platform

Linux

### OS version

Arch Linux - Linux 6.8.9-arch1-1

### App version

5.7

### Current Behavior

When I enter RTL text, the text is shown under the send button, so the first few letters are not visible while typing.

Screenshot:
SimpleX_desktop_right_to_left

### Expected Behavior

The RTL text should be shown next to the send button.

Screenshot of correct behavior on Android:
SimpleX_android_right_to_left

### Steps To Reproduce

1. Change keyboard layout to any RTL language
2. Start typing

or simply paste some RTL text: `متن راست به چپ` 

### Relevant log output

_No response_

Narasimha-sc:
On desktop, RTL text typed under an LTR system locale rendered under the send button because the text field reserved padding on the wrong horizontal side. Fix: always reserve on the global layout direction's end — the same side `Alignment.BottomEnd` resolves the send button to in `SendMsgView`. Regression from #5051, which lifted the padding modifier out of the forced-RTL `CompositionLocalProvider` scope from #4675, so `start`/`end` started resolving against the global LTR direction instead of the inner forced-RTL one.

Fixes #4137.

### DIFF
--- a/apps/multiplatform/common/src/desktopMain/kotlin/chat/simplex/common/platform/PlatformTextField.desktop.kt
+++ b/apps/multiplatform/common/src/desktopMain/kotlin/chat/simplex/common/platform/PlatformTextField.desktop.kt
@@ -83,11 +83,12 @@ actual fun PlatformTextField(
     lastTimeWasRtlByCharacters.value = isRtlByCharacters
   }
   val isLtrGlobally = LocalLayoutDirection.current == LayoutDirection.Ltr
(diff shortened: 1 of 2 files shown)


natasha-grimes:
Can this get a little more attention please it is becoming kind of a deal breaker for me

PR #6938, opened 2026-05-03
android, ios, desktop: Flip camera when calling to mirror face

### LINKED ISSUE #6040: [Bug]: flip camera when calling to mirror face

### Platform

Android

### OS version

N/A

### App version

6.2.4

### Current Behavior

This happens in IOS too. When calling the camera is not flipping the view to show yourself in a mirrored way like all other messaging apps.

### Expected Behavior

The view should be flipped. 

### Steps To Reproduce

Call with video turned on

### Relevant log output

```shell

```

ed-asriyan:
fixes #6040

### DIFF
--- a/packages/simplex-chat-webrtc/src/android/style.css
+++ b/packages/simplex-chat-webrtc/src/android/style.css
@@ -158,3 +158,7 @@ body {
   display: none !important;
   -webkit-appearance: none !important;
 }
+
+.mirror {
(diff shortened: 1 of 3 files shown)


epoberezkin:
There would be quite a few people who prefer not mirrored, we can't really make half of the users happier by making another half less happy :)

An MVP of a net improvement would be an option in calls settings, which is a bigger change.

PR #6972, opened 2026-05-12
core, ui: allow indefinite deletion from history for public channel/group owners/moderators

epoberezkin:


apps/ios/Shared/Views/Chat/ChatView.swift
```diff
@@ -2175,8 +2175,15 @@ struct ChatView: View {
                 )
             }
                 .confirmationDialog("Delete message?", isPresented: $showDeleteMessage, titleVisibility: .visible) {
-                    Button("Delete for me", role: .destructive) {
-                        deleteMessage(.cidmInternal, moderate: false)
+                    let editorial = publicGroupEditorialRole(chat)
+                    if editorial {
```
epoberezkin:
```suggestion
                    if publicGroupEditorialRole(chat) {
```

apps/ios/Shared/Views/Chat/ChatView.swift
```diff
@@ -2963,6 +2976,13 @@ class FloatingButtonModel: ObservableObject {
 
 }
 
+private func publicGroupEditorialRole(_ chat: Chat) -> Bool {
+    if case let .group(groupInfo, _) = chat.chatInfo {
+        return groupInfo.useRelays && groupInfo.membership.memberRole >= .moderator
+    }
+    return false
```
epoberezkin:
```suggestion
    if case let .group(groupInfo, _) = chat.chatInfo {
        groupInfo.useRelays && groupInfo.membership.memberRole >= .moderator
    } else {
        false
    }
```

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -764,18 +765,26 @@ processChatCommand vr nm = \case
...
           assertUserGroupRole gInfo GRObserver -- can still delete messages sent earlier
           let msgIds = itemsMsgIds items
-              events = L.nonEmpty $ map (\msgId -> XMsgDel msgId Nothing $ toMsgScope gInfo <$> chatScopeInfo) msgIds
+              events = L.nonEmpty $ map (\msgId -> XMsgDel msgId Nothing (toMsgScope gInfo <$> chatScopeInfo) False) msgIds
+          mapM_ (sendGroupMessages user gInfo Nothing False recipients) events
+          delGroupChatItems user gInfo chatScopeInfo items False
+        CIDMHistory -> do
+          unless (publicGroupItemDeletable gInfo (membership gInfo)) $ throwChatError CEInvalidChatItemDelete
+          recipients <- getGroupRecipients vr user gInfo chatScopeInfo groupKnockingVersion
+          assertUserGroupRole gInfo GRObserver
```
spaced4ndy: owner only for history?

epoberezkin: this is redundant check, missed it. It's checked in 782

PR #6990, opened 2026-05-18
desktop: show unread count in window title

Narasimha-sc:
**What.** Adds a `SimpleX [N]` unread suffix to the desktop app's window/taskbar title when `N > 0`, where `N` counts unread messages across all profiles that the user would actually be notified about.

**Why.** The existing tray icon's dot is binary and the count lives only in the tray tooltip (requires hovering). The window/taskbar entry is the canonical surface for per-app unread counts — visible across workspaces, in every window switcher, and at zero attention cost.

**How.** A `derivedStateOf` in `AppWindow` (`DesktopApp.kt`) sums two parts and feeds the result into the `Window` composable's `title` parameter: the active profile walks `ChatModel.chats` with the same `enableNtfs` mute-mode dispatch that drives `Chat.unreadTag` (avoiding the drift in `users[active].unreadCount` whose increment skips the mute gate); other non-hidden, non-profile-level-muted profiles pass through `UserInfo.unreadCount`, which the core's SQL and the per-chat dispatcher gate at `SimpleXAPI.kt:2786-2789` both keep mute-filtered. Plans: `plans/2026-05-18-desktop-taskbar-unread-count-spec.md` and `plans/2026-05-18-desktop-taskbar-unread-count-impl.md`.


shumvgolove (approved):

epoberezkin:
Not sure I like this UX, it's confusing. Let's think about pros/cons, and let's discuss UX changes before making PRs.

PR #6999, opened 2026-05-21
android, desktop: strip metadata from sent images and videos

shumvgolove:


CherrieTheShifter:
Yes This Please 
We need less Metadata and such

PR #7011, opened 2026-05-25
android, desktop: fix E2E encryption section divider rendered inside card

Narasimha-sc:
## Bug
On the contact info screen, the "E2E encryption — Quantum resistant" card has a horizontal black line cutting through it followed by an empty padded area that looks like a second, empty card.

## Cause
`SectionDividerSpaced()` is called *inside* the `SectionView { ... }` block (`ChatInfoView.kt:622-625`), so the divider's `Divider` composable is drawn inside the card's `CardColumn` instead of as inter-section spacing. The stray call was left behind in `9e3f528d4` (removed the PQ toggle UI without moving the divider out), and the new card-based section design in `df5ea3d46` made it visually broken.

## Fix
Move `SectionDividerSpaced()` outside `SectionView { ... }`, matching the surrounding sections (e.g. `ChatInfoView.kt:614-618`).


Narasimha-sc:
Superseded by #7012 — that PR targets `stable` (where the bug ships in 6.5.3) and is branched from the same repo as `nd/fix-e2e-encryption-section-divider`.

PR #7036, opened 2026-06-01
core: signed XMember in public group

spaced4ndy:


src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -891,8 +891,16 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
           GCInviteeMember
             | isRelay m -> do
                 withStore' $ \db -> updateGroupMemberStatus db userId m GSMemConnected
-                gLink <- withStore $ \db -> getGroupLink db user gInfo
-                setGroupLinkDataAsync user gInfo gLink
+                case rosterVersion gInfo of
+                  Nothing -> do
```
spaced4ndy, line 907: not worth to special case here? channel without roster would be an exception from the rule, perhaps better to have unified protocol?

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -1590,19 +1595,35 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
           case gLinkInfo_ of
             Just GroupLinkInfo {groupId, memberRole = gLinkMemRole} -> do
               gInfo <- withStore $ \db -> getGroupInfo db vr user groupId
-              mem <- acceptGroupJoinRequestAsync user uclId gInfo invId chatVRange p Nothing (Just joiningMemberId) Nothing GAAccepted gLinkMemRole Nothing (Just joiningMemberKey)
+              existing_ <- withStore' $ \db -> eitherToMaybe <$> runExceptT (getGroupMemberByMemberId db vr user gInfo joiningMemberId)
+              case existing_ of
+                Just rosterMem
+                  -- a privileged memberId's key is owner-authoritative (the roster); the joiner must prove
+                  -- possession of that exact key, otherwise this is an attempt to impersonate it
+                  | memberRole' rosterMem >= GRModerator ->
```
spaced4ndy: use isRosterRole here

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -1590,19 +1595,35 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+                      if verifiedKeyPossession gInfo && memberPubKey rosterMem == Just joiningKey
```
spaced4ndy: can we reuse withVerifiedMsg here? extract some shared part between it and verifiedKeyPossession?

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -3172,13 +3193,26 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+                cached <-
+                  ( Right
+                      <$> withStore
+                        ( \db -> do
+                            res <- processRoster db defaultRole
+                            liftIO $ setCachedGroupRoster db gInfo newVer (groupMemberId' author) brokerTs sm
+                            pure res
+                        )
+                    )
+                    `catchAllErrors` (pure . Left)
```
spaced4ndy: make what we're catching a function

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -3172,13 +3193,26 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+                case cached of
+                  Right results -> do
+                    emitRosterResults results
+                    sendRosterAck author newVer Nothing
```
spaced4ndy: relay should only be sending roster ack when accepting request to serve channel, not each time. see comment below on xGrpRosterAck. Relay should probably check it's in RSAccepted status to send roster acknowledgement

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -3227,6 +3261,25 @@ processAgentMessageConn vr user@User {userId} corrId agentConnId agentMessage =
...
+    xGrpRosterAck gInfo m ackVer err = do
+      relay_ <- withStore' $ \db -> eitherToMaybe <$> runExceptT (getGroupRelayByGMId db (groupMemberId' m))
+      case relay_ of
+        Just relay@GroupRelay {relayStatus = RSAwaitingRoster} -> case err of
+          Nothing
+            | rosterVersion gInfo == Just ackVer -> do
+                withStore' $ \db -> void $ updateRelayStatus db relay RSAccepted
+                gLink <- withStore $ \db -> getGroupLink db user gInfo
+                setGroupLinkDataAsync user gInfo gLink
+            | otherwise -> messageWarning "x.grp.roster.ack: stale version, awaiting ack for the current roster"
```
spaced4ndy: relay should only ack roster on joining channel, as this only matters for newly added relays. Existing relays will have the roster when new members join, and new members join in observer role. consequently:

- owner wouldn't do setGroupLinkDataAsync on each roster update -> ack, only when new relay connects and accepts.
- I think RSAwaitingRoster is unnecessary, we can keep relay in status RSAccepted and check it on owner side.

if owner receives XGrpRosterAck from relay in status other than RSAccepted, simply ignore it, don't update link. E.g. it can happen with relay in status RSActive (on owner side) if relays itself hasn't yet checked link data to verify it's in list of relays, has its own status as RSAccepted and continues to send ack

src/Simplex/Chat/Types/Shared.hs
```diff
@@ -82,6 +82,7 @@ data RelayStatus
   = RSNew -- only for owner
   | RSInvited
   | RSAccepted
+  | RSAwaitingRoster -- connected, holding off publish until the relay confirms it cached the roster
```
spaced4ndy: as discussed above, it seems new status is not necessary

PR #7044, opened 2026-06-03
Recognise every QR code type in every scanner, with a clear wrong-type alert

Narasimha-sc:
Every QR scanner in the app now recognises every kind of code it could be shown — connection links, server addresses, database migration links and desktop addresses — instead of silently rejecting anything that isn't the kind that screen expects. When a code of the wrong kind is scanned, one clear alert tells the user exactly what they scanned and where to scan it, and offers to connect directly when it's a connection link. The behaviour is mirrored across Android, desktop and iOS.


shumvgolove (approved):

Narasimha-sc:
closing in favor of https://github.com/simplex-chat/simplex-chat/pull/7226

PR #7052, opened 2026-06-05
core: fix group link use after admin demotion

pbottine:
## Summary
- Revoke and block group-link operations once a local member is demoted below admin.

## Tests
- Attempted: `cabal test -j1 simplex-chat-test --test-show-details=direct --test-option=--match --test-option='demoted admin cannot keep using group link'`
- Blocked locally by the sandbox permission profile denying `*.key` fixture files during Cabal dependency checkout.


epoberezkin:
moved to #7111

PR #7055, opened 2026-06-05
android: preserve service restart state on destroy

pbottine:
## Summary
- Preserve `ServiceState.STARTED` during `SimplexService.onDestroy()` when the service is still configured to restart.
- Track explicit `safeStopService()` requests so intentional stops continue to persist `ServiceState.STOPPED`.
- Add unit coverage for the destroy-state decision.

## Tests
- `git diff --check origin/master..HEAD`
- Not run: `GRADLE_USER_HOME=/private/var/folders/xg/y64hd3px7h5_5qbwynrwx5xr0000gn/T/gradle-home-issue-4 ./gradlew :android:testDebugUnitTest --tests chat.simplex.app.SimplexServiceTest` because this local environment has no Java runtime installed.


pbottine:

PR #7064, opened 2026-06-09
core: fix member support chats staying marked unread after they're read

Narasimha-sc:
Members in a group's support ("chat with admins") list stay marked unread even after their support chat is fully read. Marking a group read without a scope (`updateGroupChatItemsRead`) flips support-scope items to read but never decrements the per-member `support_chat_items_*` counters, leaving them permanently out of sync. Restrict the no-scope group read (and its timed-items query) to main-scope items, so support chats are only read and decremented via their own scope.


shumvgolove (dismissed):

Narasimha-sc:
Superseded by #7281 — recreated on an upstream branch (this one was opened from a fork), rebased onto current master. Same two commits, no content changes.

PR #7072, opened 2026-06-13
android, desktop: fix chat preview and unread counter updating wrong chat

Narasimha-sc:
When a secondary chat view is open — "Chat with admins", "chat with member", or a group's reports view — an incoming message could be applied to the wrong chat in the main list: an unrelated chat took the message as its last-message preview, its unread badge incremented, and it popped to the top.

The cause is that `addChatItem` looked up the chat index in the secondary context's own list but read and wrote the primary list, so the index pointed at a different chat. This uses the receiver context's own list instead, reverting an unintended change from #5909 — and since a sent item only ever reaches the active context, it also dispatches sent items to the primary context explicitly, so a pending invitee's own support messages keep showing as the group's preview.

Details in `plans/chat-list-wrong-chat-preview-fix.md`.

### DIFF
--- a/apps/ios/Shared/Model/ChatModel.swift
+++ b/apps/ios/Shared/Model/ChatModel.swift
@@ -1109,13 +1109,10 @@ final class ChatModel: ObservableObject {
     func changeUnreadCounter(_ chatIndex: Int, by count: Int, unreadMentions: Int) {
         let wasUnread = chats[chatIndex].unreadTag
         let stats = chats[chatIndex].chatStats
-        // clamp at 0 so an over-decrement can't drive the per-chat count or the user badge negative
-        // (e.g. deleting an item that never incremented the badge); feed the badge the clamped delta
-        let newUnreadCount = max(stats.unreadCount + count, 0)
-        chats[chatIndex].chatStats.unreadCount = newUnreadCount
-        chats[chatIndex].chatStats.unreadMentions = max(stats.unreadMentions + unreadMentions, 0)
+        chats[chatIndex].chatStats.unreadCount = stats.unreadCount + count
+        chats[chatIndex].chatStats.unreadMentions = stats.unreadMentions + unreadMentions
         ChatTagsModel.shared.updateChatTagRead(chats[chatIndex], wasUnread: wasUnread)
-        changeUnreadCounter(user: currentUser!, by: newUnreadCount - stats.unreadCount)
+        changeUnreadCounter(user: currentUser!, by: count)
     }
 
     func increaseUnreadCounter(user: any UserLike) {
--- a/apps/ios/Shared/Model/ChatModel.swift
+++ b/apps/ios/Shared/Model/ChatModel.swift
@@ -661,7 +661,11 @@ final class ChatModel: ObservableObject {
                 chats[i].chatItems = switch cInfo {
                 case .group:
                     if let currentPreviewItem = chats[i].chatItems.first {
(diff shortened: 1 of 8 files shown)


plans/2026-06-13-fix-wrong-chat-preview-addchatitem.md
```diff
@@ -0,0 +1,100 @@
...
+
+`getChatIndex` resolves against the receiver context (`this.chats`), but
+`chatsContext` (unqualified) always resolves to `ChatModel.chatsContext` — the
+**primary** list. On the secondary context the two lists hold different chats in a
+different order, so `i` (a secondary-list index) points at an unrelated chat in the
+primary list, which then gets clobbered.
+
+### How it was introduced
+Commit `b97e1e0f1` (#5909, "show new messages from support scope in main chat preview
+when invitee is pending") changed `chats[i]` → `chatsContext.chats[i]` at four sites
```
spaced4ndy: was it tested that this isn't broken?

Narasimha-sc: just tested, broken

shumvgolove (dismissed):

apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
```diff
@@ -533,14 +538,18 @@ object ChatModel {
...
           val newPreviewItem = when (cInfo) {
             is ChatInfo.Group -> {
               val currentPreviewItem = chat.chatItems.firstOrNull()
               if (currentPreviewItem != null) {
-                if (cItem.meta.itemTs >= currentPreviewItem.meta.itemTs) {
+                // Pending invitee: surface the latest support message (broker vs local itemTs aren't
+                // comparable); don't let a no-content event re-cover an already-shown message.
+                if (cInfo.groupInfo_?.membership?.memberPending == true) {
+                  if (cItem.content.msgContent != null || currentPreviewItem.content.msgContent == null) cItem else currentPreviewItem
+                } else if (cItem.meta.itemTs >= currentPreviewItem.meta.itemTs) {
```
epoberezkin: this is very confusing. It uses 3 different conditions for what is effectively 2 possibilities - show cItem or show currentPreviewItem - so needs to be simplified.

There are 4(!) conditional branches for only two possibilities - show cItem or show currentPreviewItem. We want to change condition when either is shown, but it doesn't mean that we need to double the number of conditions.

apps/ios/Shared/Model/ChatModel.swift
```diff
@@ -661,7 +661,11 @@ final class ChatModel: ObservableObject {
                 chats[i].chatItems = switch cInfo {
                 case .group:
                     if let currentPreviewItem = chats[i].chatItems.first {
-                        if cItem.meta.itemTs >= currentPreviewItem.meta.itemTs {
+                        // Pending invitee: surface the latest support message (broker vs local itemTs
+                        // aren't comparable); don't let a no-content event re-cover an already-shown message.
+                        if cInfo.groupInfo?.membership.memberPending ?? false {
+                            (cItem.content.msgContent != nil || currentPreviewItem.content.msgContent == nil) ? [cItem] : [currentPreviewItem]
+                        } else if cItem.meta.itemTs >= currentPreviewItem.meta.itemTs {
```
epoberezkin: same problem as below - 4 conditions for 2 possibilities

shumvgolove (dismissed):

PR #7085, opened 2026-06-16
android, desktop, ios: document missing formatting in markdown help

Narasimha-sc:
**Problem:** The "How to use" markdown help screen documented only some of the supported formatting codes. **Fix:** Added the missing small text (`!- text!`) and link (`[text](https://…)`) formats, with translations, on Android, desktop and iOS. **Cause:** These two formats were never added to the help screen when they were introduced to the parser.


Narasimha-sc:
Superseded by #7086 (same change from a branch in the upstream repo instead of a fork).

PR #7086, opened 2026-06-16
android, desktop, ios: document missing markdown formats in help and make examples interactive

Narasimha-sc:
The "How to use" markdown help didn't list all formats and its examples weren't interactive. It now documents small text (`!- text!`) and links (`[text](url)`), opens the link via the standard "Open external link?" prompt, and reveals secret text on tap — matching how they behave in chat.
fixmissingformatting


shumvgolove (dismissed):

shumvgolove (approved):

epoberezkin:
this makes UI very overloaded for the sake of completeness. I'd park it, and if anything is done, we could reduce wha't shown to bold, color and secret (that's what most people would need), and move the full list onto a separate page, as then we can extend it further (and it will extend).

PR #7100, opened 2026-06-19
ios: open SimpleX links in chat messages via in-app connect flow

Narasimha-sc:
## Problem

On iOS (reproduced on iPhone 17, v6.5.2 and v6.5.5), tapping a **SimpleX connection/invitation link inside message text** does nothing — it never reaches the connection flow. Tapping web links (open in browser) and the connection-link **card** both work, so only inline SimpleX links appear dead.

## Root cause

Inline links are dispatched in `MsgContentView.handleTextTaps`:

- web links (`webLinkAttrKey`) → `openBrowserAlert` (Safari)
- everything else → `UIApplication.shared.open(url)`

SimpleX links fell into the second branch. iOS drops `UIApplication.shared.open` for a URL that the **same** app owns while it is in the foreground — both the `simplex:` scheme and the `simplex.chat` universal links are registered to this app. So the tap was silently swallowed: `onOpenURL` never fired and `connectViaUrl` was never reached.

`mailto:`/`tel:` use the same branch but target **other** apps (Mail/Phone), so they keep working — which is why the failure looked specific to SimpleX links.

The connection-link card never had this bug because it calls `planAndConnect` in-process (`FramedItemView`), and the multiplatform clients do the same via `openVerifiedSimplexUri` → `connectIfOpenedViaUri` → `planAndConnect`. iOS inline links were the only path relying on an OS round-trip.

## Fix

Restore the three-way dispatch that the multiplatform clients use (`WEB_URL` / `OTHER_URL` / `SIMPLEX_URL`):

- web → `openBrowserAlert` (unchanged)
- `mailto:`/`tel:` → `UIApplication.shared.open` (unchanged, other apps)
- **SimpleX → `ChatModel.appOpenUrl`** — the same sink `onOpenURL` feeds, leading to `connectViaUrl` → `planAndConnect`, with no OS round-trip

SimpleX links are identified by a dedicated attribute key set on the `.simplexLink` format (mirroring the multiplatform `SIMPLEX_URL` tag), rather than sniffing the URL string, so all link types (contact, invitation, group, channel, relay) and both URI schemes are covered.

This is the same mechanism already used by `ShareSheet.openExternalLink`, which routes SimpleX links to `appOpenUrl`.

Also fixes the same problem for the **"Send questions and ideas"** (Settings) and **"connect to SimpleX Chat developers"** (chat help) buttons, which open `simplexTeamURL` (a `simplex:` link) the same broken way.

## Scope

`MsgContentView.swift`, `SettingsView.swift`, `ChatHelp.swift` — 19 insertions, 5 deletions. No behavior change for web/mailto/tel links.

## Testing

Needs a build + device check (the dispatch is iOS-version-dependent). Verify: tapping an inline SimpleX invitation/contact link in a message opens the connection sheet, and the two developer-contact buttons open the connect flow.

### DIFF
--- a/apps/ios/Shared/Views/Chat/ChatItem/MsgContentView.swift
+++ b/apps/ios/Shared/Views/Chat/ChatItem/MsgContentView.swift
@@ -191,9 +191,14 @@ private func handleTextTaps(
                         }
                     }
                 }
-                if let index, let (uri, browser) = attributedStringLink(s, for: index) {
+                if let index, let (uri, browser, simplex) = attributedStringLink(s, for: index) {
                     if browser {
                         openBrowserAlert(uri: uri)
+                    } else if simplex, let url = URL(string: uri) {
+                        // SimpleX links target this same app (simplex: scheme / simplex.chat universal link),
+                        // so UIApplication.shared.open is dropped by iOS while the app is in the foreground.
+                        // Route to the in-app connect flow instead (same sink onOpenURL feeds).
+                        ChatModel.shared.appOpenUrl = url
                     } else if let url = URL(string: uri) {
                         UIApplication.shared.open(url)
                     } else {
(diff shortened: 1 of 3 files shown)


Narasimha-sc:
Superseded by #7101, which is based on a branch in the main repo instead of a fork.

PR #7103, opened 2026-06-19
desktop: fix in-app updater deleting the download before "Open file location"

Narasimha-sc:
**Problem:** the in-app updater's "Download completed" dialog could open an empty folder with the downloaded file gone, and on some systems (e.g. Whonix) the update couldn't proceed at all.

**Cause:** `downloadAsset` ignored the boolean result of `file.renameTo(newFile)`, so when that rename failed the bytes were left at the temp path and `createTmpFileAndDelete`'s `finally { tmpFile.delete() }` deleted the only copy.

**Fix:** use `Files.move(..., REPLACE_EXISTING)`, which does the same in-place rename when possible, falls back to copy otherwise, and throws real failures into the existing `catch` instead of silently losing the download.

Full root-cause analysis, compatibility notes, and regression reasoning: `plans/2026-06-19-fix-updater-open-file-location.md`.


Narasimha-sc:
Superseded by #7104 — same change from a branch on this repository instead of a fork.

PR #7109, opened 2026-06-20
nodejs: fix apiChatItemReaction checking wrong response type

Mirza-Samad-Ahmed-Baig:
## Problem

`ChatApi.apiChatItemReaction` in `packages/simplex-chat-nodejs/src/api.ts` checks the response against `"chatItemsDeleted"` and returns `r.chatItemDeletions`:

```ts
const r = await this.sendChatCmd(CC.APIChatItemReaction.cmdString({chatRef: {chatType, chatId}, chatItemId, add, reaction}))
if (r.type === "chatItemsDeleted") return r.chatItemDeletions
throw new ChatCommandError("error setting item reaction", r)
```

Those values were copy-pasted from `apiDeleteMemberChatItem` directly above it. The `/_reaction` command (`APIChatItemReaction`) **never** returns `chatItemsDeleted` — its success response is `CR.ChatItemReaction` (`type: "chatItemReaction"`, field `reaction: T.ACIReaction`), as declared by `APIChatItemReaction.Response` in `commands.ts`.

### Impact
- The method **always throws** `ChatCommandError("error setting item reaction")`, even when adding/removing a reaction succeeds.
- `r.chatItemDeletions` does not exist on the reaction response, and the method had no return type annotation.

## Fix

Match on `"chatItemReaction"` and return `r.reaction`, typed `Promise<T.ACIReaction>`:

```ts
): Promise<T.ACIReaction> {
  const r = await this.sendChatCmd(CC.APIChatItemReaction.cmdString({chatRef: {chatType, chatId}, chatItemId, add, reaction}))
  if (r.type === "chatItemReaction") return r.reaction
  throw new ChatCommandError("error setting item reaction", r)
}
```

This matches:
- `APIChatItemReaction.Response = CR.ChatItemReaction | CR.ChatCmdError` in `packages/simplex-chat-client/types/typescript/src/commands.ts`
- the `CR.ChatItemReaction` interface (`type: "chatItemReaction"`, `reaction: T.ACIReaction`) in `responses.ts`
- the Python client's `api_chat_item_reaction`, which already returns `r["reaction"]` on `type == "chatItemReaction"`

## Notes
- One-line behavioural change plus a return-type annotation; no API surface change for callers that were (necessarily) catching the spurious error.

### DIFF
--- a/packages/simplex-chat-nodejs/src/api.ts
+++ b/packages/simplex-chat-nodejs/src/api.ts
@@ -498,10 +498,10 @@ export class ChatApi {
     chatItemId: number,
     add: boolean,
     reaction: T.MsgReaction
-  ) {
+  ): Promise<T.ACIReaction> {
     const r = await this.sendChatCmd(CC.APIChatItemReaction.cmdString({chatRef: {chatType, chatId}, chatItemId, add, reaction}))
-    if (r.type === "chatItemsDeleted") return r.chatItemDeletions
-    throw new ChatCommandError("error setting item reaction", r)  
+    if (r.type === "chatItemReaction") return r.reaction
+    throw new ChatCommandError("error setting item reaction", r)
   }
 
   /**


Mirza-Samad-Ahmed-Baig:

saschadaemgen:
Independently reproduced this while load-testing the Node SDK with an embedded core. Confirming it is still present in **6.5.4** (current `latest`) and **7.0.0-beta.3** (current `beta`).

**Symptom:** adding a reaction always throws, but the reaction *is* applied. Removing one returns normally. That asymmetry makes it look like the emoji was rejected when it wasn't.

**Reproduction:** send a message, then `apiChatItemReaction(ChatType.Group, groupId, itemId, true, {type: "emoji", emoji: "\u{1F44D}"})`. The call throws `ChatCommandError("error setting item reaction")`, and reading the chat back afterwards shows the reaction present:

```
reactions: [{"reaction":{"type":"emoji","emoji":"👍"},"userReacted":true,"totalReacted":1}]
```

Sending the same command via `sendChatCmd(CC.APIChatItemReaction.cmdString(...))` returns exactly what this PR describes:

```
response type: chatItemReaction
added: true
```

**One extra detail that cost us some debugging time:** the thrown `ChatCommandError` carries the successful response on `.response`, and `.chatError` is `undefined`. Error handlers that inspect `.chatError` (as most do, since that is what `ChatAPIError` uses) therefore log an empty error, which makes this look like an emoji-validation problem rather than a response-type mismatch. Worth mentioning in case anyone hits the same dead end.

**Workaround until this lands:** bypass the wrapper and accept both success shapes.

```ts
const r = await chat.sendChatCmd(
  CC.APIChatItemReaction.cmdString({chatRef: {chatType, chatId}, chatItemId, add, reaction})
)
if (r.type === "chatItemReaction" || r.type === "chatItemsDeleted") return r
```

With the fix applied on our side, all eight reactions the clients offer (👍 👎 😀 😂 😢 ❤ 🚀 ✅) are accepted by the core, and everything outside that set is correctly rejected with `commandError`. So the core behaviour is fine; it is purely the response check in the wrapper.

PR #7115, opened 2026-06-22
core, ui: message signing

spaced4ndy:


src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -2275,15 +2279,22 @@ processAgentMessageConn cxt user@User {userId} corrId agentConnId agentMessage =
...
-            CChatItem SMDRcv ci@ChatItem {chatDir = CIChannelRcv, meta = CIMeta {itemLive}, content = CIRcvMsgContent oldMC}
-              | maybe True (\m -> memberRole' m == GROwner) m_ -> updateCI True ci scopeInfo oldMC itemLive Nothing
+            CChatItem SMDRcv ci@ChatItem {chatDir = CIChannelRcv, meta = CIMeta {itemLive, msgSigned = itemSigned}, content = CIRcvMsgContent oldMC}
+              | maybe True (\m -> memberRole' m == GROwner) m_ -> requireVerifiedMutation (CDChannelRcv gInfo scopeInfo) itemSigned $ updateCI True ci scopeInfo oldMC itemLive Nothing
               | otherwise -> messageError "x.msg.update: member attempted to update channel message" $> Nothing
             _ -> messageError "x.msg.update: invalid message update" $> Nothing
           where
             isSender m' = maybe False (\m -> sameMemberId (memberId' m) m') m_
+            -- only a verified mutation may alter a verified item; reject otherwise (fail-closed)
+            requireVerifiedMutation :: ChatDirection 'CTGroup 'MDRcv -> Maybe MsgSigStatus -> CM (Maybe DeliveryTaskContext) -> CM (Maybe DeliveryTaskContext)
```
spaced4ndy: rename requireVerifiedEdit

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -2383,6 +2397,19 @@ processAgentMessageConn cxt user@User {userId} corrId agentConnId agentMessage =
           | senderRole < GRModerator || senderRole < memberRole =
               messageError "x.msg.del: message of another member with insufficient member permissions" $> Nothing
           | otherwise = a
+        -- only a verified delete may remove a verified item; reject otherwise (fail-closed)
+        requireVerifiedDelete :: CChatItem 'CTGroup -> CM (Maybe DeliveryTaskContext) -> CM (Maybe DeliveryTaskContext)
```
spaced4ndy, line 2414: requireVerifiedMutation doesn't create RGEMsgBadSignature chat item and requireVerifiedDelete does. should be aligned - both to create item

tests/ChatTests/Groups.hs
```diff
@@ -12069,6 +12076,397 @@ testChannelMemberMessageDelete ps =
...
+            cathMsgId <- lastItemId cath
+            cath #$> ("/_delete item #1 " <> cathMsgId <> " broadcast", id, "message marked deleted")
+            bob <# "#team cath> [marked deleted] secret"
+            concurrentlyN_
+              [ alice <# "#team cath> [marked deleted] secret",
+                dan <# "#team cath> [marked deleted] secret",
+                eve <# "#team cath> [marked deleted] secret"
+              ]
+  where
+    memberIdByName :: TestCC -> T.Text -> IO MemberId
```
spaced4ndy: memberIdByName, relayConnIdToMember, itemSharedMsgId must be reused across new tests

spaced4ndy:
> **One blocking issue (iOS):** `MsgVerified` uses Swift's synthesized `Decodable`, which does not match the core's `"type"`-tagged `sumTypeJSON` encoding. Since `CIMeta.msgVerified` is always present and non-optional, this breaks decoding of every `CIMeta`, i.e. no items load on iOS. Fix staged (custom `init(from:)` matching the file's convention). The plan already lists iOS as in-progress, so this is expected but must be resolved before merge. See the inline suggestion for the related forward-compat note (non-optional `msgVerified` on `CIMeta`).

Blocking point is a false positive — verified on iOS that items load fine; MsgVerified uses the same sumTypeJSON/synthesized-decode path as CIStatus, not the flat "type" shape.

PR #7125, opened 2026-06-23
android, desktop: keep wide images at natural aspect ratio without crashing

Narasimha-sc:
**Problem:** Very wide images (e.g. 4000×1) used to crash the chat with a `Constraints` overflow; the merged fix (#7123) stopped the crash by clamping the framed preview's aspect ratio at 2.33, but that reshapes every image wider than 2.33:1 to 2.33:1.

**Cause:** The box derived its height via `Modifier.aspectRatio`, which during measurement computes `width = height × ratio` and overflows for extreme ratios — so the only safe way to keep `aspectRatio` was to clamp it.

**Fix:** Compute the box height directly (`height = w × min(h / w, 2.33)`), removing the overflow-prone derivation entirely. Wide images now keep their natural ratio (no upper clamp), tall images stay capped at 2.33, and it matches how iOS already sizes previews. Plan updated under `plans/`.

### DIFF
--- a/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/item/CIImageView.kt
+++ b/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/item/CIImageView.kt
@@ -176,9 +176,6 @@ fun CIImageView(
       .then(
         if (!smallView) {
           val w = if (previewBitmap.width * 0.97 <= previewBitmap.height) imageViewFullWidth() * 0.75f else DEFAULT_MAX_IMAGE_WIDTH
-          // Derive the height directly instead of via aspectRatio: a very wide image keeps its natural ratio
-          // (no upper clamp) while taller images are capped at 2.33. aspectRatio would instead derive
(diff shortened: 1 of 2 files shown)


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/item/CIImageView.kt
```diff
@@ -176,7 +176,10 @@ fun CIImageView(
       .then(
         if (!smallView) {
           val w = if (previewBitmap.width * 0.97 <= previewBitmap.height) imageViewFullWidth() * 0.75f else DEFAULT_MAX_IMAGE_WIDTH
-          Modifier.width(w).aspectRatio((previewBitmap.width.toFloat() / previewBitmap.height.toFloat()).coerceIn(1f / 2.33f, 2.33f))
+          // Derive the height directly instead of via aspectRatio: a very wide image keeps its natural ratio
+          // (no upper clamp) while taller images are capped at 2.33. aspectRatio would instead derive
+          // width = height * ratio during measurement and overflow Constraints for extreme ratios (e.g. 4000x1).
```
epoberezkin:
```suggestion
```

PR #7156, opened 2026-06-28
desktop, android: fix file chat item meta layout

spaced4ndy:


rajpratham1 (commented):
This PR fixes an actual UI consistency issue by passing the full metadata into CIFileView instead of only an edited flag, allowing the file metadata reservation logic to account for timestamps, TTL, proxy status, encryption, and other indicators. The implementation is clean, updates all call sites, and doesn't introduce unnecessary complexity.

PR #7182, opened 2026-07-01
Reply to messages from notifications [android, desktop, iOS] 

### LINKED ISSUE #2437: Feature: add a notification action for a quick reply
Please add a "reply" [action](https://developer.android.com/reference/android/app/Notification.Action) to new message notifications. For reference, this feature exists in Signal messenger. It allows click a "Reply" button attached to a push notification, and type a reply into a text box that appears, all without the need to open the messenger itself.

This feature is handy for productivity, especially when used with smart watches. Without this feature, I am able no able to send quick replies to incoming messages, since the moment I transitioned from Signal to Simplex.

larrasket:
Closes #2437 

Adds an inline "Reply" action to new message notifications so you can answer without opening the app (like Signal). Useful in general and especially on smartwatches.

Android's message notifications now use MessagingStyle with a RemoteInput reply action. 

Desktop uses the `two-slices` inline reply (input()) where the notification backend supports it, and falls back to focusing the window and chat where it doesn't. I added this capability in two-slices recentlt: https://github.com/sshtools/two-slices/pull/8

iOS adds a `UNTextInputNotificationAction` to the message category. The NSE can't send, so the typed reply is handled by the app launched in the background.

New "Reply from notifications" setting in notification settings. It's on by default; when app lock (SimpleX Lock / passcode) is enabled, replying from notifications is off unless you turn on "Allow when locked".

### DIFF
--- a/apps/ios/Shared/Model/NtfManager.swift
+++ b/apps/ios/Shared/Model/NtfManager.swift
@@ -15,6 +15,7 @@ import SimpleXChat
 let ntfActionAcceptContact = "NTF_ACT_ACCEPT_CONTACT"
 let ntfActionAcceptCall = "NTF_ACT_ACCEPT_CALL"
 let ntfActionRejectCall = "NTF_ACT_REJECT_CALL"
+let ntfActionReply = "NTF_ACT_REPLY"
 
 private let ntfTimeInterval: TimeInterval = 1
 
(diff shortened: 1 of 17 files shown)


larrasket:

larrasket:
recheck

larrasket:
Friendly ping @epoberezkin. Can we get this reviewed any time soon?

PR #7183, opened 2026-07-01
ios, android, desktop: show member labels in groups

### LINKED ISSUE #7178: [Feature]: Member labels in groups — surface member self-description as a visible label
## User problem

In groups — especially larger or more active ones — it's hard to tell who someone is or what their role is at a glance; you only see a display name. Other apps (e.g. Signal's member labels) let a member add a short self-description / role — "Organizer", "she/her", "Support", "Maintainer" — that's visible to the group. This helps people orient in a group without asking, and lets members signal pronouns/role/context.

## What already exists in SimpleX

SimpleX profiles (and group-member profiles) already carry `shortDescr` — a short bio limited to 160 characters (`src/Simplex/Chat/Types.hs`, `data Profile`). It's part of each member's `memberProfile`, so it's already shared **end-to-end with the group, with no server involvement**, via the existing profile propagation. It's already editable in the UI (e.g. `GroupProfileView.kt`) and shown on profile detail.

So the underlying mechanism for a per-member, group-visible, self-set description already exists.

## Proposed change (the gap)

Surface this as a member **label** within group views:

- Show a member's self-description inline in the **group member list** and member info — not only when you tap into their profile.
- Optionally add a dedicated, compact **label** field (e.g. ≤24 chars) distinct from the 160-char bio, so it renders cleanly as an inline chip next to the name. (Alternatively, just truncate `shortDescr`.)
- Per-group scoping fits naturally: members already have per-group profiles (incognito / per-group profile), so a label can differ per group.

Because the data path already exists, this is largely a UI/surfacing change with low protocol/DB cost. It's self-set (members control their own profile, consistent with SimpleX's model), and older clients ignore unknown/extra profile fields gracefully.

## Notes

Raising per the "discuss early" contributing guideline before any PR. Happy to scope this to "surface the existing `shortDescr` as a member label" first (smallest change), or add a dedicated compact label field if the team prefers that. Prior art: Signal member labels.

michaelsam94:
## Summary
- show each group member profile `shortDescr` as a one-line label in Android/Desktop group member rows
- show the same one-line label in iOS group and channel member rows
- leave existing member detail description rendering unchanged

Fixes #7178

## Tests
- `git diff --check`
- `./gradlew :common:compileKotlinMetadata --dry-run` (fails during configuration: Android SDK location is not configured via `ANDROID_HOME` or `apps/multiplatform/local.properties`)
- `xcodebuild -list -project apps/ios/SimpleX.xcodeproj`
- `xcodebuild -project apps/ios/SimpleX.xcodeproj -scheme "SimpleX (iOS)" -configuration Debug -sdk iphonesimulator -destination "generic/platform=iOS Simulator" CODE_SIGNING_ALLOWED=NO build` (interrupted after package targets and `SimpleXChat` compilation began; only pre-existing Swift warnings were observed before interruption)

### DIFF
--- a/apps/ios/Shared/Views/Chat/Group/ChannelMembersView.swift
+++ b/apps/ios/Shared/Views/Chat/Group/ChannelMembersView.swift
@@ -59,12 +59,19 @@ struct ChannelMembersView: View {
                 .font(.caption).baselineOffset(2).kerning(-2)
                 .foregroundColor(theme.colors.secondary) + nameText
             : nameText
+        let shortDescr = member.shortDescr?.trimmingCharacters(in: .whitespacesAndNewlines)
         let row = HStack {
             MemberProfileImage(member, size: 38)
                 .padding(.trailing, 2)
             VStack(alignment: .leading) {
                 NameWithBadge(displayName, member.nameBadge)
                     .lineLimit(1)
+                if let shortDescr = shortDescr, shortDescr != "" {
+                    Text(shortDescr)
+                        .font(.caption)
+                        .foregroundColor(theme.colors.secondary)
+                        .lineLimit(1)
+                }
                 if user {
                     Text("you")
                         .font(.caption)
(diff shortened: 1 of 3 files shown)


michaelsam94:

michaelsam94:
recheck

PR #7192, opened 2026-07-02
SIMPLEX theme: sunrise gradient on chat bubbles, avatars and text

another-simple-pixel:
The SIMPLEX theme now paints the whole chat over one soft "sunrise" gradient — dark at the bottom-left rising to warm at the top-right — instead of flat panels. Bubbles, avatars and links pick up the gradient by where they sit on screen, so the chat reads as one continuous surface. Dates, delivery ticks, names and other small text switch to a warm, slightly-transparent tint that stays readable at any height over the gradient. Error and deleted-message bubbles get the same gradient as normal received messages. The gradient follows a per-chat wallpaper if you set one, and previews correctly in the wallpaper picker.

### DIFF
--- a/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
+++ b/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/model/ChatModel.kt
@@ -4850,7 +4850,7 @@ sealed class Format {
   val isSimplexLink = this is SimplexLink
 
   companion object {
-    val linkStyle @Composable get() = SpanStyle(color = MaterialTheme.colors.primary, textDecoration = TextDecoration.Underline)
+    val linkStyle @Composable get() = SpanStyle(color = LocalSimplexLinkColor.current ?: MaterialTheme.colors.primary, textDecoration = TextDecoration.Underline)
   }
 }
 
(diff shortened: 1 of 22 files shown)


another-simple-pixel:
A few things worth flagging for review:

- Based on the wallpapers scope (#7186) — the base branch is `ae/simplex-wallpapers`, so this diff is only the chat-rendering layer.

- Three changes are visible in every theme, not only SIMPLEX (all small improvements — happy to gate them if you'd rather keep this SIMPLEX-only):
  1. event / feature-change notices now follow the active theme instead of a colour that was frozen at start-up (this also fixes them not updating on a live theme switch);
  2. the "sent via proxy" arrow in the meta row now matches the white meta row on image/video messages (it was grey);
  3. the decryption / integrity-error and deleted / marked-deleted bubble wrappers changed from `Surface` to `Box` + gradient background — pixel-identical, one minor delta: a tap on the two non-clickable "deleted" bubbles no longer blocks touch-through.

- The second commit makes links a static accent colour and default avatars a flat semi-transparent tint, rather than sampling the gradient by on-screen position. That removes the per-frame link recomposition while scrolling and the per-avatar offscreen compositing layer — only bubbles still sample the axis now.

- Known follow-up, not in this PR: the wallpaper-picker preview shows the dark variant for SIMPLEX — that lives in the wallpapers layer, I'll fix it in #7186.

PR #7199, opened 2026-07-04
fix: use single-pane layout on narrow desktop windows

### LINKED ISSUE #7191: [Bug]: Scaling Broken in Desktop Application

### Platform

Linux

### OS version

postmarketOS edge

### App version

6.5.5

### Current Behavior

Using the desktop application on a phone (as the OS is Linux, not Android) causes the entire screen to be either filled with the Chats list (with a tiny section on the right side of the screen for the active chat) or split 50:50/60:40 with tiny UI elements (barely visible, and extremely difficult to tap).

There should be some sort of scaling for when Linux is used on small-display devices. Proposals I have for this feature are **a)** to allow the Chats section on the left side of the screen to be hidden via a button in the signal-status bar, and **b)** to make the SimpleX application share the appearance of the Android and iOS applications when a small display is detected.

Without this feature, using SimpleX on small-display devices is not possible.

### Expected Behavior

The SimpleX Chats and active-chat screens should be better-scaled for small displays, whether by hiding the Chats section on the left side of the screen, or making it more like the Android and iOS applications when a small display is detected.

### Steps To Reproduce

1. Open SimpleX application on Linux on a device with a small display, such as a phone
2. Observe the tiny UI elements _without_ increased scaling, and the off-screen active chat _with_ increased scaling

### Relevant log output

```shell

```

harshjainnn:
## Summary
Fixes broken UI scaling on small-display Linux devices (e.g. phones running postmarketOS). The desktop app always used `DesktopScreen`'s fixed two/three-column layout with a hardcoded 388.dp chat-list column, regardless of actual window width — resulting in either an unusable 50/50 split or a barely-tappable chat list on narrow screens.

Fixes #7191

## Changes
- In `App.kt`, wrapped the screen-selection logic in `BoxWithConstraints` to read the current window width
- Changed the layout condition from `appPlatform.isAndroid` to `appPlatform.isAndroid || maxWidth < 600.dp`, so narrow desktop windows now fall back to `AndroidScreen`'s responsive single-pane layout instead of `DesktopScreen`
- No changes to `AndroidScreen` or `DesktopScreen` internals — only which one gets selected based on window width

## Testing
- Ran the Linux desktop build in a window resized below 600.dp width — chat list and chat view now stack in a single pane with proper animated transitions, instead of splitting into tiny side-by-side columns
- Resized the window back above 600.dp — confirmed `DesktopScreen`'s normal multi-column layout still renders correctly
- Verified on actual Android build that behavior is unchanged (still always uses `AndroidScreen`)

### DIFF
--- a/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/App.kt
+++ b/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/App.kt
@@ -164,12 +164,16 @@ fun MainScreen() {
             }
           }
           SetupClipboardListener()
-          if (appPlatform.isAndroid) {
-            AndroidWrapInCallLayout {
+          BoxWithConstraints {
+            if (appPlatform.isAndroid) {
+              AndroidWrapInCallLayout {
+                AndroidScreen(userPickerState)
+              }
+            } else if (maxWidth < 600.dp) {
               AndroidScreen(userPickerState)
+            } else {
+              DesktopScreen(userPickerState)
             }
-          } else {
-            DesktopScreen(userPickerState)
           }
         }
       }


harshjainnn:

harshjainnn:
recheck

PR #7200, opened 2026-07-04
core: use channel link for SimpleX Status preset contact

Narasimha-sc:
**Problem:** The SimpleX Status preset contact used an old long `simplex:/contact` link on smp4.simplex.im.

**Cause:** `simplexStatusContactProfile.contactLink` was hardcoded to the outdated full contact-request URI.

**Fix:** Replace it with the current short channel link `https://smp5.simplex.im/c#…`, which decodes to a `CLShort` channel contact link.


Narasimha-sc:
Superseded by #7231 — the channel link can't work on a preset contact card (contact cards connect via the direct-contact handshake; channel links require APIConnectPreparedGroup), so the preset is removed entirely instead.

PR #7202, opened 2026-07-04
SIMPLEX theme: darker chat list and fainter dividers

another-simple-pixel:
On the SIMPLEX theme the chat list now sits on a darker panel, and the hairline dividers across the app — chat list, chat, message input and top bar — are fainter, matching the theme's dark, low-contrast look. Every other theme is unchanged.


another-simple-pixel:
Independent scope — base is master, not stacked on the other SIMPLEX PRs.

Non-SIMPLEX themes are byte-identical: the new `dividerColor()` helper's else branch is Material's own default (`onSurface` at 0.12 alpha), and the darker chat-list background is applied on SIMPLEX only. The divider change covers the chrome frame dividers — chat list, chat, compose bar, top app bar.

PR #7203, opened 2026-07-05
Refresh the chat wallpaper patterns (larger, even density)

another-simple-pixel:
The six chat wallpaper patterns are refreshed with new artwork — larger, and with an even density across all six. The pattern is a single shape tinted per theme, so this updates the wallpaper look on every theme.


another-simple-pixel:
Cross-theme asset update — not SIMPLEX-specific. The pattern is a shared shape coloured per theme (`ColorFilter.tint`), so the new artwork plus the unified scale (0.5) refresh the wallpaper look on Light / Dark / SIMPLEX / Black alike.

Independent of the SIMPLEX scope PRs (#7184/#7185/#7186/#7192/#7202): the wallpapers scope (#7186) deliberately kept the master patterns, so there's no overlap — this is the separate end-of-series assets PR.

PR #7211, opened 2026-07-06
core: don't re-approve directory channels on stale publicGroupId

Narasimha-sc:
## Problem

After an owner adds a SimpleX name (blockchain domain claim) to a public channel, the SimpleX Directory service re-approves the channel **every ~30 minutes, indefinitely**, with no changes on the owner's side (it keeps flipping to "hidden until approved").

## Root cause

The directory's periodic link check re-approves whenever `updateGroupFromLinkData` reports the profile changed. That flag was a full structural comparison `p /= groupProfile` (`Simplex.Chat.Library.Internal`), which includes `publicGroupId` — an **immutable identity field that `updateGroupProfile` never re-syncs** (it's written only at group creation and by `updateRelayGroupKeys`).

So if the directory's stored `public_group_id` drifts from the link's, the comparison is true on **every** link check and can never converge, re-approving forever. Adding a name is merely the occasion that surfaces the pre-existing mismatch. By elimination, `publicGroupId` is the only `Eq`-relevant field `updateGroupProfile` doesn't persist, so it's the only field that can produce a permanent mismatch.

## Fix

Normalize `publicGroupId` out of the change comparison in `updateGroupFromLinkData` (used for both the store guard and the returned flag), so only fields `updateGroupProfile` actually manages drive the change signal. Mirrors the existing `sameGroupProfileInfo` / `clearProofs` content-comparison idiom.

One-field change; no schema, wire-format, or API change. The flag is consumed only by the directory service.

## Testing

Adds `testLinkCheckStalePublicGroupId` (`tests/Bots/DirectoryTests.hs`): register + approve a channel, corrupt the directory's stored `public_group_id` so it differs from the link's, run a link check, and assert the channel stays listed.

- Passes with the fix.
- **Fails without** the fix — the link check emits `The channel ID 1 (news) profile changed` (the exact re-approval symptom) — confirming the test guards the bug.

`simplex-directory-service` and `simplex-chat-test` both build.

Design notes and the full root-cause analysis are in `plans/2026-07-06-fix-directory-name-reapproval.md`.

### DIFF
--- a/plans/2026-07-06-fix-directory-name-reapproval.md
+++ b/plans/2026-07-06-fix-directory-name-reapproval.md
@@ -74,6 +74,19 @@ is permanent → `groupUpdated` true every cycle → re-approval on repeat.
 is merely the *occasion* (the owner republishing the link) that surfaces the
 pre-existing `publicGroupId` staleness.
 
+The claim's round-trip fidelity — the one field that *changes* when a name is
+added, hence the last assumption behind the elimination — was verified: `ToField
+SimplexDomain = decodeLatin1 . strEncode` and `FromField = strDecode . encodeUtf8`
+(`simplexmq SimplexName.hs:127-129`); domains are ASCII, so the stored value is
+`strDecode (strEncode d)`, which equals `d` for a canonical (lower-cased) domain,
+and the link's domain is already canonical (it arrives via `StrJSON`/`strDecode`).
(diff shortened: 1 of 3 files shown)


Narasimha-sc:
Superseded by #7213 (same-repo branch instead of fork).

PR #7212, opened 2026-07-07
Refresh the built-in chat wallpapers (light, dark, black)

another-simple-pixel:
The built-in chat wallpapers on the light, dark and black themes are refreshed to look cleaner and read more easily.

Incoming and outgoing messages are easier to tell apart: incoming bubbles get the most contrast for comfortable reading, while your own bubbles carry the theme's color — deeper and richer on the dark and black themes — so the two sides look clearly different at a glance.

The background patterns are redrawn larger and more even, so the wallpaper sits calmly behind the chat and text stays easy to read over it — crisp on both phone and desktop.


another-simple-pixel:
Notes for review:
- Two commits that belong together — the patterns and the colors are tuned to each other, so please keep them as a pair.
- The colors are oklch and look best on top of the oklch→Display P3 conversion (#7184); suggested order is that conversion first, then this PR, then the SIMPLEX‑theme PRs.
- One follow-up I'll push to this branch: move the desktop-only wallpaper images out of the shared resources into the desktop source set.

PR #7222, opened 2026-07-08
ios: keep desktop remote session alive in background

lukasa1993:
## Summary
- keep verified desktop remote-control sessions connected when the iOS app backgrounds by skipping normal chat suspension and running a bounded keepalive
- preserve normal mobile usage: dismiss the connected desktop sheet, keep explicit Disconnect as the stop path, and show a small chat-list indicator that opens the desktop connection screen
- keep mobile UI current while desktop is connected by forwarding allowed remote events to both local and desktop queues; stop desktop if the remote event queue fills instead of leaving stale state

## Implementation notes
- `RemoteCtrlBGKeepAlive` starts only from the iOS background scene-phase path and stops on active without disconnecting desktop
- iOS 26+ uses `BGContinuedProcessingTaskRequest` with a wildcard permitted identifier and falls back to `UIApplication.beginBackgroundTask`
- manual disconnect and expiration are serialized through one helper so `stopRemoteCtrl()` is not called concurrently
- normal suspend/background-refresh behavior is unchanged when no desktop is connected

## Testing
- `git diff --check`
- `plutil -lint apps/ios/SimpleX--iOS--Info.plist`
- `cabal test simplex-chat-test --test-options='--match "Remote/With compression/disconnects desktop when remote event queue is full"'`
- attempted iOS simulator build with `xcodebuild -project apps/ios/SimpleX.xcodeproj -scheme 'SimpleX (iOS)' -destination 'platform=iOS Simulator,name=iPhone 17 Pro' -configuration Debug build`; it reached Swift compilation and then failed at the existing local missing Haskell simulator static library `HSsimplex-chat-7.0.0.8...ghc9.6.3`

## Notes
- The iOS 26 system-visible continued-processing activity only starts when the app backgrounds. Starting it immediately on desktop connect was intentionally not kept because this change is scoped to background hold, not foreground behavior.

### DIFF
--- a/apps/ios/Shared/Views/Chat/ChatView.swift
+++ b/apps/ios/Shared/Views/Chat/ChatView.swift
@@ -47,7 +47,7 @@ struct ChatView: View {
     @Environment(\.colorScheme) var colorScheme
     @Environment(\.presentationMode) var presentationMode
     @Environment(\.scenePhase) var scenePhase
-    @State @ObservedObject var chat: Chat
+    @State var chat: Chat
     @ObservedObject var im: ItemsModel
     @State var mergedItems: BoxedValue<MergedItems>
     @State var floatingButtonModel: FloatingButtonModel
@@ -93,6 +93,18 @@ struct ChatView: View {
     @AppStorage(DEFAULT_TOOLBAR_MATERIAL) private var toolbarMaterial = ToolbarMaterial.defaultMaterial
 
     let userSupportScopeInfo: GroupChatScopeInfo = .memberSupport(groupMember_: nil)
+
+    private var selectedChatItemsCount: Int {
+        selectedChatItems?.count ?? 0
+    }
+
+    private var deleteSelectedMessagesTitle: LocalizedStringKey {
+        selectedChatItemsCount == 1 ? "Delete message?" : "Delete \(selectedChatItemsCount) messages?"
+    }
+
+    private var archiveSelectedReportsTitle: LocalizedStringKey {
+        selectedChatItemsCount == 1 ? "Archive report?" : "Archive \(selectedChatItemsCount) reports?"
+    }
 
     // Spec: spec/client/chat-view.md#body
     var body: some View {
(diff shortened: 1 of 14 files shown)


lukasa1993:

lukasa1993:
recheck

apps/ios/Shared/Views/Chat/ChatView.swift
```diff
@@ -923,15 +954,13 @@ struct ChatView: View {
...
-                        let maxWidth = cInfo.chatType == .group
-                        ? channelReceivedNoAvatar
-                        ? g.size.width - 26
-                        : voiceNoFrame || channelReceived
-                        ? (g.size.width - 28) - 42
-                        : (g.size.width - 28) * 0.84 - 42
-                        : voiceNoFrame
-                        ? (g.size.width - 32)
-                        : (g.size.width - 32) * 0.84
+                        let maxWidth = chatItemMaxWidth(
```
epoberezkin, line 957: how is it related to remote session?

apps/ios/Shared/Views/Chat/ChatView.swift
```diff
@@ -94,6 +94,18 @@ struct ChatView: View {
...
+        selectedChatItems?.count ?? 0
+    }
+
+    private var deleteSelectedMessagesTitle: LocalizedStringKey {
+        selectedChatItemsCount == 1 ? "Delete message?" : "Delete \(selectedChatItemsCount) messages?"
+    }
+
+    private var archiveSelectedReportsTitle: LocalizedStringKey {
+        selectedChatItemsCount == 1 ? "Archive report?" : "Archive \(selectedChatItemsCount) reports?"
+    }
```
epoberezkin, line 107: seems unrelated to remote session

apps/ios/Shared/Views/Chat/ChatView.swift
```diff
@@ -418,25 +430,7 @@ struct ChatView: View {
...
-                        im.chatState.clear()
-                        chatModel.groupMembers = []
-                        chatModel.groupMembersIndexes.removeAll()
-                        chatModel.membersLoaded = false
-                        ChannelRelaysModel.shared.reset()
-                    }
-                }
-            }
-        }
+        .onDisappear(perform: handleChatViewDisappear)
```
epoberezkin, line 433: seems unnecessary refactoring

PR #7226, opened 2026-07-09
Recognise wrong QR code type in every scanner

Narasimha-sc:
**Problem:** Each QR scanner accepts only one kind of code and rejects everything else with a generic "invalid" message, so scanning a valid code on the wrong screen gives no hint that the code is fine, just scanned in the wrong place.

**Cause:** Each scanner runs only its own single parser and has no notion of the other code types.

**Fix:** On a parse failure each scanner now falls back to a shared classifier that tries the other known parsers (connection link, server address, migration link, desktop address) and shows "This is a X — scan it here", or the generic message when nothing matches. Mirrored on Android, desktop and iOS; plan in `plans/2026-07-09-qr-recognise-wrong-type.md`.


Narasimha-sc:
Superseded by #7269 — the head branch was renamed from `sh/qr-recognise-wrong-type` to `nd/qr-recognise-wrong-type` (correct prefix), which auto-closed this PR.

PR #7234, opened 2026-07-10
ios: keep desktop session alive in background

lukasa1993:
- keep a verified remote-desktop session alive while the iOS app is backgrounded using iOS background audio with a silent looping track
- show a Live Activity when available and keep the desktop sheet presented while connected
- stop the keepalive and restore normal chat suspension when the remote session ends

`BGContinuedProcessingTask` was tested but was not reliable for this flow.


lukasa1993:
BGContinuedProcessingTaskRequest submission belongs to the explicit foreground user action, which is the successful verification path. remoteCtrlConnected is currently only a redundant event for that command and may be delivered outside that foreground boundary, so it should not submit the task. If this event ever becomes an independent connection path, we’ll add a separate foreground initiation flow; meanwhile it falls back to the legacy task on background.

lukasa1993:
Registration timing: I confirmed this against the iOS 26.5 SDK. BGTaskScheduler explicitly exempts BGContinuedProcessingTask registrations from the before-launch requirement. register(...) also returns Bool rather than throwing, and the permitted identifier wildcard is present in the plist. A real-device smoke test is still useful, but lazy registration is supported here.
Expiration cleanup: Agreed that this is a valid risk. stopRemoteCtrl() ultimately performs synchronous controller cancellation and connection cleanup, so it could consume the short expiration window before the system task is completed. Simply completing first is also unsafe because the app may suspend before remote teardown, database suspension, and state cleanup finish. Given the fallback semantics, I’m leaving this unchanged unless a bounded teardown mechanism is required.
Duplicate disconnect logic: Agreed. The paths have slightly different failure/UI behavior, and extracting them would broaden this PR. I’ll leave the duplication here to follow the zero-unrelated-changes policy.

apps/ios/Shared/Views/RemoteAccess/ConnectDesktopView.swift
```diff
@@ -145,7 +148,6 @@ struct ConnectDesktopView: View {
                 mkAlert(title: title, message: error)
             }
         }
-        .interactiveDismissDisabled(m.activeRemoteCtrl)
```
epoberezkin, line 148: why are we allowing dismissing this sheet here? We didn't allow using mobile concurrently anyway, so dismissing sheet would show a wrong UI.

PR #7252, opened 2026-07-14
Bump cli docs to avoid ambiguity

Scott-Simmons:
Solves for Cabal-7132 error, see below:

```sh
Error: [Cabal-7132] Ambiguous target 'simplex-chat'. It could be: exe:simplex-chat (component) lib:simplex-chat (component)
```

```sh
cabal --version
cabal-install version 3.12.1.0
compiled using version 3.12.1.0 of the Cabal library
```


Scott-Simmons:

PR #7268, opened 2026-07-17
ios, desktop, android: clearer error message when connecting via removed link (AUTH)

Narasimha-sc:
**Problem:** connecting via a deleted or already-used link shows "Connection error (AUTH)" and suggests it "might be a bug - please report it", which confuses new users and invites bug reports for expected behavior.

**Cause:** AUTH from the server on connect means the link's queue no longer exists — the contact removed the link, or a one-time link was already used — but the message hedges instead of saying so.

**Fix:** replace the alert on iOS, Android and desktop with "Connection link removed" and a definitive explanation with a clear next step (ask the contact for a new link); string values only, translation keys unchanged.


apps/ios/Shared/Model/SimpleXAPI.swift
```diff
@@ -1104,8 +1104,8 @@ private func apiConnectResponseAlert<R>(_ r: APIResult<R>) async {
             )
         case .errorAgent(.SMP(_, .AUTH)):
             showAlert(
-                NSLocalizedString("Connection error (AUTH)", comment: ""),
-                message: NSLocalizedString("Unless your contact deleted the connection or this link was already used, it might be a bug - please report it.\nTo connect, please ask your contact to create another connection link and check that you have a stable network connection.", comment: "")
+                NSLocalizedString("Connection link removed", comment: ""),
+                message: NSLocalizedString("This link no longer works: your contact removed it, or it was a one-time link that was already used.\nTo connect, ask your contact to create a new link.", comment: "")
```
epoberezkin:
```suggestion
                message: NSLocalizedString("Your contact removed this link, or it was a one-time link that was already used.\nTo connect, ask your contact to create a new link.", comment: "")
```

PR #7284, opened 2026-07-20
core: prohibit to change or assign relay role

spaced4ndy:


src/Simplex/Chat/Library/Commands.hs
```diff
@@ -2884,6 +2884,8 @@ processChatCommand cxt nm = \case
       when (length memberIds > 1 && (anyAdmin || newRole >= GRAdmin)) $
         throwCmdError "can't change role of multiple members when admins selected, or new role is admin"
       when anyPending $ throwCmdError "can't change role of members pending approval"
+      when (any (\m -> isNothing $ roleRequiredToChange (memberRole' m) newRole) (invitedMems <> currentMems <> unchangedMems)) $
```
spaced4ndy: instead of additional pass through the list, return anyRelay from selectMembers fold; also throw if target (new) role is relay. can be single condition with OR and a common message

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -1292,10 +1292,13 @@ isPrivilegedRole :: GroupMemberRole -> Bool
 isPrivilegedRole r = r >= GRMember
 
 -- Minimum role allowed to change a member's role from `from` to `to` (moderators only within observer..member).
-roleRequiredToChange :: GroupMemberRole -> GroupMemberRole -> GroupMemberRole
+-- Nothing when no role may: relay is established from the relay link when the member is created, and in channels
+-- it alone gates forwarding (isMemberGrpFwdRelay) while sorting below GRObserver, so `maximum` would not guard it.
+roleRequiredToChange :: GroupMemberRole -> GroupMemberRole -> Maybe GroupMemberRole
```
spaced4ndy: revert all changes to this function, delete moderatable from where, change moderator condition to:
`from <= GRMember && to <= GRMember`
relay checks we'll do outside

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -3382,6 +3384,8 @@ processAgentMessageConn cxt user@User {userId} corrId agentConnId agentMessage =
                   else pure (gInfo', m)
               toView CEvtMemberRole {user, groupInfo = gInfo'', byMember = m', member = member {memberRole = memRole}, fromRole, toRole = memRole, msgSigned}
               pure $ memberEventDeliveryScope member
+          where
```
spaced4ndy: revert

src/Simplex/Chat/Library/Subscriber.hs
```diff
@@ -3363,7 +3363,9 @@ processAgentMessageConn cxt user@User {userId} corrId agentConnId agentMessage =
         -- applyMember writes the change (role, or role + pinned key for a freshly TOFU-created member);
         -- the delivery scope (relay forwarding) is computed on the pre-change role
         changeMemberRole gInfo' member@GroupMember {memberRole = fromRole} created applyMember gEvent createItem
-          | senderRole < roleRequiredToChange fromRole memRole =
+          | isNothing requiredRole =
```
spaced4ndy: revert check, only check that member role is not relay; also add check that new role is not relay before as the first guard of the function - before `| membershipMemId == memId =`

tests/ChatTests/Groups.hs
```diff
@@ -279,6 +280,8 @@ chatGroupTests = do
       it "should preserve owner key in link data after profile update" testChannelOwnerKeyAfterLinkUpdate
       it "should update channel preferences (signed)" testChannelUpdatePrefsSigned
       it "should change member role (signed)" testChannelChangeRoleSigned
+      it "should not change relay role" testChannelRelayRoleNotChangeable
```
spaced4ndy: revert all changes to this module, no need for new tests here

PR #7293, opened 2026-07-23
core: prohibit file and media content in groups where files are not allowed, check content of updated messages

Narasimha-sc:
Images can be sent to a group that has "Files and media" disabled, e.g. `/_send #1 json [{"msgContent":{"type":"image","text":" ","image":"data:image/jpg;base64,..."}}]` — the sender's client raises no error and recipients display the image. `prohibitedGroupContent` decided the rule from the presence of a file, but `MCImage`/`MCVideo` carry the preview inline and `MCFile` is file content by itself, so with no file attached the check never fired; the update path was unchecked in the same way, on both the send and the receive side.

The fix tests the content type instead, the way voice messages already are, covering `/_send`, `x.msg.new` and `x.msg.update` through the shared check, on both sides. Editing only the text of a message that already has media or voice is allowed (the media is compared, not just the content type), so captions can still be edited after the feature is disabled, while replacing the media is not. An update is checked with the scope of the item being updated, not the scope of the command, so a main chat message cannot be edited with a support chat reference to skip the scope-gated guards. The same voice hole in direct chats, where the update path had no content check, is closed the same way.

The SimpleX links commit removes the update path's own links check, which applied in all scopes on both sides while the same guard inside `prohibitedGroupContent` is scope-gated — so a link could be posted in a support chat but not added by editing a message there. Nothing stops being validated: links in updates are now judged by the same scope-gated guard as links in new messages.

History items with expired or cancelled files are sent without a file invitation, so an older media item would be received as prohibited content by a member joining a group that has since disabled the feature; such items are now sent as text, and a signed item whose media cannot be changed is dropped from history instead.

Justification, alternatives considered and known effects are in `plans/2026-07-23-fix-group-media-without-file.md`. `testProhibitFiles`, `testGroupPrefsSimplexLinksForRole`, `testGroupHistoryProhibitedMedia` and `testProhibitVoiceUpdate` cover sending and receiving, with the receiving side driven by resetting preferences in the sender's database to emulate a client that does not check them; each behaviour has a negative control that fails when its fix is reverted.

### DIFF
--- a/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ComposeView.kt
+++ b/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ComposeView.kt
@@ -1630,9 +1630,11 @@ fun ComposeView(
       ReportReasonView(ctx.reason)
     }
 
+    // when editing, the media is already in the message and only its text is changed,
+    // so file and voice prohibitions don't apply (a link can be added by editing text, so it still does)
     val simplexLinkProhibited = chatsCtx.secondaryContextFilter == null && hasSimplexLink.value && !chat.groupFeatureEnabled(GroupFeature.SimplexLinks)
-    val fileProhibited = chatsCtx.secondaryContextFilter == null && composeState.value.attachmentPreview && !chat.groupFeatureEnabled(GroupFeature.Files)
-    val voiceProhibited = composeState.value.preview is ComposePreview.VoicePreview && !chat.chatInfo.featureEnabled(ChatFeature.Voice)
+    val fileProhibited = chatsCtx.secondaryContextFilter == null && composeState.value.attachmentPreview && !composeState.value.editing && !chat.groupFeatureEnabled(GroupFeature.Files)
+    val voiceProhibited = composeState.value.preview is ComposePreview.VoicePreview && !composeState.value.editing && !chat.chatInfo.featureEnabled(ChatFeature.Voice)
     val disableSendButton = simplexLinkProhibited || fileProhibited || voiceProhibited
     if (composeState.value.preview !is ComposePreview.VoicePreview || composeState.value.editing) {
       if (simplexLinkProhibited) {
--- a/src/Simplex/Chat/Library/Commands.hs
+++ b/src/Simplex/Chat/Library/Commands.hs
@@ -763,6 +763,8 @@ processChatCommand cxt nm = \case
         CChatItem SMDSnd ci@ChatItem {meta = CIMeta {itemSharedMsgId, itemTimed, itemLive, editable}, content = ciContent} -> do
           case (ciContent, itemSharedMsgId, editable) of
             (CISndMsgContent oldMC, Just itemSharedMId, True) -> do
+              when (isVoice mc && not (sameMedia mc oldMC) && not (featureAllowed SCFVoice forUser ct)) $
+                throwCmdError $ "feature not allowed " <> T.unpack (chatFeatureNameText CFVoice)
               let changed = mc /= oldMC
               if changed || fromMaybe False itemLive
                 then do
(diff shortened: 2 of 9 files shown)


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/chat/ComposeView.kt
```diff
@@ -1630,9 +1630,11 @@ fun ComposeView(
       ReportReasonView(ctx.reason)
     }
 
+    // when editing, the media is already in the message and only its text is changed,
+    // so file and voice prohibitions don't apply (a link can be added by editing text, so it still does)
     val simplexLinkProhibited = chatsCtx.secondaryContextFilter == null && hasSimplexLink.value && !chat.groupFeatureEnabled(GroupFeature.SimplexLinks)
-    val fileProhibited = chatsCtx.secondaryContextFilter == null && composeState.value.attachmentPreview && !chat.groupFeatureEnabled(GroupFeature.Files)
-    val voiceProhibited = composeState.value.preview is ComposePreview.VoicePreview && !chat.chatInfo.featureEnabled(ChatFeature.Voice)
+    val fileProhibited = chatsCtx.secondaryContextFilter == null && composeState.value.attachmentPreview && !composeState.value.editing && !chat.groupFeatureEnabled(GroupFeature.Files)
+    val voiceProhibited = composeState.value.preview is ComposePreview.VoicePreview && !composeState.value.editing && !chat.chatInfo.featureEnabled(ChatFeature.Voice)
```
epoberezkin, line 1637: this is unnecessary, we don't allow editing voice.

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -782,39 +784,40 @@ processChatCommand cxt nm = \case
...
-                    else pure $ CRChatItemNotChanged user (AChatItem SCTGroup SMDSnd (GroupChat gInfo chatScopeInfo) ci)
-                _ -> throwChatError CEInvalidChatItemUpdate
-            CChatItem SMDRcv _ -> throwChatError CEInvalidChatItemUpdate
+      -- TODO [knocking] check chat item scope?
+      (cci, itemScopeInfo) <- withFastStore $ \db ->
+        (,) <$> getGroupCIWithReactions db user gInfo itemId <*> getGroupChatScopeInfoForItem db cxt user gInfo itemId
+      case cci of
+        CChatItem SMDSnd ci@ChatItem {meta = CIMeta {itemSharedMsgId, itemTimed, itemLive, editable, showGroupAsSender, msgVerified}, content = ciContent} -> do
+          case (ciContent, itemSharedMsgId, editable) of
+            -- the content is checked with the scope of the updated item, not with the scope of the command
```
epoberezkin, line 795: they must be the same, the command should just fail if they are different

src/Simplex/Chat/Library/Internal.hs
```diff
@@ -339,15 +339,17 @@ quoteContent mc qmc ciFile_
...
     qTextOrFile = if T.null qText then qFileName else qText
 
-prohibitedGroupContent :: GroupInfo -> GroupMember -> Maybe GroupChatScopeInfo -> MsgContent -> Maybe MarkdownList -> Maybe f -> Bool -> Maybe GroupFeature
-prohibitedGroupContent gInfo@GroupInfo {membership = mem@GroupMember {memberRole = userRole}} m scopeInfo mc ft file_ sent
+prohibitedGroupContent :: GroupInfo -> GroupMember -> Maybe GroupChatScopeInfo -> MsgContent -> Maybe MsgContent -> Maybe MarkdownList -> Maybe f -> Bool -> Maybe GroupFeature
+prohibitedGroupContent gInfo@GroupInfo {membership = mem@GroupMember {memberRole = userRole}} m scopeInfo mc oldMC_ ft file_ sent
   | not supportAllowed = Just GFSupport
-  | isVoice mc && not (groupFeatureMemberAllowed SGFVoice m gInfo) && not hostApprovalVoice = Just GFVoice
-  | isNothing scopeInfo && not (isVoice mc) && isJust file_ && not (groupFeatureMemberAllowed SGFFiles m gInfo) = Just GFFiles
+  | isVoice mc && newMedia && not (groupFeatureMemberAllowed SGFVoice m gInfo) && not hostApprovalVoice = Just GFVoice
```
epoberezkin, line 345: also unnecessary, we don't allow editing voice

PR #7294, opened 2026-07-23
desktop: copy the caption of an image instead of the image

Narasimha-sc:
Copy on an image or file that has a caption puts the file path on the clipboard instead of the caption text, so pasting it back into the message box re-attaches the file instead of inserting the text (reported on Windows, but the same on macOS and Linux). The desktop `copyItemToClipboard` checks the attached file before the item text, so an item that has both can never have its text copied. The file path is now used only when the item has no text — the rule iOS applies, and Android reaches the same outcome by always copying the text.


epoberezkin:
How is copying the text better than copying the image? The text you can copy via info or select.

PR #7295, opened 2026-07-23
desktop: fix chat switch deleting the in-app updater's download

Narasimha-sc:
The desktop in-app updater's download reaches 100% and then nothing happens — no "App update is downloaded" dialog, no error, and the downloaded file is gone.

`downloadAsset` downloads into `createTmpFileAndDelete`, which registers its file in `ChatModel.filesToDelete`. `ComposeView.deleteUnusedFiles()` empties that set on chat open/switch/close, so switching chats during the multi-minute download deletes the file; `copyTo` keeps writing through the still-open fd so progress still reaches 100%, and only the following `Files.move` fails with `NoSuchFileException`, into a catch that logs to stdout only.

Fixed as suggested in review: the updater no longer uses `createTmpFileAndDelete`. It downloads into `<asset name>.part` in `tmpDir` and moves it onto the asset name when the transfer completes, so the file is never registered and chat-scoped cleanup cannot reach it. Cleanup is unchanged in substance — the `finally` removes the partial file when the download fails or is cancelled, `deleteOnExit()` covers a clean exit, `Main.kt` wipes `tmpDir` at startup after a crash, and a leftover `.part` is truncated by the next download.

`createTmpFileAndDelete` keeps its previous behaviour; the only change there is a comment warning that its file does not survive a chat switch.

The first commit removed the registration from the helper and the second reverts that and fixes the updater instead, so the squashed diff is the updater change plus that comment.

Rationale, the check of every other call site, the alternatives, and a follow-up left out of this PR (the registration is redundant with the helper's own `finally`, and it mutates a non-thread-safe set from IO threads) are in `plans/2026-07-23-fix-updater-download-deleted-by-chat-switch.md`.

### DIFF
--- a/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/helpers/Utils.kt
+++ b/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/helpers/Utils.kt
@@ -389,14 +389,13 @@ fun removeWallpaperFilesFromAllChats(user: User) {
   }
 }
 
-// Not registered in ChatModel.filesToDelete: that set is emptied on every chat switch
-// (ComposeView.deleteUnusedFiles), which would delete this file while [onCreated] is still
-// using it - it deleted the in-app updater's download mid-transfer. The finally block below
-// already removes the file, so the registration only ever duplicated that.
+// The created file is registered in ChatModel.filesToDelete, which ComposeView.deleteUnusedFiles()
+// empties on chat switch - do not use this helper for a file that has to survive a chat switch.
 fun <T> createTmpFileAndDelete(dir: File = tmpDir, onCreated: (File) -> T): T {
   val tmpFile = File(dir, UUID.randomUUID().toString())
   tmpFile.parentFile.mkdirs()
   tmpFile.deleteOnExit()
+  ChatModel.filesToDelete.add(tmpFile)
   try {
     return onCreated(tmpFile)
   } finally {
--- a/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/helpers/Utils.kt
+++ b/apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/helpers/Utils.kt
@@ -389,8 +389,6 @@ fun removeWallpaperFilesFromAllChats(user: User) {
   }
 }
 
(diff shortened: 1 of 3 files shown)


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/helpers/Utils.kt
```diff
@@ -389,11 +389,14 @@ fun removeWallpaperFilesFromAllChats(user: User) {
...
 
+// Not registered in ChatModel.filesToDelete: that set is emptied on every chat switch
+// (ComposeView.deleteUnusedFiles), which would delete this file while [onCreated] is still
+// using it - it deleted the in-app updater's download mid-transfer. The finally block below
+// already removes the file, so the registration only ever duplicated that.
 fun <T> createTmpFileAndDelete(dir: File = tmpDir, onCreated: (File) -> T): T {
   val tmpFile = File(dir, UUID.randomUUID().toString())
   tmpFile.parentFile.mkdirs()
   tmpFile.deleteOnExit()
-  ChatModel.filesToDelete.add(tmpFile)
```
epoberezkin, line 396: I don't think it's a correct fix - the correct fix is to not use createTmpFileAndDelete for downoaded file.

Narasimha-sc: Done in 5a15496ff — the updater no longer uses `createTmpFileAndDelete`. It downloads into `<asset name>.part` in `tmpDir` and `Files.move`s it onto the asset name when the transfer completes; the `finally` deletes the partial file if the download fails or is cancelled (and is a no-op after a successful move, as the helper's own `finally` was). The helper itself is back to its previous behaviour, with a comment warning that its file does not survive a chat switch.

Two things I did not include, in case you want them as a follow-up:

- the registration is redundant at all 6 remaining call sites — none of them lets the file outlive the lambda, so the helper's own `finally` has already deleted it before `deleteUnusedFiles()` could reach it;
- `createTmpFileAndDelete` adds to `ChatModel.filesToDelete` (a plain `mutableSetOf`) from whatever thread calls it, including `Dispatchers.IO` (settings save, `saveFileFromUri`), while `deleteUnusedFiles()` iterates that set on the UI thread — a concurrent `add` during that `forEach` throws `ConcurrentModificationException`.

PR #7333, opened 2026-08-02
directory: support listing businesses and service bots

epoberezkin:


apps/simplex-directory-service/src/Directory/Store/Postgres/Migrations.hs
```diff
@@ -50,3 +51,29 @@ DROP INDEX idx_sx_directory_group_regs_owner_contact_id_user_group_reg_id;
...
 DROP TABLE sx_directory_group_regs;
   |]
+
+m20260801_directory_contact_regs :: Text
+m20260801_directory_contact_regs =
+  T.pack
+    [r|
+CREATE TABLE sx_directory_contact_regs(
+  contact_reg_id BIGINT PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  contact_id BIGINT REFERENCES contacts(contact_id) ON UPDATE RESTRICT ON DELETE CASCADE,
```
epoberezkin, line 61:
```suggestion
  contact_id BIGINT REFERENCES contacts(contact_id) ON UPDATE RESTRICT ON DELETE CASCADE, -- nullable, to allow extension to business chats and non-owned addresses
```

PR #7337, opened 2026-08-03
ui: highlight chat list filter button while any chat is unread

Narasimha-sc:
The unread filter button in the chat list looked identical whether or not anything was unread, so the only way to find out was to tap it and see whether the list emptied. This tints the filter icon with the accent colour while any chat is unread — same icon, same size, only the colour — on Android, desktop and iOS.

It matches on `Chat.unreadTag` plus the `!chatDeleted && !contactCard` guard that `filteredChats` applies, i.e. exactly the chats the unread filter itself lists, so the button cannot claim unread chats the filter would not show. Justification and rejected alternatives in `plans/2026-08-03-unread-filter-button-highlight.md`.


Narasimha-sc:
Superseded by #7340, which is opened from a branch in this repository rather than from a fork. Same commits (52c7f22).

PR #7348, opened 2026-08-05
ui: connect when a SimpleX link is pasted as a markdown hyperlink

Narasimha-sc:
Pasting a short SimpleX link written as a markdown hyperlink — `[label](https://smp6.simplex.im/a#...)` — into chat list search, new chat search, or "Tap to paste link" fails with "Invalid connection link" instead of connecting.

The parser returns one fragment whose format is `SimplexLink` but whose `text` is the whole `[label](link)` source, and `strConnectTarget` passes that `text` to the connect plan, which the core rejects. It now uses the already-parsed `simplexUri` when `showText` is set — an exact marker of the hyperlink parser — so every bare-link path is unchanged.

### DIFF
--- a/apps/ios/Shared/Views/NewChat/NewChatView.swift
+++ b/apps/ios/Shared/Views/NewChat/NewChatView.swift
@@ -859,8 +859,6 @@ enum ConnectTarget {
 func strConnectTarget(_ str: String) -> ConnectTarget? {
     let parsedMd = parseSimpleXMarkdown(str)
     let links = parsedMd?.filter { $0.format?.isSimplexLink ?? false } ?? []
-    // showText is set only for a link written as a markdown hyperlink, whose text is the whole
-    // "[label](link)" source rather than the link - connect via the parsed uri in that case
     return if links.count == 1, case let .simplexLink(showText, linkType, simplexUri, smpHosts) = links[0].format {
         .link(text: showText != nil ? simplexUri : links[0].text, linkType: linkType, linkText: simplexLinkText(linkType, smpHosts))
     } else if links.isEmpty,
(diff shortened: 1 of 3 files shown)


apps/multiplatform/common/src/commonMain/kotlin/chat/simplex/common/views/newchat/NewChatView.kt
```diff
@@ -836,7 +836,10 @@ fun strConnectTarget(str: String): ConnectTarget? {
   val links = parsedMd.filter { it.format?.isSimplexLink ?: false }
   if (links.size == 1) {
     val fmt = links[0].format as Format.SimplexLink
-    return ConnectTarget.Link(links[0].text, fmt.linkType, fmt.simplexLinkText)
+    // showText is set only for a link written as a markdown hyperlink, whose text is the whole
+    // "[label](link)" source rather than the link - connect via the parsed uri in that case
```
epoberezkin:
```suggestion
```

apps/ios/Shared/Views/NewChat/NewChatView.swift
```diff
@@ -859,8 +859,10 @@ enum ConnectTarget {
 func strConnectTarget(_ str: String) -> ConnectTarget? {
     let parsedMd = parseSimpleXMarkdown(str)
     let links = parsedMd?.filter { $0.format?.isSimplexLink ?? false } ?? []
-    return if links.count == 1, case let .simplexLink(_, linkType, _, smpHosts) = links[0].format {
-        .link(text: links[0].text, linkType: linkType, linkText: simplexLinkText(linkType, smpHosts))
+    // showText is set only for a link written as a markdown hyperlink, whose text is the whole
+    // "[label](link)" source rather than the link - connect via the parsed uri in that case
```
epoberezkin:
```suggestion
```

PR #7416, opened 2026-08-25
android, desktop, ios : reconnect calls instead of ending them on the first connection issue

GuavaGoat:
Before a call would disconnect from any slight connection issue, for example using cellular data and having a bad connection would disconnect your call on any slight connection issue or just using wifi and having a bad connection would also break the call.

Now a call has 30 seconds to reconnect and shows users reconnecting if the call is reconnecting (and plays a connecting sound loop so you know it's reconnecting if you are not looking at the screen) because of any slight  connection issues,

Tested on Linux desktop, built from source, with two instances calling each other and the media path interrupted with iptables and the audio was resumed just fine.


GuavaGoat:

epoberezkin:
moved to #7504

PR #7427, opened 2026-08-27
core: badge codes schema, plan

spaced4ndy:


apps/simplex-badge-service/src/BadgeService/Store/Postgres/Migrations.hs
```diff
@@ -31,7 +31,43 @@ m20260806_badge_service_schema =
       servicePrefix
       [r|
 ALTER TABLE @payments ADD COLUMN receipt_hash BYTEA;
+
+CREATE TABLE @badge_codes(
```
epoberezkin, line 35: we could insert prefix? or better take from code?

PR #7447, opened 2026-09-01
website: translations

shumvgolove:


Narasimha-sc:
Ran an adversarial review over every changed string in this branch and pushed the corrections in f7b256f.

### No malicious content

No injected markup, no scheme-changed or attacker-controlled URLs, no bidi / zero-width / homoglyph characters, and no security or privacy claim weakened, strengthened or reversed. Every `href` in every changed string was byte-identical to `en.json`, and all 15 external URLs resolve. JSON is valid, tag-balanced, duplicate-free, key-order-preserving, and placeholders (`%size%`, `%error%`) are intact.

### Blocking corrections

| | |
|---|---|
| **fr** `tap-the-connect-button-in-the-app` | Named a button that does not exist. The French app label is **« Se connecter »** (`MR/fr/strings.xml` `connect_button`); the branch had dropped "se". This is the only website string that quotes an in-app label. |
| **it** `index-token-p2-cf` | "equity crowdfunding" → « raccolta fondi », the Italian term for a **charitable fundraiser**, on a string describing a securities offering. It also contradicted its own untouched neighbour `index-token-cta-cf` ("Investi in SimpleX Chat"). Restored to « l'equity crowdfunding ». |
| **es** `index-hero-invest` | Infinitive « Invertir » rendered as *"Invertir en SimpleX Chat. Descubre más."* — the hero is *tú*-imperative everywhere else (`Sé Libre`, `Haz Crecer Tus Comunidades`, `Únete al equipo`). French got this right in the same branch. |
| **es** `index-hero-invest-cta` | "Learn more **on** Wefunder" → « acerca de » (= *about Wefunder*). `acerca de` was the only occurrence in the file; every other "about" is `sobre`. |
| **hu** `more-info` / `hide-info` | `more-info` moved to « tudnivalók » while its partner `hide-info` still said « információ ». They are the two states of one button (`sections/simplex_unique.html:38,46`), so `hide-info` is realigned. |

### French quality

- **`ratchet steps` → « chaîne de clés »** (= keychain). de/es/it/pt_BR/cs all keep *ratchet*, as does the sibling string in the same table.
- **Five Wikipedia links de-localised** to `en.wikipedia.org`, while `simplex-network-overlay-card-1-p-1` was left pointing at `fr.wikipedia.org` — so the file was split. All seven French targets still resolve, and es/it deliberately localise theirs, so the `fr.` targets are restored, along with the French release title *Désigné coupable*.
- **"each node is discoverable" → « identifiable »**, contradicting the same sentence's « ne permet pas de découvrir les serveurs ».
- **« relais de messagerie »** reads as an SMTP mail relay; `docs-dropdown-16` now says « relais de chat » and matches the nominal pattern of `docs-dropdown-4/5`.
- **Terminology realigned** against untouched strings: `base de données des chats`, `chatbot`, `files d'attente`, `code QR`, `annuaire`, and the infinitive used in the comparison table.
- **Dangling feminine pronoun** in `the-instructions--source-code` ("comment **la** télécharger" with no antecedent in the rendered sentence).
- **Typography unified** across the 92 strings this branch touches — they were split three ways on non-breaking spaces (U+202F / U+00A0 / plain) and two ways on apostrophes and dashes.

47 strings changed, no keys added or removed, all four files still complete against `en.json`. Only two of the 47 were outside this branch's own diff (`hu/hide-info`, `fr/index-directory-cta`), both required to fix a split the branch introduced.

### Worth keeping — real fixes in this branch

`simplex-unique-overlay-card-1-p-1` corrected the stale first-person « **nous** ne savons pas … **nos** serveurs » (which implied SimpleX runs all servers) to « les opérateurs des serveurs »; « graphiques sociaux » → « **graphes** sociaux » across five strings (*graphique* means *chart*); agreement fixes in « qui **a** utilisé », « bloqué**es** », « sous-réseaux isolé**es** »; and French goes from 324 to 380 keys, closing the whole `/file` and `/why` pages.

### Left alone — pre-existing, not introduced here

- **fr `hero-overlay-card-3-p-3` has no link at all.** EN carries `<a href="/blog/20241014-…">Read more</a>`; the French string just ends. Only fr is affected.
- **hu `hero-overlay-card-3-p-3` overstates the 2024 review** as a repeat *audit* of cryptographic **and networking** components; EN says "reviewed cryptographic design". es/it/fr render the EN scope correctly.
- **es `docs-dropdown-4/5`** « Servidor SMP/XFTP anfitrión » reads as a noun phrase, not "Host an SMP server"; `docs-dropdown-16` also loses "relay".
- **`#join-simplex` is a dead anchor on `/file`** — `file-too-large` (from `en.json`, in every locale) links to it, but `sections/join_simplex.html` is included only by `contact_page.html` and `old.html`.
- **`index-hero-invest-cta` is unreferenced** — `index.html:168` builds that CTA from `index-hero-invest` + `learn-more`.

### One thing worth a CI guard

`src/file.html:244-276` injects 32 `file-*` translations into an inline `<script>` via `{{ … | dump | safe }}`. Nunjucks `dump` is `JSON.stringify`, which does **not** escape `</script>`. A translation containing that literal would break out of the script element and inject arbitrary HTML into every `/<lang>/file` page. Every locale on `master` and on this branch is clean — I scanned all 23 — but these strings come from Weblate contributors, so rejecting `</script` / `<!--` in `file-*` values (or escaping `<` in the dump) seems worth adding.

⚠️ Note: these commits were pushed straight to the Weblate-managed branch, so Weblate does not know about them — they may need to be pulled into Weblate, or they will be lost on its next push.

Ranak4n:

PR #7463, opened 2026-09-07
ui: use backward compatible Java api in stripFormattedTextLink

Narasimha-sc:
Fixes a crash when showing a message with a chat link where the text before the link ends with markdown formatting (e.g. `*hi*`). Reported in 7.0.1 on Android 10:

```
java.lang.NoSuchMethodError: No interface method removeLast()Ljava/lang/Object; in class Ljava/util/List; or its super classes (declaration of 'java.util.List' appears in /apex/com.android.runtime/javalib/core-oj.jar)
    at chat.simplex.common.views.chat.item.TextItemViewKt.stripFormattedTextLink(TextItemView.kt:625)
    at chat.simplex.common.views.chat.item.TextItemViewKt.MarkdownText-PQrcpis(TextItemView.kt:151)
    at chat.simplex.common.views.chat.item.FramedItemViewKt.FramedItemView$ciQuotedMsgTextView(FramedItemView.kt:60)
```

`MutableList.removeLast()` resolves to `java.util.List.removeLast()` when compiled with JDK 21. That method only exists from API 35, so it throws on Android below 15.

#6839 changed the other `removeLast()` in this function but missed this one.


Narasimha-sc:
Recreating from a branch on the upstream repo.

PR #7475, opened 2026-09-08
[cli] Wallet API

brenzi:
Implement a wallet and its secret seed and derivation management. cli only `/_wallet ...` API. No need to become user-facing as there will be `/name` commands using the wallet depending on user intention
Must be future proof for (at least) all that comes in prototype #7425.

Limitations
* max one seed per device. refuse to import or create more
* no concept of names or other assets yet, apart from its own derivation dimension
* no hidden names/assets for hidden profiles yet (see #7509 )

Dependencies:
* https://github.com/simplex-chat/simplexmq/pull/1843


src/Simplex/Chat/Controller.hs
```diff
@@ -741,6 +749,13 @@ allowRemoteCommand = \case
   DeleteRemoteCtrl _ -> False
   ExecChatStoreSQL _ -> False
   ExecAgentStoreSQL _ -> False
+  APIWallet -> False
+  APIWalletBind {} -> False
+  APIWalletCreate -> False
+  APIWalletImport _ -> False
+  APIWalletExportSeedMnemonic -> False
+  APIWalletExportDerivedSecret {} -> False
+  APIWalletDelete -> False
```
epoberezkin, line 754: should probably be allowed as remote

brenzi: the idea was to not expose the seed to the remote app which might even log it in plaintext, but it seems this is not the right way to prevent that. on it

brenzi: answere in [4255b46](https://github.com/simplex-chat/simplex-chat/pull/7475/commits/4255b46c04d74230f2c89ab5f23f6212f8123b32). plz give it a look

brenzi: answered in [4255b46](https://github.com/simplex-chat/simplex-chat/pull/7475/commits/4255b46c04d74230f2c89ab5f23f6212f8123b32). plz give it a look. I'm not sure about it

src/Simplex/Chat/Controller.hs
```diff
@@ -435,6 +419,11 @@ data ChatCommand
   | APIRejectContact {contactReqId :: Int64, notify :: Bool}
   | APISendServiceRequest {userId :: UserId, sendTarget :: ConnectTarget 'CMContact, requestTimeout :: Maybe NominalDiffTime, signKey :: Maybe (C.StoredPrivateKey 'C.Ed25519), request :: J.Object}
   | APISendServiceResponse {userId :: UserId, requestId :: AgentInvId, responseData :: J.Object}
+  | APIWallet
```
epoberezkin:
```suggestion
  | APIGetWallet
```

it's confusing otherwise

src/Simplex/Chat/Controller.hs
```diff
@@ -435,6 +419,11 @@ data ChatCommand
...
+  | APIWalletCreate {recoveryPhrase :: Maybe Text}
+  | APIWalletExportSeedMnemonic
+  | APIWalletExportNameSecret {nameIndex :: NameIndex}
+  | APIWalletDelete
```
epoberezkin: We usually follow the convenision APIVerbObject

src/Simplex/Chat/Controller.hs
```diff
@@ -767,6 +751,16 @@ allowRemoteCommand = \case
   ExecAgentStoreSQL _ -> False
   _ -> True
 
+-- | Command text for a log, with any secret blanked. A secret is the last
+-- argument and takes the rest of the line, so blanking from its name is enough.
+redactedCommand :: Text -> Text
+redactedCommand s = foldl' blank s ["mnemonic=", "secret="]
```
epoberezkin: this is very ad-hoc, needs to be more systematic when command constructor is processed, not the text

src/Simplex/Chat/Remote.hs
```diff
@@ -552,7 +552,7 @@ liftRC = liftError (ChatErrorRemoteCtrl . RCEProtocolError)
 
 handleSend :: (ByteString -> Int -> CM' (Either ChatError ChatResponse)) -> Text -> Int -> CM' RemoteResponse
 handleSend execCC command retryNum = do
-  logDebug $ "Send: " <> tshow command
+  logDebug $ "Send: " <> tshow (redactedCommand command)
```
epoberezkin: I would just remove this log, or it has to be pushed to after-parser

src/Simplex/Chat/Wallet.hs
```diff
@@ -0,0 +1,73 @@
...
+type NameIndex = Word32
+
+data WalletSeed = WalletSeed
+  { wsId :: SeedId,
+    wsEntropy :: ByteString
+  }
+  deriving (Eq)
+
+instance Show WalletSeed where
+  show s = "WalletSeed " <> show (wsId s) <> " <redacted>"
```
epoberezkin: The solution to avoid ByteString from being used in Show instances is to use ScrubbedBytes, not custom show instances

src/Simplex/Chat/Store/Postgres/Migrations/M20260921_wallet_seeds.hs
```diff
@@ -0,0 +1,30 @@
...
+
+module Simplex.Chat.Store.Postgres.Migrations.M20260908_wallet_seeds where
+
+import Data.Text (Text)
+import Text.RawString.QQ (r)
+
+m20260908_wallet_seeds :: Text
+m20260908_wallet_seeds =
+  [r|
+CREATE TABLE wallet_seeds (
```
epoberezkin, line 13: there should be user_id in the table, as seed should be attached to a user_id in the schema - I thought we agreed that?

src/Simplex/Chat/Store/Wallets.hs
```diff
@@ -0,0 +1,49 @@
...
+#if defined(dbPostgres)
+import Database.PostgreSQL.Simple (Only (..))
+#else
+import Database.SQLite.Simple (Only (..))
+#endif
+
+toSeed :: (Int64, ByteString) -> WalletSeed
+toSeed (sId, seed) = WalletSeed {wsId = sId, wsEntropy = seed}
+
+getDeviceSeed :: DB.Connection -> IO (Maybe WalletSeed)
```
epoberezkin: why not user seed?

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -5905,6 +5453,18 @@ withExpirationDate globalTTL chatItemTTL action = do
   let ttl = fromMaybe globalTTL chatItemTTL
   when (ttl > 0) $ action $ addUTCTime (-1 * fromIntegral ttl) currentTs
 
+walletNamesShown :: Int
+walletNamesShown = 2
+
+deviceSeed :: CM WalletSeed
+deviceSeed = withFastStore' getDeviceSeed >>= maybe (throwCmdError "no wallet key on this device") pure
```
epoberezkin: trivial function, don't think it is needed

docs/rfcs/2026-09-10-wallet-keys.md
```diff
@@ -0,0 +1,185 @@
...
+- **A paired device.** Can run any of these commands, because they are not blocked from one: `export master` reads the whole wallet, `create` on a device that has none plants a seed the pairing controls, and `delete` destroys the only copy. Blocking `ExecChatStoreSQL` while allowing `export master` is not a coherent line, and the wallet commands need their own decision rather than the catch-all.
+- **Someone reading the logs.** A remote session logs the whole command line at debug level, so a phrase typed into `create` reaches the log there. Answers are never logged. Narrowing that line is work in the remote session rather than in the wallet.
+- **A page open in the user's browser.** The websocket server in `apps/simplex-chat/Server.hs` accepts any local connection, asks for no token and checks no `Origin`, and websockets are not bound by the same origin policy, so any page loaded while that server runs can send `export master` and read the answer. It also prints every command it receives, that phrase included. Both are properties of that server, which this change gives something worth taking, and closing them is work there rather than in the wallet.
```
brenzi, line 153: @epoberezkin I don't think we're ok with this. but how do you suggest we mitigate it? The latter is limited to cli use, but still.

brenzi: claude: 

> For the remote session, there is a clean in-scope fix — and your RFC already names it. allowRemoteCommand is a deny-list in Controller.hs, a file this PR already touches, and the RFC says "the wallet commands need their own decision rather than the catch-all." Denying APIExportWalletMnemonic, APIExportWalletAccount, APICreateWallet and APIDeleteWallet over a pairing costs about four lines and closes the bigger threat you list two bullets earlier — a paired device reading the whole wallet or destroying the seed. It also drains the log exposure of its content: with import refused over a pairing, there's no longer a reason for a recovery phrase to be typed into a remote session at all. The logDebug line still logs whatever is sent, so this is mitigation by removing the reason, not by redaction.
> 
> For the browser page, no wallet-side fix exists. I checked: Server.hs calls execChatCommand', which takes no CommandSource, so the handler genuinely cannot tell a browser page from the terminal. Making it able to would mean plumbing a source through shared code —  exactly the drive-by we just removed. Two things do soften it, though: the server is opt-in (chatServerPort :: Maybe String, only with --chat-server-port), which is worth stating in that bullet since it currently reads as always-on; and the two defects there are

src/Simplex/Chat/Store/Wallets.hs
```diff
@@ -0,0 +1,163 @@
...
+createWalletSeed :: DB.Connection -> BA.ScrubbedBytes -> Maybe AccountIndex -> IO Bool
+createWalletSeed db entropy nextAccount =
+  getWalletSeed db >>= \case
+    Just _ -> pure False
+    Nothing ->
+      True
+        <$ DB.execute
+          db
+          "INSERT INTO wallet_seeds (entropy, next_account_index) VALUES (?, ?)"
+          (DB.Binary (BA.convert entropy :: ByteString), accountIndexCol <$> nextAccount)
```
epoberezkin: this is not concurrency safe, in general, so better to use insert with on conflict and returning clause.

src/Simplex/Chat/Library/Commands.hs
```diff
@@ -1497,6 +1500,36 @@ processChatCommand cxt nm = \case
...
+  APIGetWallet -> withUser $ \user@User {userId} ->
+    CRWallet user <$> withFastStore' (\db -> getWalletSeed db $>>= \WalletSeed {wsId} -> Just <$> getUserAccounts db wsId userId)
+  APICreateWallet mnemonic_ -> withUser $ \_ -> do
+    -- a generated seed has taken no accounts, an imported one does not say how many it has taken
+    (entropy, nextAccount) <- case mnemonic_ of
+      Nothing -> (,Just 0) <$> (asks random >>= atomically . newSeedEntropy)
+      Just phrase -> (,Nothing) <$> liftWallet (entropyFromMnemonic $ encodeUtf8 phrase)
+    created <- withFastStore' $ \db -> createWalletSeed db entropy nextAccount
+    unless created $ throwWalletError WEMasterExists
+    processChatCommand cxt nm APIGetWallet
```
epoberezkin: this is unnecessary indirection, better to extract a function and call it.

epoberezkin: which is arguably not even needed as it's just db read.

epoberezkin: though we already read to update? so why do we read again here?

src/Simplex/Chat/Store/Wallets.hs
epoberezkin: and in general more common convention to read in controller, fail there if it's already exists, and use insert with on conflict.

docs/rfcs/2026-09-10-wallet-keys.md
brenzi: EP: drop the log line. don't worry about websockets, they will be marked as deprecated practice

src/Simplex/Chat/Store/Wallets.hs
brenzi: > Done: the controller reads the seed and fails with WEMasterExists before generating or parsing anything, and the insert is ON CONFLICT 
> 
> (single_seed) DO NOTHING RETURNING wallet_seed_id, so a concurrent create fails the same way.

src/Simplex/Chat/Library/Commands.hs
brenzi: > Removed. Create answers directly, as a new seed has no accounts to read, and bind gets the profile's accounts from the same transaction that binds, so nothing is read twice.

docs/rfcs/2026-09-10-wallet-keys.md
brenzi: > Removed the Send: debug log in Remote.hs, so a phrase sent from a paired device reaches no log, and updated the bullet. 
> 
> The websocket server stays as it is in this PR: it is opt-in (--chat-server-port) and listens on 127.0.0.1 only, and the missing Origin check and the command printing predate the wallet. (to be deprecated anyway)

PR #7488, opened 2026-09-11
fix(markdown): Implement standard Markdown formatting (fixes #7221) 

### LINKED ISSUE #7221: [Bug]: Markdown support is not working as expected

### Platform

Android

### OS version

Android 16

### App version

6.5.6

### Current Behavior

Markdown is used everywhere. *Even here!*. 

Yet, there's a false markdown support in SimpleX Chat. I've found #5675 that's requesting advanced Markdown features, but even the basic feature aren't implemented correctly. So maybe make the basic work as expected before the fancy stuff:

1. In markdown `*X*` puts X in *italic* (emphasis) not bold like SimpleX is doing
2. `**X**` puts X in **bold** (not ignored like SimpleX is doing)
3. `_X_` is _italic_ in both software but `__X__` is bold in Markdown
4. `~X~` or `~~X~~` is ~strike-through~ in Markdown (no equivalent in SimpleX IIUC)

Can you fix the behavior so it matches Markdown basic feature set? This is important when integrating SimpleX with tools (like AI bots that outputs standard Markdown like they breath, or when copy & pasting from the web) 

### Expected Behavior

Very basic Markdown compatible parsing working (like bold, italic, strike-through). Links are working already, so that's great.

### Steps To Reproduce

1. Open private note
2. Type `*This is italic*. **This is bold**` and send it
3. You should see : *This is italic*. **This is bold**
4. You see: **This is italic**. `**This is bold**`

### Relevant log output

```shell

```

X-Ryl669:
Currently, simplex-chat implemented markdown wrongly (as reported in #7221)
So this PR implements very basic Markdown correctly for supported text spans (bold, italic, strikethrough):

- Single asterisk (*text*) now produces italic instead of bold
- Double asterisk (**text**) now produces bold
- Single underscore (_text_) produces italic
- Double underscore (__text__) produces bold
- Tilde (~text~ or ~~text~~) produces strikethrough

This makes SimpleX Chat compatible with standard Markdown syntax, enabling proper integration with AI bots and content from the web.

Since Markdown is ambiguous (`*T*` and `_T_` both produce the same format), instead of converting the AST to a single node and loose the original utterance, that wouldn't survive round tripping, the logic was to enhance the AST format to remember the initial formatting char. This ensures that the text supports lossless parsing and regenerating (and this can be improved for later implementation of other Markdown features).

I've changed the type for this additional field and let it propagate to documentation. 

Actual clients (swift, ...) ignore the additional field so it shouldn't impact them.

I've run the tests for this change and it's all green.

### DIFF
--- a/bots/api/TYPES.md
+++ b/bots/api/TYPES.md
@@ -2140,12 +2140,15 @@ NO_FILE:
 
 Bold:
 - type: "bold"
+- mark: string
 
 Italic:
 - type: "italic"
+- mark: string
 
 StrikeThrough:
 - type: "strikeThrough"
+- mark: string
 
 Snippet:
 - type: "snippet"
--- a/src/Simplex/Chat/Help.hs
+++ b/src/Simplex/Chat/Help.hs
@@ -309,9 +309,9 @@ markdownInfo =
   map
     styleMarkdown
     [ green "Markdown:",
-      indent <> highlight "*bold*         " <> " - " <> markdown Bold "bold text",
-      indent <> highlight "_italic_       " <> " - " <> markdown Italic "italic text" <> " (shown as underlined)",
-      indent <> highlight "~strikethrough~" <> " - " <> markdown StrikeThrough "strikethrough text" <> " (shown as inverse)",
+      indent <> highlight "**bold**       " <> " - " <> markdown (Bold "**") "bold text",
+      indent <> highlight "*italic*       " <> " - " <> markdown (Italic "*") "italic text" <> " (shown as underlined)",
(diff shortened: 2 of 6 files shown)


X-Ryl669:

epoberezkin:
We won't be adopting "standard markdown" (whatever it means, as there is no such thing) piecemeal, it will only happen wholesale, and MVP is:

- large message support (where wider markdown support becomes important).
- full-screen large message editing - that would include headings and tables support.
- nested markdown (requires full rework, not cosmetic changes).
- rich text editor mode (e.g. like in Reddit) - because typing ** is just inconvenient, and most people would continue using *, until a better alternative is offered.

A recent change already makes ** supported as bold during parsing, so it simplifies LLM usage.

I see it all as the next year problems, also related to email integration (which is likely 2028 problem).

PR #7525, opened 2026-09-17
Promote buy/obtain/renew names in name lookup flows and surface mutations

brenzi:
surface SMP v22 enhancements to user: More detailed information about names, promoting opportunities to register/renew names or reach out for reserved ones.


src/Simplex/Chat/Controller.hs
```diff
@@ -1471,12 +1473,21 @@ data ChatError
   | ChatErrorRemoteHost {rhKey :: RHKey, remoteHostError :: RemoteHostError}
   deriving (Show, Exception)
 
--- why a resolved SimpleX name could not be used (the name itself resolved; an unregistered name is the agent's NAME NOT_FOUND)
+-- why a resolved SimpleX name could not be used
 data SimplexDomainError
   = SDENoValidLink -- the name's record has no usable contact/channel link
-  | SDEUnknownDomain -- the resolved link's profile has no name, or a different name
+  | SDEUnknownDomain {claimedDomain :: Maybe SimplexDomain} -- the resolved link's profile has no name, or a different name
```
epoberezkin: we don't need to communicate which name, we don't show it anyway.

src/Simplex/Chat/Controller.hs
```diff
@@ -1471,12 +1473,21 @@ data ChatError
...
+  | SDEUnavailable {registration :: NameRegistration} -- the name is not registered
```
epoberezkin: That's what I objected against - I propose it's a separate case in the plan, not error: CPSimplexDomain

src/Simplex/Chat/Controller.hs
```diff
@@ -1471,12 +1473,21 @@ data ChatError
...
   deriving (Eq, Show)
 
+-- the registry answer for a resolved name
+data NameStatus = NameStatus
+  { response :: NameResponse,
+    ownerChanged :: Bool, -- the owner differs from the found chat's stored answer
+    otherUser :: Maybe User -- another profile on this device the name leads to
```
epoberezkin: probably can be dropped - it would complicate things a lot

src/Simplex/Chat/Controller.hs
```diff
@@ -1471,12 +1473,21 @@ data ChatError
...
   = SDENoValidLink -- the name's record has no usable contact/channel link
-  | SDEUnknownDomain -- the resolved link's profile has no name, or a different name
+  | SDEUnknownDomain {claimedDomain :: Maybe SimplexDomain} -- the resolved link's profile has no name, or a different name
+  | SDEUnavailable {registration :: NameRegistration} -- the name is not registered
   deriving (Eq, Show)
 
+-- the registry answer for a resolved name
+data NameStatus = NameStatus
+  { response :: NameResponse,
+    ownerChanged :: Bool, -- the owner differs from the found chat's stored answer
```
epoberezkin: how it will be used?

brenzi: I think the user should be told if a name points to a different chat and if yes, if the ownership changed as well. it's about maintaining authenticity...

brenzi:
tested desktop UI manually against smp7: Enter these in the search bar:
* [x] samba.testing -> 2e (too short)
* [x] abcdefg.testing -> 2c
* [x] brenzi.testing -> 2a
* [x] kingofthebongo.testing -> 2f
* [x] #brenzi -> 2g

src/Simplex/Chat/Controller.hs
```diff
@@ -712,6 +712,7 @@ data PlanResolveMode
   = PRMAllGroups -- resolve all known groups and all unknown chats
   | PRMUnknown -- only resolve if chat is unknown (default)
   | PRMNever -- do not resolve links and names, only do local search
+  | PRMAll -- always resolve, also known chats
```
epoberezkin, line 714: not really needed as a separate option, PRMAllGroups can be replaced with PRMAll

apps/ios/Shared/Model/AppAPITypes.swift
```diff
@@ -1455,13 +1455,93 @@ enum PlanResolveMode: String {
...
+    var reservedForCommunity: Bool {
+        switch self {
+        case let .reserved(reason): reason == NameRegistration.reservedCommunity
+        case let .registered(_, _, reason): reason == NameRegistration.reservedCommunity
+        case .available: false
+        }
+    }
+}
+
+extension NameRegistration: Decodable {
```
epoberezkin, line 1500: why does it need non-standard JSON derivation and can't use stock JSON?

apps/ios/Shared/Model/ChatModel.swift
```diff
@@ -1499,3 +1499,51 @@ enum UIRemoteCtrlSessionState {
...
+enum NameResolution {
+    private static let daySeconds: Int64 = 24 * 60 * 60
+
+    private struct Resolved: Codable {
+        var at: Int64
+        var expires: Int64?
+    }
+
+    // when each SimpleX name was last resolved, so a name with a chat is not re-resolved on every tap
+    private static let resolvedAtDefault = CodableDefault<[String: Resolved]>(defaults: UserDefaults.standard, forKey: DEFAULT_SIMPLEX_NAMES_RESOLVED_AT, withDefault: [:])
```
epoberezkin: this is mixing concerns - don't put access to prefs into data types, it should be in views

epoberezkin: this is classic OOO anti-pattern, we don't do OOO. Types should hold data, they should not hold IO operations.

PR #7530, opened 2026-09-17
minimal name registration API

brenzi:


brenzi:
TODO:
* [x] screen 3: 

> I am also not sure that typing name and pressing continue is the best UX - all domain registrar allow you to search multiple options even if the name is available, so I think continue button should lead to registration AFTER it's confirmed it's available, but all information is just shown on the same screen.
> and you can search for another name.
> So Continue -> Register, and maybe some search button on the right of the field? like looking glass?
> and keyboard also has Search option
> so you type, tap search, and it shows the name status not as alert, but as screen state, with options like:
> 
> 1 - available to register and price - Register button gets enabled
> 2 - reserved - button Connect to us, disabled Register below
> 3 - taken - one or two found chats, with Open [new] chat buttons and disabled Register.
> 
> This way you can type multiple names to decide what you want, which is what best domain registrars offer - they don't assume you want the name simply because it's available, they offer multiple options.
> 
> Also consider the evolution of this UX when we can offer:
> 
> - multiple TLDs in SimpleX network
> - maybe we can do web domain registration integration too (and auto-map registered domain to SimpleX)
> - partial lookups - post MVP but valuable.
> 
> So it's a "name search" screen, not "name check" screen
> 
> this name search philosophy also suggests that if the name is taken, we don't show what uses it, but it's one tap away - maybe we show "taken" and channel/contact/both as tappable, and if they are tapped it would show alert to open chat

* [x] 5a/5b: don't drop. they don't add much complexity. but don't show alerts but integrate them into 3b (like 3a, but not checked while typing) lookup result messages, a tappable "see what uses this name" leading to our standard alert for opening a connection (might need another lookup?). Applies to 5c as well: move to 3b inline instead of alert

PR #7532, opened 2026-09-17
core, ui: alert the user when badge renewal is failing

spaced4ndy:


src/Simplex/Chat/Badges/Service.hs
```diff
@@ -205,74 +204,6 @@ data StatementDebitType
   | SDUnknown {tag :: Text, json :: J.Object}
   deriving (Show)
 
-data BadgeServiceErrorCode
```
spaced4ndy, line 208: moved to Types to avoid cyclic dependency

PR #7536, opened 2026-09-18
desktop: fix laggy bars blur by blurring a narrower offscreen copy

Narasimha-sc:
The bars blur (Settings > Appearance > Blur) makes desktop scrolling lag: at the default radius the two bars of a chat cost about 80 ms per frame on the software renderer, cutting scrolling to 5-9 fps.

Skia's raster blur runs at full resolution for every radius the slider offers, so each bar blurs width x (bar + 6 sigma) pixels every frame, and scaling the layer cannot help because a layer's filter is evaluated in device space, which scales the sigma back up with it.

Each bar is now drawn into an offscreen surface up to 8x narrower and blurred there, measured 5-6x cheaper and within 0.25/255 of the original blur; see plans/desktop-bars-blur-performance.md. Android is untouched.


Narasimha-sc:
Reopened from a branch in this repo instead of my fork: #7556. Same commit, rebased on master.

